PR #5, opened 2018-03-05
Quick fix to allow int parameters to be positive and negative

ggrieco-tob:


japesinator:
This feels kinda hacky to me. I think we have a couple options for doing it slightly more cleanly:

  * multiply by `Gen.choice [1, -1]`

  * rewrite to use `Integer` and `fromIntegral`

I think I mildly prefer the second, but am open to either

lib/Echidna/ABI.hs
```diff
@@ -62,21 +62,23 @@ encodeSig n ts = n <> "(" <> T.intercalate "," (map abiTypeSolidity ts) <> ")"
...
-  liftM2 Word256 (liftM2 Word128 w64 w64) (liftM2 Word128 w64 w64)
-
 genAbiAddress :: MonadGen m => m AbiValue
 genAbiAddress = let w64 = Gen.word64 $ constant minBound maxBound in
   fmap AbiAddress . liftM2 Word160 Gen.enumBounded $ liftM2 Word128 w64 w64
 
+maxInteger :: Integer
+maxInteger = 2^(256 :: Integer)
+
+genUInt :: MonadGen m => m Integer
```
japesinator: I'm not sure these two are quite correct. Aren't 256 bit UInts in [0, 2^256 -1], and Ints in [-2^128, 2^128-1]?

lib/Echidna/ABI.hs
```diff
@@ -62,21 +62,23 @@ encodeSig n ts = n <> "(" <> T.intercalate "," (map abiTypeSolidity ts) <> ")"
...
+genUInt = Gen.integral $ exponential 0 maxInteger
+genInt :: MonadGen m => m Integer
```
japesinator: Let's add a line between these

lib/Echidna/ABI.hs
```diff
@@ -62,21 +62,23 @@ encodeSig n ts = n <> "(" <> T.intercalate "," (map abiTypeSolidity ts) <> ")"
...
+genInt = liftM2 (*) genUInt $ Gen.element [1, -1]
+
 genAbiUInt :: MonadGen m => Int -> m AbiValue
-genAbiUInt n = AbiUInt n . scale <$> genWord256 where
-  scale = if n == 256 then id else flip mod (2^n)
+genAbiUInt n = AbiUInt n . fromInteger <$> genUInt where
```
japesinator: Not sure this is correct when n < 256

lib/Echidna/ABI.hs
```diff
@@ -62,21 +62,19 @@ encodeSig n ts = n <> "(" <> T.intercalate "," (map abiTypeSolidity ts) <> ")"
...
-  scale = if n == 256 then id else flip mod (2^n)
+genAbiUInt n = AbiUInt n . fromInteger <$> genUInt (toInteger n)
+               where genUInt x = Gen.integral $ exponential 0 $ 2^x - 1
 
 genAbiInt :: MonadGen m => Int -> m AbiValue
-genAbiInt n = AbiInt n . signedWord . scale <$> genWord256 where
-  scale = if n == 256 then id else flip mod (2^n)
+genAbiInt n = AbiInt n . fromInteger <$> genInt (toInteger n)
+              where genInt x = liftM2 (*)
+                                (Gen.integral $ exponential 0 $ 2^(x `div` 2) - 1)
```
japesinator: this looks like a bug, INT_MAX == 2^32-1, not 2^16-1

lib/Echidna/ABI.hs
```diff
@@ -62,21 +62,19 @@ encodeSig n ts = n <> "(" <> T.intercalate "," (map abiTypeSolidity ts) <> ")"
...
-  liftM2 Word256 (liftM2 Word128 w64 w64) (liftM2 Word128 w64 w64)
-
 genAbiAddress :: MonadGen m => m AbiValue
 genAbiAddress = let w64 = Gen.word64 $ constant minBound maxBound in
   fmap AbiAddress . liftM2 Word160 Gen.enumBounded $ liftM2 Word128 w64 w64
 
 genAbiUInt :: MonadGen m => Int -> m AbiValue
-genAbiUInt n = AbiUInt n . scale <$> genWord256 where
-  scale = if n == 256 then id else flip mod (2^n)
+genAbiUInt n = AbiUInt n . fromInteger <$> genUInt (toInteger n)
```
japesinator: instead of having a unary `genUInt`, `n` is in scope in the local def so why not just have it constant?

lib/Echidna/ABI.hs
```diff
@@ -62,21 +62,19 @@ encodeSig n ts = n <> "(" <> T.intercalate "," (map abiTypeSolidity ts) <> ")"
...
+               where genUInt x = Gen.integral $ exponential 0 $ 2^x - 1
 
 genAbiInt :: MonadGen m => Int -> m AbiValue
-genAbiInt n = AbiInt n . signedWord . scale <$> genWord256 where
-  scale = if n == 256 then id else flip mod (2^n)
+genAbiInt n = AbiInt n . fromInteger <$> genInt (toInteger n)
+              where genInt x = liftM2 (*)
```
japesinator: as above, why unary?

PR #11, opened 2018-03-08
Added a stack reset to cleanUp to avoid overflow and fix #7

### LINKED ISSUE #7: Echidna seems to fails a test that always returns true
I tried to test the following contract:

```solidity
contract Test {
  function f(int val) returns (bool){
    return(true);
  }
  function echidna_alwaystrue() returns (bool){
    return(true);
  }
}
```

increasing the amount of tests:

```diff
diff --git a/src/Main.hs b/src/Main.hs
index b9280e4..a19c2b6 100644
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -14,6 +14,6 @@ main = getArgs >>= \case
   []  -> putStrLn "Please provide a solidity file to analyze"
   f:_ -> do
     (v,a,ts) <- loadSolidity f
-    let prop t = (PropertyName $ show t, ePropertySeq v a (`checkETest` t) 10)
+    let prop t = (PropertyName $ show t, ePropertySeq v a (`checkETest` t) 1000)
     _ <- checkParallel . Group (GroupName f) $ map prop ts
     return ()
```

But echidna fails when it runs echidna_alwaystrue:

```
$ echidna-test test.sol 
━━━ test.sol ━━━
  ↯ "echidna_alwaystrue" failed after 46 tests (shrinking)
  ○ 0/1 complete (running)
```

(the shrinking seems to take forever, so I don't know what it actually finds)

ggrieco-tob:


japesinator:
🙏

PR #13, opened 2018-03-10
update readline osx specific install instructions 

### LINKED ISSUE #12: OS X Install instructions for readline need more details
For me (and I suspect other people) the real solution is to 

```
port install readline +universal
export CPPFLAGS=-I/opt/local/include
export LDFLAGS=-L/opt/local/lib
stack install readline --extra-include-dirs=/usr/local/opt/readline/include --extra-lib-dirs=/usr/local/opt/readline/lib; stack install
```

(readline is pretty likely to cause some failures during stack otherwise, and the suggested fix is often not quite what people need; there's probably a brew equivalent but I couldn't find it)

snd:
resolves #12


japesinator:
Thanks for adding this! Good catch

PR #20, opened 2018-03-12
Add Dockerfile

lazzarello:
Initially done to bypass the readline issue on macOS. Also good for portability!


japesinator:
Thanks for coming back and finishing this up! Once I've wrapped up compiling the universe and can make sure this works, I'm ready to merge. One thing you might be interested in though, is I've done some preliminary work on building binaries in CI (https://github.com/trailofbits/echidna/blob/master/.travis/attach-binary.sh) and if we'd like things to go a lot faster, we could consider making this grab those binaries instead of compiling them itself. Is that reasonable? I know essentially nothing of docker

PR #24, opened 2018-03-14
resolve #23 (test specific contract)

### LINKED ISSUE #23: test specific contract in file
right now `echidna-test` loads the first contract in a file and tests it.

this is inflexible and doesn't work well with `import`s as they appear to echidna as the first contracts in a file.

**modify `echidna-test` and `loadSolidity` to take `contract-name` as optional second argument**

this would open up a useful pattern where one writes a contract as usual without any echidna related modifications and then inherits a test contract from it that contains the echidna properties and a constructor with 0 params that does the required initialization (see https://github.com/trailofbits/echidna/issues/15).

right now this is not possible since the unmodified contract that is inherited from must come first.

snd:
- resolved #23 
- extracted contract selection into function `selectContract` because `loadSolidity` became unwieldy otherwise
- made `warn` print warning if predicate is true (instead of false) which makes more sense imho

works as expected with contracts in multiple files and imports

### DIFF
--- a/lib/Echidna/Solidity.hs
+++ b/lib/Echidna/Solidity.hs
@@ -11,6 +11,7 @@ import Control.Monad.State.Strict (execState, runState)
 import Data.Foldable              (toList)
 import Data.List                  (find, partition)
 import Data.Map                   ()
+import Data.Maybe                 (isNothing)
 import Data.Monoid                ((<>))
 import Data.Text                  (Text, isPrefixOf)
 import System.Process             (readProcess)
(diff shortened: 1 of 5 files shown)


snd:
will update documentation tomorrow. looking forward to feedback already

ggrieco-tob:
Thanks a lot for taking the time for this contribution! :tada: . The code looks good to me. I just wanted to say that we definitely need to define and implement some command line convention to follow because we cannot continue adding parameters in `echidna-test`. In any case, that is outside the scope of this PR.

japesinator:
thanks for all the work to put this together! I'll do some detailed code review tomorrow morning

lib/Echidna/Solidity.hs
```diff
@@ -22,37 +23,46 @@ import EVM
...
+  show CompileFailure       = "Couldn't compile given file"
+  show NoContracts          = "No contracts found in given file"
+  show (ContractNotFound c) = "Given contract " ++ show c ++ " not found in given file"
+  show (TestArgsFound t)    = "Test " ++ show t ++ " has arguments, aborting"
 
 instance Exception EchidnaException
 
-loadSolidity :: (MonadIO m, MonadThrow m) => FilePath -> m (VM, [SolSignature], [Text])
-loadSolidity f = liftIO solc >>= \case
+selectContract :: (MonadThrow m) => [SolcContract] -> Maybe Text -> m SolcContract
```
japesinator: This isn't quite the abstraction boundary I'd draw. How would you feel about something of type `(MonadIO m, MonadThrow m) => FilePath -> Maybe Text -> m SolcContract`, and then your logic can be something like `let choose Nothing x = True; choose (Just s) x = x ^. contractName == s in liftIO solc >>= \case [...] Just m -> findOf (_1 . each) (choose n) m [...]`

src/Main.hs
```diff
@@ -5,15 +5,17 @@ module Main where
...
 import Echidna.Exec
 import Echidna.Solidity
 
 main :: IO ()
 main = getArgs >>= \case
   []  -> putStrLn "Please provide a solidity file to analyze"
-  f:_ -> do
-    (v,a,ts) <- loadSolidity f
+  filepath:args -> do
+    (v,a,ts) <- loadSolidity filepath $ fmap pack $ listToMaybe args
```
japesinator: nit, but I've tried to prefer `f . g $ h x` over `f $ g $ h x`

lib/Echidna/Solidity.hs
```diff
@@ -22,37 +23,46 @@ import EVM
...
+selectContract [] _ = throwM NoContracts
+selectContract (c:_) Nothing = return c
+selectContract cs (Just name) = case find (\x -> name == x ^. contractName) cs of
+  Nothing -> throwM $ ContractNotFound name
+  Just c  -> return c
+
+loadSolidity :: (MonadIO m, MonadThrow m) => FilePath -> Maybe Text -> m (VM, [SolSignature], [Text])
```
japesinator: Can we refactor the example in README.md to use this new convention?

snd: sure, will do

src/Main.hs
snd: curious why you prefer `f . g $ h x` over `f $ g $ h x`? i find the latter a bit easier to read (more uniform, 1 concept rather than 2) and readability important

japesinator: Consistency when moving from `$` to `<$>`

lib/Echidna/Solidity.hs
snd: that would couple "reading all contracts from file" and "choosing the one to test" together. then they can't be used separately. makes the function more complex and the API less flexible.

how about having a function `loadContracts :: (MonadIO m, MonadThrow m) => FilePath -> m [SolcContract]`.  `(MonadIO m, MonadThrow m) => FilePath -> Maybe Text -> m SolcContract` would then simply be a composition of `loadContracts` and `selectContract`

japesinator: How do you see the average Echidna user using `selectContract`? I'm not sure what analysis would require such a function. I could see the utility of `loadContracts`, and don't have a strong preference between that and the function I proposed above (mildly prefer the latter I think?), but I'm unconvinced we should export a function for selecting a contract from a list thereof in a `MonadThrow`

snd: what do you think of `selectContract :: [SolcContract] -> Maybe Text -> Either EchidnaException SolcContract`? that would get rid of the `MonadThrow`. though the `MonadThrow` makes calling `selectContract` from `loadSolidity` a bit nicer.

i think of `selectContract` as a function encapsulating the strategy of selecting the contract to test. since it has two distinct failure cases we want to communicate we can't have it return `Maybe SolcContract` which would be preferable.

no need to export `selectContract` though i think extracting it into a function makes both functions easier to understand.

snd: > How do you see the average Echidna user using `selectContract`?

i'd like to enable the user to make that choice.
if we export `(MonadIO m, MonadThrow m) => FilePath -> Maybe Text -> m SolcContract` then we make a decision for the user that every time they want to read contracts from a file one contract is chosen using our strategy.
if the user wants something else they have to reimplement half of the function themselves.
if they want to use just the strategy but load contracts another way they have to reimplement again.

not saying that any choice should be enabled for the user. but if it's as straightforward as in this case i think it's worth doing.

big fan of APIs that export a set of small simple functions that each do one thing and can be combined in flexible unforeseen ways. reflects the reality of how people use APIs better imho.

src/Main.hs
snd: how about `loadSolidity filepath $ pack <$> listToMaybe args`?

japesinator: that's fine with me

lib/Echidna/Solidity.hs
japesinator: I think I'm just unclear on when the user would want something not covered by `(MonadIO m, MonadThrow m) => FilePath -> Maybe Text -> m SolcContract`. What kind of echidna testing does this not suffice for? I'd be happy to provide more granular functionality for users that require it, but I think I'd need to see some code that did first.

snd:
changed the abstraction boundaries a bit.

there's now `readContract :: (MonadIO m, MonadThrow m) => FilePath -> Maybe Text -> m SolcContract` which is the abstraction you preferred.

i kept `readContracts :: (MonadIO m, MonadThrow m) => FilePath -> m [SolcContract]`
because i think it's useful on its own.

made `chooseContract :: (MonadThrow m) => [SolcContract] -> Maybe Text -> m SolcContract` an internal function of `readContract`. i'd still slightly prefer for it to be its own function just to simplify `readContract`.

let me know what you think : )

japesinator:
I'm pretty open to changing the API further to reflect the needs of our users in the future, but I think I wanna move forwards with this for now.

PR #27, opened 2018-03-19
Add coverage tracking

japesinator:
This is still a bit of a WIP if only because it's really slow, but we can count unique PCs now

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ import Echidna.Solidity
 
 main :: IO ()
 main = do (v,a,ts) <- loadSolidity "test.sol" Nothing
-          let prop t = (PropertyName $ show t, ePropertySeq v a (`checkETest` t) 10)
+          let prop t = (PropertyName $ show t, ePropertySeq (`checkETest` t) a v 10)
           _ <- checkParallel . Group (GroupName "test.sol") $ map prop ts
           return ()
 ```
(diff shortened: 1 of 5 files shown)


snd:
this is coverage for the contract under test, right?

really excited for this 🎉 

10000 passing tests would mean a lot more in addition to the information that it hit 100% of branches.
makes it much clearer that the fuzzer actually did something useful.

does this track line coverage only or does it track branch coverage as well?

does the fuzzer attempt to hit all branches currently?

japesinator:
so right now, we just toss `state . pc` in a `MultiSet` at every step and merge those in some big `IORef`, so it's just a unique instruction count. At some point, I'd really like to correlate that with the `SrcMap`s solc emits to get line coverage (and then maybe some kind of branch coverage from there? idk static analysis is hard), but rn my priority is making it less slow + getting some notion of total possible coverage

snd:
> getting some notion of total possible coverage

seems like a great first step

japesinator:
also, rn fuzzing is 100% grammar based with no feedback loop. I'm not sure what hedgehog's notion of `CoArbitrary` even looks like, or how to do afl style genetic input improvement on something like `SolCall`, so you may not see anything like that in master for some time, but it is something that interests me

japesinator:
if you're interested in doing this, I guess the easiest thing to do is make `execCallCoverage` happen in a `MonadIO` + take an `IORef` where global coverage is, keep track of the delta per-call, then throw that in some other `IORef` and generate in another `MonadIO` with some kinda mutator `genGoodCall` unioned into the existing machinery, but idk how quickly I'll have time to dev/test all that.

japesinator:
Superseded by #34

Issue #10, opened 2018-03-08
echidna fails to run outside the repository folder


makoto:
Tried the example solution @japesinator suggested during his ETHCC workshop but does not seem working on the latest solc compiler.

```
$  echidna-test contracts/testme.sol 
━━━ contracts/testme.sol ━━━
  ○ 0/1 complete (running)
  ↯ "echidna_alwaystrue" failed after 1 test (shrinking)
echidna-test: lib/Echidna/Exec.hs: openFile: does not exist (No such file or directory)
```

```
pragma solidity 0.4.21;

contract Canal {
  bool private first_gate_up=false;
  bool private second_gate_up=true;

  function lower(bool first_gate) public returns (bool){
    if(first_gate) {
      if(second_gate_up) {
        first_gate_up = false;
        return(true);
      } else {
        return(false);
      }
    } else {
      if(first_gate_up) {
        second_gate_up = false;
        return true;
      } else {
        return(false);
      }
    }
  }
  function raise(bool first_gate) public {
    if(first_gate) {
      first_gate_up = true;
    } else {
      second_gate_up = true;
    }
  }

  function echidna_alwaystrue() public view returns (bool){
    return (first_gate_up || first_gate_up);
  }
}
```

ggrieco-tob:
I cannot reproduce exactly your error:

```bash
$ solc --version
solc, the solidity compiler commandline interface
Version: 0.4.21+commit.dfe3193c.Linux.g++
$ echidna-test contracts/testme.sol 
━━━ contracts/testme.sol ━━━
  ✗ "echidna_alwaystrue" failed after 1 test and 1 shrink.
  
      │ Call sequence: lower(true);
  
  ✗ 1 failed.
```

snd:
i get the same error with solc 0.4.18

```
$ solc --version
solc, the solidity compiler commandline interface
Version: 0.4.18+commit.9cf6e910.Darwin.appleclang
$ echidna-test contracts/testme.sol 
━━━ contracts/testme.sol ━━━
  ○ 0/1 complete (running)
  ↯ "echidna_alwaystrue" failed after 1 test and 1 shrink (shrinking)
echidna-test: lib/Echidna/Exec.hs: openFile: does not exist (No such file or directory)
```

OSX 10.11.6

built from https://github.com/trailofbits/echidna/commit/159b595172e85e67b90ba374c253a40e32c1f568

installed as described there: https://github.com/trailofbits/echidna/issues/12#issuecomment-372033419

[...]

PR #29, opened 2018-03-27
A fix for #10

ggrieco-tob:


lib/Echidna/Internal/Runner.hs
```diff
@@ -565,24 +565,33 @@ takeLines sloc =
   snd . Map.split (spanStartLine sloc - 1) .
   declarationSource
 
+
+checkFilePath :: FilePath -> IO Bool
+checkFilePath path = do
+   strOrExc <- try $ readFile path
+   case (strOrExc :: Either IOError String) of
```
japesinator: how about https://hackage.haskell.org/package/base-4.11.0.0/docs/Data-Either.html#v:isRight?

japesinator: `isRight <$> try $ readFile path`?

lib/Echidna/Internal/Runner.hs
```diff
@@ -565,24 +565,33 @@ takeLines sloc =
...
       line =
         fromIntegral line0
-
-    pure . Declaration path line name .
-      Map.fromList .
-      zip [line..] .
-      zipWith (Line ()) [line..] $
-      lines src
-
+    if (not exists)
```
japesinator: I don't think the parens in this conditional are necessary (but also idk quite what to do about style in this file, so feel free to leave them if you think they're more consistent or etc.)

PR #31, opened 2018-04-02
Extended support to generate and execute more transaction fields (msg.value, msg.sender, ..)

### LINKED ISSUE #30: Allow to specify which user calls each method in a contract
Some smart contracts define an address as "owner" to perform certain "privileged" operations. For instance:

```solidity
contract C {
  uint state = 1;
  address owner;

  function C() {
     owner = msg.sender;
  }

  function only_owner(uint x) public {
      require(msg.sender == owner);
      state = x;
  }

      return (state == 1);
  }
}
```

Unfortunately, Echidna does not allow to specify the use of different users (contract owner, user, attacker) when the transactions are executed. It could be nice to have a command line option for that.

ggrieco-tob:
**This PR is an early prototype for discussion. Not intended to merge it soon.**

Echidna should be able to handle the generation of all the transaction data fields to trigger deep parts of a contract code (for instance, code depending on msg.value or msg.sender). This PR allows to generate and execute them (so, it should fix #30 when it is ready) so echidna can find a failed test in this small example:

```solidity
contract C {

  uint state = 0;
  function f() payable {
    if (msg.value > 128)
        state = msg.value;
  }

  function echidna_test() returns (bool) {
    return(state == 0);
  }
}
```

It outputs:

```bash
  ✗ "echidna_test" failed after 5 tests and 5 shrinks.
  
      │ Call sequence: f() paying 129;
  
  ✗ 1 failed.
```

### DIFF
--- a/examples/revert/Revert.hs
+++ b/examples/revert/Revert.hs
@@ -17,6 +17,6 @@ main = do (v,a,ts) <- loadSolidity "solidity/revert.sol" Nothing
           return ()
 
 checkRTest :: VM -> Text -> Bool
-checkRTest v t = case evalState (execCall (t, [])) v of
+checkRTest v t = case evalState (execCall (t, [], 0, 0)) v of
   (VMFailure Revert) -> False
   _                  -> True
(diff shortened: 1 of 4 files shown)


japesinator:
Partially superseded by #67, mostly not compatible with the new config stuff

PR #37, opened 2018-04-30
Add ability to specify solc compiler options

### LINKED ISSUE #36: Solidity issue with compiling
```
reporting/IFeeToken.sol:4:1: Error: Source "libraries/Initializable.sol" not found: File outside of allowed directories.
import 'libraries/Initializable.sol';
^-----------------------------------^
```

Solidity doesn't allow files in other directories within a source folder by default:

> For security reasons the compiler has restrictions what directories it can access. Paths (and their subdirectories) of source files specified on the commandline and paths defined by remappings are allowed for import statements, but everything else is rejected. Additional paths (and their subdirectories) can be allowed via the --allow-paths /sample/path,/another/sample/path switch.

Echidna should allow passing flags to the solidity compiler so that it can be used for analysis of more complicated contracts. For instance, to resolve this issue it'd be great if I could do something like:

`echidna-test path/to/File.sol --allow-paths {path name}`

Coffee-Boyy:
This fixes #36 allowing custom solc options to be passed through which has the benefit of being able to supply solc with allowable paths to use for imports.

This is my first ever work done in Haskell so all constructive criticism is appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,11 @@ For each unit test it finds, it will execute a fuzzing campaign to try and find 
 An example contract with tests can be found [solidity/cli.sol](solidity/cli.sol)
 `echidna-test solidity/cli.sol` should find a call sequence such that `echidna_sometimesfalse` fails, but be unable to do so for `echidna_alwaystrue`.
 
+Support for multiple contracts in a single file along with importing files from an unsupported directory has bee added by using the following optional command line arguments:
+```
+echidna-test solidity/cli.sol Test2 --solc-args="--allow-paths=/echidna/solidity"
+```
+
 ## Usage (as a library)
 
 Echidna is actively being developed with relatively little regard for stability.
@@ -139,11 +144,6 @@ The [state machine example](examples/state-machine/StateMachine.hs) is a pretty 
 ### [Echidna.Solidity](lib/Echidna/Solidity.hs)
 
 This module provides `loadSolidity`, which takes a solidity source file and provides a VM with the first contract therein loaded as well as a `fuzz`-compatible ABI definition.
-
-Support for multiple contracts in a single file along with importing files from an unsupported directory has bee added by using the following optional command line arguments:
-```
-echidna-test solidity/cli.sol Test2 --solc-args="--allow-paths=/Users/connor/Documents/echidna/solidity"
-```
 
 ## Questions/complaints/etc.
 
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -5,7 +5,7 @@ module Main where
 
 import Control.Concurrent.MVar (takeMVar, newMVar)
 import Data.MultiSet           (distinctSize)
-import Data.Text               (Text, pack)
+import Data.Text               (pack)
 import Data.Semigroup          ((<>))
 
 import Echidna.Exec
@@ -33,10 +33,6 @@ options = Options
           ( long "solc-args"
          <> help "Optional solidity compiler arguments" ))
 
-maybePack :: Maybe String -> Maybe Text
-maybePack (Just s) = Just (pack s)
-maybePack Nothing = Nothing
-
 opts :: ParserInfo Options
 opts = info (options <**> helper)
   ( fullDesc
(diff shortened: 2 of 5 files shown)


stack.yaml
```diff
@@ -22,3 +22,4 @@ extra-deps:
 - tree-view-0.5
 - vty-5.16
 - word-wrap-0.4.1
+- optparse-applicative-0.13.2.0
```
japesinator: I've also tried to alphabetize these, could we similarly move this up a few lines?

src/Main.hs
```diff
@@ -1,29 +1,57 @@
...
+          ( metavar "FILE"
+         <> help "Solidity file to analyze" )
+      <*> optional ( argument str
+          ( metavar "CONTRACT"
+         <> help "Contract inside of file to analyze" ))
+      <*> optional ( strOption
+          ( long "solc-args"
+         <> help "Optional solidity compiler arguments" ))
+
+maybePack :: Maybe String -> Maybe Text
```
japesinator: I think this is just `fmap pack`/`pack <$>`, we probably don't need to define it like this

src/Main.hs
```diff
@@ -1,29 +1,57 @@
...
+import Data.Text               (Text, pack)
+import Data.Semigroup          ((<>))
 
 import Echidna.Exec
 import Echidna.Solidity
 
 import Hedgehog hiding (checkParallel)
 import Hedgehog.Internal.Property (GroupName(..), PropertyName(..))
 
+import Options.Applicative
```
japesinator: Can we explicitly list imports here?

README.md
```diff
@@ -139,7 +139,11 @@ The [state machine example](examples/state-machine/StateMachine.hs) is a pretty
 ### [Echidna.Solidity](lib/Echidna/Solidity.hs)
 
 This module provides `loadSolidity`, which takes a solidity source file and provides a VM with the first contract therein loaded as well as a `fuzz`-compatible ABI definition.
-At the moment, it is only compatible with single-contract solidity files, though if given a multi-contract file it will load the first and print a warning.
+
+Support for multiple contracts in a single file along with importing files from an unsupported directory has bee added by using the following optional command line arguments:
+```
+echidna-test solidity/cli.sol Test2 --solc-args="--allow-paths=/Users/connor/Documents/echidna/solidity"
```
japesinator: Let's change the example path here to something a bit more generic

README.md
```diff
@@ -139,7 +139,11 @@ The [state machine example](examples/state-machine/StateMachine.hs) is a pretty
 ### [Echidna.Solidity](lib/Echidna/Solidity.hs)
 
 This module provides `loadSolidity`, which takes a solidity source file and provides a VM with the first contract therein loaded as well as a `fuzz`-compatible ABI definition.
-At the moment, it is only compatible with single-contract solidity files, though if given a multi-contract file it will load the first and print a warning.
+
+Support for multiple contracts in a single file along with importing files from an unsupported directory has bee added by using the following optional command line arguments:
```
japesinator: Let's restructure this a bit for clarity. In my opinion, this section should just be a description of Echidna.Solidity's API, and then we can move stuff about CLI usage to the "Usage as an Executable" section

src/Main.hs
ConnorChristie: When I started adding explicit imports, there were more than a dozen to add since each of the methods to build the parser info is a separate method. (i.e. fullDesc, progDesc, metavar, ...)

joeykrug:
This functionality seems to have gotten broken and/or accidentally removed:

Running ```echidna-test trading/FillOrder.sol --solc-args="--allow-paths=."
``` on augur contracts yields:

```trading/Order.sol:11:1: Error: Source "IAugur.sol" not found: File outside of allowed directories.
import 'IAugur.sol';
^------------------^
trading/Order.sol:12:1: Error: Source "IController.sol" not found: File outside of allowed directories.
import 'IController.sol';
^-----------------------^
trading/Order.sol:13:1: Error: Source "libraries/math/SafeMathUint256.sol" not found: File outside of allowed directories.
import 'libraries/math/SafeMathUint256.sol';
^------------------------------------------^
trading/Order.sol:14:1: Error: Source "reporting/IMarket.sol" not found: File outside of allowed directories.
import 'reporting/IMarket.sol';```

Running 

```solc trading/FillOrder.sol --allow-paths=.``` works

PR #38, opened 2018-04-30
Added more variety in the generation of strings

ggrieco-tob:
This small improvement should allow to increase coverage in contracts that uses strings.


japesinator:
Let's merge on green!

PR #41, opened 2018-05-01
Dev good input detection

blperez01:
Uses a genetic algorithm for coverage guidance where each generation is given the approximate minimal number of inputs from the last generation that achieved maximal coverage. 

Also modified the command line options as follows:
`echidna-test filename n` runs echidna with coverage using `n` generations for the genetic algorithm
`echidna-test filename` runs echidna without coverage

### DIFF
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -1,7 +1,7 @@
 {-# LANGUAGE FlexibleContexts, KindSignatures, LambdaCase, StrictData #-}
 
 module Echidna.Exec (
-    Coverage
+    ExecEnv(..)
   , checkETest
   , eCommand
   , eCommandCoverage
@@ -10,12 +10,14 @@ module Echidna.Exec (
   , execCall
   , execCallCoverage
   , fuzz
+  , mutateGoodInputs
   , module Echidna.Internal.Runner
   ) where
 
 import Control.Concurrent.MVar    (MVar, takeMVar, putMVar)
+import Control.Concurrent.Chan    (Chan, writeChan, readChan)
 import Control.Lens               ((^.), (.=), use)
-import Control.Monad              (forM_, replicateM)
+import Control.Monad              (forM_, replicateM, forever)
 import Control.Monad.Catch        (MonadCatch)
 import Control.Monad.IO.Class     (MonadIO, liftIO)
 import Control.Monad.State.Strict (MonadState, StateT, evalState, evalStateT, execState, runState)
@@ -23,7 +25,7 @@ import Control.Monad.Reader       (MonadReader, ReaderT, runReaderT, ask)
 import Data.IORef                 (IORef, modifyIORef', newIORef, readIORef)
 import Data.List                  (intercalate)
 import Data.Maybe                 (listToMaybe)
-import Data.MultiSet              (MultiSet, insert, union)
+import Data.Set                   (Set, insert, union, size, difference)
 import Data.Text                  (Text)
 import Data.Typeable              (Typeable)
 import Data.Vector                (fromList)
@@ -44,8 +46,34 @@ import EVM.Exec     (exec)
 import Echidna.ABI (SolCall, SolSignature, displayAbiCall, encodeSig, genInteractions)
 import Echidna.Internal.Runner
 
-type Coverage = MultiSet Int
+  
+data ExecEnv = ExecEnv
+  { coverage :: Coverage --Cumulative coverage accross all tests
+  , recent   :: [Int] -- Coverage for previous 10 tests
+  , avg      :: Float -- Moving average of coverage for prior tests
+  }  
+
+type Coverage    = Set Int      
 type CoverageRef = IORef Coverage
+
+--This doesn't do anything right now... but it will be a thread that processes inputs
+--deemed interesting from a channel
+mutateGoodInputs :: Chan [SolSignature] -> IO ()
+mutateGoodInputs c = forever $ do
+  input <- readChan c
+  return ()
+
+updateExecEnv :: ExecEnv -> Coverage -> ExecEnv
+updateExecEnv ExecEnv{coverage = c, recent = r, avg = a} cov = ExecEnv { coverage = c', recent = r', avg = a'}
+  where c' = union c cov
+        r' = (size cov) : (take 5 r)
+        a' = fromIntegral $ (sum r) `div` (length r)
+
+-- Use to determine if input is worth handing off to the "good input" pool for mutation
+-- Currently just sees if current input has higher than avg new pc's
+isInteresting :: ExecEnv -> Coverage -> Bool
+isInteresting e c = if fromIntegral (size new) > avg e then True else False
+  where new = coverage e `difference` c
 
 execCallUsing :: MonadState VM m => m VMResult -> SolCall -> m VMResult
 execCallUsing m (t,vs) = cleanUp >> (state . calldata .= cd >> m) where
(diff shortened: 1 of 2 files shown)


lib/Echidna/Exec.hs
```diff
@@ -41,15 +43,45 @@ import EVM.ABI      (AbiValue(..), abiCalldata, abiValueType, encodeAbiValue)
...
+      vs' <- V.unsafeThaw vs
+      M.write vs' i emptyCovInfo
+      res <- V.unsafeFreeze vs'
+      setCover res c tot newCalls
+  
 
 execCallUsing :: MonadState VM m => m VMResult -> SolCall -> m VMResult
 execCallUsing m (t,vs) = cleanUp >> (state . calldata .= cd >> m) where
-  cd = B . abiCalldata (encodeSig t $ abiValueType <$> vs) $ fromList vs
+  cd = B . abiCalldata (encodeSig t $ abiValueType <$> vs) $ V.fromList vs
```
japesinator: Instead of going to/from lists, can we just change everything to use Vectors? I don't think we ever actually want linked lists for reasons other than they're easy. Alternatively, stuff over arbitrary Foldables is even better if that's viable for some functions

lib/Echidna/Exec.hs
```diff
@@ -106,62 +140,65 @@ instance HTraversable VMAction where
...
+eCommandUsing gen ex p = Command (\_ -> pure $ Call <$> gen) ex
   [ Ensure $ \_ (VMState v) _ _ -> assert $ p v
   , Update $ \(VMState v) (Call c) _ -> VMState $ execState (execCall c) v
   ]
   
 
-eCommand :: (MonadGen n, MonadTest m) => (VM -> Bool) -> [SolSignature] -> Command n m VMState
-eCommand = eCommandUsing (\_ -> pure ())
+eCommand :: (MonadGen n, MonadTest m) => n SolCall -> (VM -> Bool) -> Command n m VMState
+eCommand x y = eCommandUsing x (\_ -> pure ()) y
```
japesinator: Nit, but may as well eta reduce

lib/Echidna/Exec.hs
```diff
@@ -106,62 +140,65 @@ instance HTraversable VMAction where
...
-    a            <- evalStateT s v
-    threadCov    <- liftIO $ readIORef threadCovRef
-    globalCov    <- liftIO $ takeMVar globalCovRef
-    liftIO $ putMVar globalCovRef (union threadCov globalCov)
-    return a
+ePropertySeqCoverage calls cov p ts v = ePropertyUsing (eCommandCoverage calls p ts) writeCoverage v
+  where
+    writeCoverage :: MonadIO m => ReaderT CoverageRef (StateT VM m) a -> m a
+    writeCoverage m = do
+      threadCovRef <- liftIO $ newIORef (("",[]),empty)
```
japesinator: we defined this on line 56 as emptyCoverageInfo, let's either delete that or use it here

src/Main.hs
```diff
@@ -17,13 +18,38 @@ import Hedgehog.Internal.Property (GroupName(..), PropertyName(..))
...
+  filepath:n:args -> do
+    (v,a,ts)   <- loadSolidity filepath $ pack <$> listToMaybe args
+    tests      <- mapM (\t -> return . ((,) t) =<< newMVar []) ts
+    let prop (cov,t,mvar) = (PropertyName $ show t
+                            , ePropertySeqCoverage cov mvar (flip checkETest t) a v 10
+                            )
+
+    let epochs = read n :: Int
+    forM_ [1..epochs] $ \i -> do
+      putStrLn $ "Beginning Epoch " ++ show i
```
japesinator: Is this info helpful to CLI users?

src/Main.hs
```diff
@@ -17,13 +18,38 @@ import Hedgehog.Internal.Property (GroupName(..), PropertyName(..))
...
+      xs <- forM tests $ \(x,y) -> do
+        cov <- readMVar y
+        lastGen <- getCover cov
+        modifyMVar_ y (\_ -> pure [])
```
japesinator: Is this better than `swapMVar` while discarding the result?

lib/Echidna/Exec.hs
```diff
@@ -41,15 +43,45 @@ import EVM.ABI      (AbiValue(..), abiCalldata, abiValueType, encodeAbiValue)
...
-type CoverageRef = IORef Coverage
+
+--------------------------------------------------------------------
+-- COVERAGE HANDLING
+
+type CoverageInfo = (SolCall, Set Int)
+type CoverageRef  = IORef CoverageInfo
+
+emptyCovInfo :: CoverageInfo
+emptyCovInfo = (("",[]),empty)
```
japesinator: This is just mempty

lib/Echidna/Exec.hs
```diff
@@ -106,62 +140,65 @@ instance HTraversable VMAction where
...
+eCommand x y = eCommandUsing x (\_ -> pure ()) y
 
 
 eCommandCoverage :: (MonadGen n, MonadTest m, MonadState VM m, MonadReader CoverageRef m, MonadIO m)
-                 => (VM -> Bool) -> [SolSignature] -> Command n m VMState
-eCommandCoverage = eCommandUsing $ \(Call c) -> execCallCoverage c
+                 => [SolCall] -> (VM -> Bool) -> [SolSignature] -> [Command n m VMState]
+eCommandCoverage cov p ts = case cov of
+  [] -> [eCommandUsing (genInteractions ts) (\(Call c) -> execCallCoverage c) p]
+  xs -> map (\x -> eCommandUsing (mutateCall x) (\(Call c) -> execCallCoverage c) p) xs
```
japesinator: If I'm reading this correctly, we'll stop synthesizing new inputs after the first epoch. Can we instead Gen.choice between mutation and generation?

lib/Echidna/Exec.hs
```diff
@@ -41,15 +43,45 @@ import EVM.ABI      (AbiValue(..), abiCalldata, abiValueType, encodeAbiValue)
...
+import Echidna.ABI (SolCall, SolSignature, displayAbiCall, encodeSig, genInteractions, mutateCall)
+import Echidna.Internal.Runner 
 
-type Coverage = MultiSet Int
-type CoverageRef = IORef Coverage
+
+--------------------------------------------------------------------
+-- COVERAGE HANDLING
+
+type CoverageInfo = (SolCall, Set Int)
```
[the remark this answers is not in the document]
blperez01: Lets put this in a separate PR where we do a more serious rewrite of the coverage data type

lib/Echidna/Exec.hs
japesinator: Update: we push/pop a lot, so we may actually want linked lists instead of vectors here!

blperez01: Here we actually can't use a linked list because `abiCallData` has type sig `Text -> Vector AbiValue -> ByteString`

japesinator: Sorry, the above was unclear. I meant just that my original suggestion was likely unnecessary

src/Main.hs
```diff
@@ -32,22 +35,49 @@ options = Options
...
+
+  case n of
+    -- RUN WITHOUT COVERAGE
+    Nothing -> do
+      let prop t = (PropertyName $ show t
+                   , ePropertySeq (flip checkETest t) a v 10
+                   )
+
+      _ <- checkParallel . Group (GroupName f) $ map prop ts
+      return ()
```
japesinator: What is the purpose of the `return ()` at the end of both of these cases? iirc, it was originally there b/c `checkParallel` returns an `IO Bool` but I'm not sure we need it anymore

src/Main.hs
```diff
@@ -32,22 +35,49 @@ options = Options
...
+
+    -- RUN WITH COVERAGE
+    Just epochs -> do
+      tests <- mapM (\t -> return . (t,) =<< newMVar []) ts
+      let prop (cov,t,mvar) = (PropertyName $ show t
+                              , ePropertySeqCoverage cov mvar (flip checkETest t) a v 10
+                              )
+
+      forM_ [1..epochs] $ \_ -> do
```
japesinator: `replicateM`?

src/Main.hs
blperez01: We need the `return ()` in the non-coverage case because `checkParallel` returns `IO Bool`, but the second one is unnecessary.

PR #49, opened 2018-05-15
Improved fix for openFile

ggrieco-tob:
This patch allows to run echidna from any directory without a crash.


japesinator:
we shouldn't need anything like this in the long run, but this is a clear improvement over what we have now. lgtm!

Issue #46, opened 2018-05-11
Inconsistent test results in a simple contract


ggrieco-tob:
In this simple contract:

```solidity
contract TEST {
  function f(bytes1 x) returns (bool) {
        return true;
  }

  function echidna_true() returns (bool) {
    return true;
  }
}
```
the echidna_true test cannot fail. However the last revision of Echidna reports a failed test when **it runs more than one epoch**. For instance:

```bash
$ echidna-test test.sol --epochs 10
━━━ test.sol ━━━
  ✓ "echidna_true" passed 1000 tests.
  ✓ 1 succeeded.
━━━ test.sol ━━━
  ✗ "echidna_true" failed after 14 tests and 1 shrink.
  
      │ Call sequence: f("");
  
  ✗ 1 failed.
━━━ test.sol ━━━
  ⚐ "echidna_true" gave up after 100 discards, passed 762 tests.
  ⚐ 1 gave up.
━━━ test.sol ━━━
  ✗ "echidna_true" failed after 20 tests and 2 shrinks.
  
      │ Call sequence: f("");
  
  ✗ 1 failed.
  ...
```

blperez01:
Tried @japesinator's suggestion of using `Alternative` in `eCommandCoverage`:

```Haskell
eCommandCoverage :: (Alternative n, MonadGen n, MonadTest m, MonadState VM m, MonadReader CoverageRef m, MonadIO m)
                 => [SolCall] -> (VM -> Bool) -> [SolSignature] -> [Command n m VMState]
eCommandCoverage cov p ts = case cov of
  [] -> [eCommandUsing (genInteractions ts) (\(Call c) -> execCallCoverage c) p]
  xs -> map (\x -> eCommandUsing (choice [mutateCall x <|> genInteractions ts, genInteractions ts])
              (\(Call c) -> execCallCoverage c) p) xs
```
but still got the same error above. Exploring other potential fixes.

blperez01:
Update: this is actually an issue with `changeBS` in ABI.hs. Currently fixing.

PR #53, opened 2018-05-24
Fixed mutateValue

blperez01:
Fixed issue where mutateValue treated static bytestrings like dynamic bytestrings. This resolves Issue #46. Also modified off by one error in Main.hs where it used too few epochs.

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -185,9 +185,6 @@ dropBS b = Gen.choice [ BS.drop <$> Gen.element [1..BS.length b]   <*> pure b
 changeDynamicBS :: MonadGen m => ByteString -> m ByteString
 changeDynamicBS b = Gen.choice $ [changeChar, addBS, dropBS] <&> ($ b)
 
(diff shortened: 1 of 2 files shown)


lib/Echidna/ABI.hs
```diff
@@ -182,8 +182,11 @@ dropBS b = Gen.choice [ BS.drop <$> Gen.element [1..BS.length b]   <*> pure b
                       , BS.take <$> Gen.element [0..BS.length b-1] <*> pure b
                       ]
 
-changeBS :: MonadGen m => ByteString -> m ByteString
-changeBS b = Gen.choice $ [changeChar, addBS, dropBS] <&> ($ b)
+changeDynamicBS :: MonadGen m => ByteString -> m ByteString
+changeDynamicBS b = Gen.choice $ [changeChar, addBS, dropBS] <&> ($ b)
+
+changeStaticBS :: MonadGen m => ByteString -> m ByteString
+changeStaticBS b = changeChar b
```
japesinator: I'm more inclined to just use `changeChar` rather than give it a second name here

PR #58, opened 2018-06-02
Dev config

blperez01:


lib/Echidna/Config.hs
```diff
@@ -0,0 +1,42 @@
...
+  , _range :: Int
+  , _gasLimit :: W256 }
+  deriving (Show, Generic)
+
+makeLenses ''Config
+
+instance FromJSON Config
+
+defaultConfig :: Config
+defaultConfig = Config
```
japesinator: Do we document what these fields mean anywhere? I'm not sure the names would make that obvious to a new user

src/Main.hs
```diff
@@ -32,51 +35,51 @@ options = Options
...
-
-        checkParallel . Group (GroupName f) $ map prop xs
+  -- Read cmd line options and load config
+  (Options file contract configFile) <- execParser opts
+  config <- maybe (pure defaultConfig) parseConfig configFile
+
+  -- Load solidity contract and get VM
+  (v,a,ts) <- runReaderT (loadSolidity file (pack <$> contract)) config
+
+  if (config ^. epochs) <= 0
```
japesinator: I think I'm inclined to be a bit less clever here. "Set the epochs variable to zero to disable coverage-guidance (and thus, go significantly faster)" doesn't feel like the natural way to use echidna

src/Main.hs
```diff
@@ -32,51 +35,51 @@ options = Options
...
+    -- Run without coverage
+    then do
+    let prop t = (PropertyName $ show t
+                 , runReader (ePropertySeq (flip checkETest t) a v) config
```
japesinator: why not move this out a bit? I think my favorite way to do this would be to parse the config, then do everything after that in a reader with it accessible

examples/revert/Revert.hs
```diff
@@ -1,18 +1,20 @@
...
 import EVM (VM, VMResult(..), Error( Revert ))
 
+import Echidna.Config (defaultConfig)
 import Echidna.Exec
 import Echidna.Solidity
 
 main :: IO ()
-main = do (v,a,ts) <- loadSolidity "solidity/revert.sol" Nothing Nothing
-          let prop t = (PropertyName $ show t, ePropertySeq (`checkRTest` t) a v 10)
+main = do (v,a,ts) <- runReaderT (loadSolidity "solidity/revert.sol" Nothing) defaultConfig
```
japesinator: Can we add some `withDefaultConf :: MonadReader Config m => m a -> a` to account for this pattern of usage, then just put it left of `do` here?

src/Main.hs
```diff
@@ -32,51 +35,51 @@ options = Options
       <*> optional ( argument str
           ( metavar "CONTRACT"
          <> help "Contract inside of file to analyze" ))
-      <*> optional ( strOption
-          ( long "solc-args"
-         <> help "Optional solidity compiler arguments" ))
-      <*> optional ( option auto
-          ( long "epochs"
-         <> help "Optional number of epochs to run coverage guidance" ))
+      <*> optional ( option str
```
japesinator: I'm not sure how I feel about file-based conf being the only option. Can we keep the CLI options?

blperez01:
I integrated the proposed changes in Main.hs and Exec.hs. We might want to merge before a larger config refactor?

PR #67, opened 2018-06-27
Dev evm config

blperez01:
Added functionality for specifying the contract address and `sender` in the EVM during Echidna tests. Also added functionality for whitelisting a list of addresses that will be chosen from during test generation. 

It may be good to have a list of potential `sender`'s instead of a fixed `sender`. Another potential improvement would be to let users simply specify the number of unique addresses they want Echidna to use in test generation (which would make life easier if users wanted 50 addresses to be used). We could even have Echidna print the list of addresses it generated. Both of these would be minor fixes, though.

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE ConstraintKinds, FlexibleContexts, RankNTypes, TupleSections, TypeFamilies #-}
+{-# LANGUAGE ConstraintKinds, FlexibleContexts, LambdaCase, RankNTypes, TupleSections, TypeFamilies #-}
 
 module Echidna.ABI (
     SolCall
@@ -26,7 +26,7 @@ module Echidna.ABI (
 
 import Control.Lens          ((<&>), (&), view)
 import Control.Monad         (join, liftM2, replicateM)
-import Control.Monad.Reader  (MonadReader, asks, runReaderT)
+import Control.Monad.Reader  (MonadReader)
 import Data.Bool             (bool)
 import Data.DoubleWord       (Word128(..), Word160(..))
 import Data.Monoid           ((<>))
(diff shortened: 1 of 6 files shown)


lib/Echidna/ABI.hs
```diff
@@ -109,26 +113,26 @@ genAbiType = Gen.choice [ pure AbiBytesDynamicType
                         , genStaticAbiType
                         ]
 
-genVecOfType :: MonadGen m => AbiType -> Range Int -> m (Vector AbiValue)
+genVecOfType :: (MonadReader Config m, MonadGen m) => AbiType -> Range Int -> m (Vector AbiValue)
 genVecOfType t r = fmap fromList . Gen.list r $ case t of
   AbiUIntType    n    -> genAbiUInt n
   AbiIntType     n    -> genAbiInt n
-  AbiAddressType      -> genAbiAddress
+  AbiAddressType      -> genAbiAddress =<< asks (view addrList)
```
japesinator: `asks` is superfluous here, ref. specification in http://hackage.haskell.org/package/lens-4.17/docs/Control-Lens-Getter.html#v:view

japesinator: (also in other usage across this file)

lib/Echidna/ABI.hs
```diff
@@ -221,9 +226,9 @@ mutateValue (AbiArray s t a) =
 changeOrId :: (Traversable t, MonadGen m) => (a -> m a) -> t a -> m (t a)
 changeOrId f = mapM $ (Gen.element [f, pure] >>=) . (&)
 
-mutateCall :: MonadGen m => SolCall -> m SolCall
-mutateCall (t, vs) = (t,) <$> changeOrId mutateValue vs
+mutateCall :: MonadGen m => Config -> SolCall -> m SolCall
```
japesinator: let's move this back to a `MonadReader` and use `data-has` to keep this general, then everything in this file has the same style

lib/Echidna/Exec.hs
```diff
@@ -159,10 +160,10 @@ eCommand = flip eCommandUsing (\ _ -> pure ())
 
 
 eCommandCoverage :: (MonadGen n, MonadTest m, MonadState VM m, MonadReader CoverageRef m, MonadIO m)
-                 => [SolCall] -> (VM -> Bool) -> [SolSignature] -> [Command n m VMState]
-eCommandCoverage cov p ts = case cov of
-  [] -> [eCommandUsing (genInteractions ts) (\(Call c) -> execCallCoverage c) p]
-  xs -> map (\x -> eCommandUsing (choice [mutateCall x, genInteractions ts])
+                 => [SolCall] -> (VM -> Bool) -> [SolSignature] -> Config -> [Command n m VMState]
```
japesinator: let's also move this to a `MonadReader` as above

solidity/config.yaml
```diff
@@ -1,4 +1,16 @@
...
+#_solcArgs:
+#Choose the number of epochs to use in coverage-guided testing
 _epochs: 2
+#Set the gas limit for each test
 _gasLimit: 0xfffff
+#Number of tests that will run for each property
 _testLimit: 10000
-_range: 10
\ No newline at end of file
+#Honestly not sure what this is
```
japesinator: https://github.com/trailofbits/echidna/blob/master/lib/Echidna/Exec.hs#L161 do you want to rename? if it's unclear to you, might be to users as well

lib/Echidna/ABI.hs
```diff
@@ -68,8 +71,9 @@ encodeSig n ts = n <> "(" <> T.intercalate "," (map abiTypeSolidity ts) <> ")"
 genSize :: MonadGen m => m Int
 genSize = (8 *) <$> Gen.enum 1 32
 
-genAbiAddress :: MonadGen m => m AbiValue
-genAbiAddress = let w64 = Gen.word64 $ constant minBound maxBound in
+genAbiAddress :: MonadGen m => Maybe [Addr] -> m AbiValue
```
japesinator: why not just move this into a `MonadReader` and we can avoid passing it a parameter at each use site?

lib/Echidna/Exec.hs
japesinator: on second thought, fine having this return a plain Command (I think!)

PR #68, opened 2018-07-05
Improved and standardized properties 

ggrieco-tob:
This PR adds contains two main changes:
* The properties should use the `deepstate_` prefix
* The user can specify if the properties should return true, false or revert using the propertyType config option.

### DIFF
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -5,6 +5,7 @@ module Echidna.Exec (
     TestFunction
   , checkETest
   , checkRTest
+  , checkERTest
   , CoverageInfo
   , CoverageRef
   , eCommand
@@ -132,8 +133,15 @@ checkETest b v t = case evalState (execCall (t, [])) v of
 
 checkRTest :: TestFunction
 checkRTest v t = case evalState (execCall (t, [])) v of
-  (VMFailure Revert) -> False
-  _                  -> True
+  (VMFailure Revert) -> True
+  _                  -> False
+
+checkERTest :: TestFunction
+checkERTest v t = case evalState (execCall (t, [])) v of
+  (VMSuccess (B s))  -> s == encodeAbiValue (AbiBool False)
+  (VMFailure Revert) -> True
+  _                  -> False
+
 
 
 newtype VMState (v :: * -> *) =
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -5,6 +5,7 @@ module Echidna.Exec (
     checkTest
   , checkBoolExpTest
   , checkRevertTest
+  , checkTrueOrRevertTest
   , checkFalseOrRevertTest
   , CoverageInfo
   , CoverageRef
@@ -136,6 +137,12 @@ checkBoolExpTest b v t = case evalState (execCall (t, [])) v of
 
 checkRevertTest :: VM -> Text -> Bool
 checkRevertTest v t = case evalState (execCall (t, [])) v of
+  (VMFailure Revert) -> True
+  _                  -> False
+
+checkTrueOrRevertTest :: VM -> Text -> Bool
+checkTrueOrRevertTest v t = case evalState (execCall (t, [])) v of
+  (VMSuccess (B s))  -> s == encodeAbiValue (AbiBool True)
   (VMFailure Revert) -> True
   _                  -> False
 
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -2,7 +2,9 @@
     LambdaCase, StrictData, TemplateHaskell #-}
 
 module Echidna.Exec (
-    checkETest
+    TestFunction
+  , checkETest
+  , checkRTest
   , CoverageInfo
   , CoverageRef
   , eCommand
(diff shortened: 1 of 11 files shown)


japesinator:
Can we merge after #70 gets merged? I'll leave a bit of feedback now though

lib/Echidna/Exec.hs
```diff
@@ -2,7 +2,10 @@
     LambdaCase, StrictData, TemplateHaskell #-}
 
 module Echidna.Exec (
-    checkETest
+    TestFunction
+  , checkETest
```
japesinator: can we make these names a bit more clear? I think in retrospect `checkETest` was already too ambiguous, and I think this compounds the problem. I'd also probably prefer to merge them into one function that takes a `TestFunction` as a parameter, but don't feel quite as strongly there.

lib/Echidna/Exec.hs
```diff
@@ -121,11 +124,25 @@ fuzz l n ts v p = do
     where run cs = p $ execState (forM_ cs execCall) v
 
 
-checkETest :: VM -> Text -> Bool
-checkETest v t = case evalState (execCall (t, [])) v of
-  VMSuccess (B s) -> s == encodeAbiValue (AbiBool True)
+type TestFunction = VM -> Text -> Bool
```
japesinator: I don't know if I believe this alias makes things clearer, especially if we merge these functions as suggested above

lib/Echidna/Property.hs
```diff
@@ -0,0 +1,19 @@
+{-# LANGUAGE DeriveGeneric, FlexibleContexts, TemplateHaskell #-}
+
+module Echidna.Property where
+
+import GHC.Generics
+import qualified Data.HashMap.Lazy as HML        ( lookup )
+import Data.Yaml
```
japesinator: do we use this import?

lib/Echidna/Property.hs
```diff
@@ -0,0 +1,19 @@
...
+
+data PropertyType = ShouldReturnTrue | ShouldReturnFalse | ShouldRevert | ShouldReturnFalseRevert
+  deriving (Show, Generic)
```
japesinator: do we use this Generic instance?

lib/Echidna/Property.hs
```diff
@@ -0,0 +1,19 @@
+{-# LANGUAGE DeriveGeneric, FlexibleContexts, TemplateHaskell #-}
```
japesinator: Where do we use these extensions?

lib/Echidna/Solidity.hs
```diff
@@ -99,7 +99,7 @@ loadSolidity filePath selectedContract = do
                   loadContract (vm ^. state . contract)
         loaded = execState load $ execState (replaceCodeOfSelf bc) vm
         abi = map (liftM2 (,) (view methodName) (map snd . view methodInputs)) . toList $ c ^. abiMap
-        (tests, funs) = partition (isPrefixOf "echidna_" . fst) abi
+        (tests, funs) = partition (isPrefixOf "deepstate_" . fst) abi
```
japesinator: I'm not sure how I feel about this prefix change. Seems like it would break _a lot_ of existing code. I'd probably prefer to keep it as-is unless there's more justification for such a breaking change

solidity/revert.sol
```diff
@@ -7,7 +7,7 @@ contract C {
     state = x;
   }
 
-  function echidna_revert() public returns (bool) {
+  function solidity_revert() public returns (bool) {
```
japesinator: was this supposed to change to this new prefix?

ggrieco-tob: My mistake!

lib/Echidna/Exec.hs
ggrieco-tob: This alias was created to avoid exporting VM and Bool into the Main module. If you don't care about them, I can remove it.

lib/Echidna/Solidity.hs
ggrieco-tob: I updated the prefix to standardize the property names following the configuration files used in zoo (for instance, [this one](https://github.com/trailofbits/smart-contracts-testing/blob/master/zoo/tokens/vulnerableToken/contracts/configManticore-WithOwner.yaml#L12))

japesinator: Let's consider moving this to the config then? I'd really like to avoid breaking peoples' existing tests, how would you feel about some kinda "`test_prefix`" field

lib/Echidna/Exec.hs
```diff
@@ -120,12 +123,28 @@ fuzz l n ts v p = do
...
+checkBoolExpTest b v t = case evalState (execCall (t, [])) v of
+  VMSuccess (B s) -> s == encodeAbiValue (AbiBool b)
   _               -> False
 
+checkRevertTest :: VM -> Text -> Bool
+checkRevertTest v t = case evalState (execCall (t, [])) v of
+  (VMFailure Revert) -> True
+  _                  -> False
+
+checkFalseOrRevertTest :: VM -> Text -> Bool
```
japesinator: Why is there a `checkFalseOrRevert` but no `checkTrueOrRevert`?

ggrieco-tob: checkFalseOrRevert is useful to implement [the zoo property standard](https://github.com/trailofbits/smart-contracts-testing/tree/master/zoo/properties#return-value).

japesinator: Let's definitely keep it then, but maybe add a checkTrueOrRevert for completeness?

japesinator:
lgtm!

PR #69, opened 2018-07-05
Added a parameter to limit the shrinking

ggrieco-tob:
Sometimes it is useful to limit or disable shrinking. This PR adds a new global option to control the shrinkTest in the hedgehog test used by Echidna.


japesinator:
lgtm!

PR #87, opened 2018-07-19
Can now choose to output only JSON corresponding to test output

blperez01:


lib/Echidna/Solidity.hs
```diff
@@ -70,11 +70,14 @@ readContracts filePath = do
 -- | reads either the first contract found or the contract named `selectedContractName` within the solidity file at `filepath`
 readContract :: (MonadIO m, MonadThrow m, MonadReader Config m) => FilePath -> Maybe Text -> m SolcContract
 readContract filePath selectedContractName = do
+    config <- ask
```
japesinator: let's just do `useJSON <- view outputJson`

PR #90, opened 2018-07-21
make testLimit work as intended

### LINKED ISSUE #85: testLimit property
The `testLimit` property for perprop should represent the number of transactions run before checking a property, not the number of tests echidna runs for that property.

japesinator:
Fixes #85


ggrieco-tob:
I just realized we included `perprop/.Main.hs.swp` in this PR. :-(

PR #92, opened 2018-07-23
Dev perprop cmdline

blperez01:


perprop/Main.hs
```diff
@@ -26,6 +26,38 @@ import Echidna.Exec
...
+import Options.Applicative as O
+
+
+-- Cmd line parser
+-- {{{
+
+data Options = Options
+  { filePath         :: FilePath
+  , configFilepath   :: FilePath
+  , selectedContract :: Maybe String
```
japesinator: why not just have this be Text if that's what it is when used?

blperez01: I guess I was just echoing the design in 
https://github.com/trailofbits/echidna/blob/a952fbb7988e37a00a2aa4fcd76c3c386ba337d8/src/Main.hs#L25-L30
Is there a compelling reason for `selectedContract` to be a `String` there?

japesinator: I didn't write that code, and must have just missed that in my review. I don't feel too strongly either way, do what makes sense to you

Issue #97, opened 2018-07-30
When there is no methods to execute, any test will fail


ggrieco-tob:
If a contract does not define any (public) method to execute, like this one:

```solidity
contract C {
  function echidna_test() returns (bool) {
    return true;
  }
}
```
then echidna will fail to execute the test:

```
━━━ test.sol ━━━
  ✗ "echidna_test" failed after 1 test.
  
    
  
  ✗ 1 failed.
```

It is better to warn the user about this issue and abort the execution.

PR #108, opened 2018-08-04
added case to fix bug #97

er5:
It could be handled by NoFuncs, but in this way there is a difference between the empty contract and a contract with only tests. Perhaps the variable funcs should be renamed.


ggrieco-tob:
Don't forget to sign the Contributor License Agreement.

Issue #106, opened 2018-08-02
Properties randomly fail when coverage is enabled


ggrieco-tob:
If you run `echidna-test` with a solidity contract, like this one:

```solidity
contract C {
  function f(uint[] x) returns (uint) {
    return x[0];
  }
  function echidna_true() returns (bool) {
    return true;
  }
}
```

using this config file:

```yaml
testLimit: 100
epochs: 10
range: 1
printCoverage: true
```

It will cause the `echidna_true` property to fail from time to time. For instance:

```bash
$ echidna-test bug.sol --config test.yaml 
...
"Analyzing contract: bug.sol:C"
━━━ bug.sol ━━━
  ✓ "echidna_true" passed 100 tests.
  ✓ 1 succeeded.
━━━ bug.sol ━━━
  ✓ "echidna_true" passed 100 tests.
  ✓ 1 succeeded.
━━━ bug.sol ━━━
  ✓ "echidna_true" passed 100 tests.
  ✓ 1 succeeded.
━━━ bug.sol ━━━
  ✗ "echidna_true" failed after 28 tests.
  
    
  
  ✗ 1 failed.
━━━ bug.sol ━━━
  ✗ "echidna_true" failed after 26 tests.
  
    
  
  ✗ 1 failed.
━━━ bug.sol ━━━
  ✗ "echidna_true" failed after 81 tests.
  
    
  
  ✗ 1 failed.
━━━ bug.sol ━━━
...
```

PR #109, opened 2018-08-05
fixed function changeList (bug #106)

er5:


ggrieco-tob:
Thanks a lot for this PR! Can you explain how this patch fixes #106?

er5:
When coverage is enabled, after the first epoch, mutations are tried. For mutating `AbiArrayDynamic` (in `mutateValue`), the function `changeList` is used. In the old version of `changeList`, if `l` was empty, then `Gen.element l` would trigger an error and the test would fail.

PR #115, opened 2018-08-07
Improved ABI value generation to include 'magic' constants from AFL

ggrieco-tob:


japesinator:
Generation is currently undergoing a total refactor, we'll be able to just put magic constants in a dictionary

PR #122, opened 2018-08-10
Workaround to use senders in perprop

ggrieco-tob:


perprop/Main.hs
```diff
@@ -113,7 +113,7 @@ readConf :: FilePath -> IO (Maybe (Config, [Property]))
 readConf f = decodeEither <$> BS.readFile f >>= \case
   Left e -> putStrLn ("couldn't parse config, " ++ e) >> pure Nothing
   Right (PerPropConf t s p) -> pure . Just . (,p) $
-    defaultConfig & addrList ?~ (view address <$> s) & range .~ t & epochs .~ 1 & outputJson .~ True
+    defaultConfig & addrList ?~ (view address <$> s) & range .~ t & Echidna.Config.sender .~ (head $ view address <$> s) & epochs .~ 1 & outputJson .~ True
```
japesinator: can we import `sender` unqualified + avoid the use of `head`?

ggrieco-tob: I don't know how to use `sender` unqualified:

```
    Ambiguous occurrence ‘sender’
    It could refer to either ‘Echidna.Config.sender’,
                             imported from ‘Echidna.Config’ at perprop/Main.hs:27:1-21
                          or ‘Main.sender’, defined at perprop/Main.hs:90:1
```

To avoid the use of `head` (or any arbitrary way of selecting an element), we should decide which address from the sender list we should use (and this will be fixed during all the calls).

ggrieco-tob:
Superseded by #128

PR #123, opened 2018-08-10
Save corpus of unique transactions according to coverage

ggrieco-tob:
This patch enables Echidna to explore complex contract and save the unique transactions that it discovers acording to the coverage information in the "txs" directory. For instance, using this contract:

```solidity
contract C {
  bool state = true;
  function f(uint x, uint y, uint z) {
    require(x == 12);
    require(y == 8);
    require(z == 0);
    state = false;
    return;
  }

  function echidna_state() returns (bool) {
    return state;
  }
}
```
and this config:

```yaml
testLimit: 1000
epochs: 50
range: 1
printCoverage: true
```
Echidna produces the following files:

```bash
$ tail -f txs/*
==> txs/2fad4ca859d2beea328b4d89a3f6b3a0159d4ab5e2daac00b58e95101a9949b6 <==
f(83006518593401026877002066252139153289669799003303892805020259071083755513852,111789781830202758384157176293928937901288938685289627343970582402723635776859,85220154112604218682235745130937888514668036279708809103845409540750944322327)

==> txs/781ace0f6e8283543f318a3628ba5fd4ad6e06fe74efc9c2bda526b9a1c21e47 <==
f(12,8,19)

==> txs/ca3d96852f17358131206a92f2d70e252c8d906fab361180c0a01955b1b3fb59 <==
f(12,3,18)
```

Several Echidna instances can be run in parallel to accumulate more test cases in the same directory (however, they are not going to share information). Also, other tools could parse the transactions discovered and continue the exploration (e.g using symbolic execution).

### DIFF
--- a/lib/Echidna/Coverage.hs
+++ b/lib/Echidna/Coverage.hs
@@ -29,8 +29,8 @@ import Data.Set                   (Set, insert, size)
 import Data.Vector                (Vector, fromList)
 import Data.Vector.Generic        (maximumBy)
 import GHC.Generics
-import Crypto.Hash.SHA256         (hashlazy, hash)
-import System.Directory           (doesFileExist)
+import Crypto.Hash.SHA256         (hash)
+import System.Directory           (doesFileExist, createDirectoryIfMissing)
 
 import qualified Control.Monad.State.Strict as S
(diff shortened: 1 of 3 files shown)


ggrieco-tob:
Superseded by #133

PR #126, opened 2018-08-14
Update hevm using nix

ggrieco-tob:
This PR implements the use of nix to upgrade hevm to the last revision (it's tagged as version 0.15) as well as install every dependency of echidna.
This PR disables the CI tests in OSX since they take a longer than 50 minutes to run: some nix packages are missing from the binary cache (e.g. python 3.6). OSX is still supported, but we disabled the CI tests to avoid timeouts.

To restore the OSX tests, it's necessary to compile the missing packages and upload them to a [cachix](https://github.com/cachix/cachix) instance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ script:
  #- echo "$(stack ghc -- --version) [$(stack ghc -- --print-project-git-commit-id 2> /dev/null || echo '?')]"
  - GHC_OPTIONS="-Werror"
  - stack setup
- - stack install hedgehog
+ - stack install lens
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,43 +1,46 @@
...
 sudo: true
-
+  
 cache:
     directories:
         - $HOME/.ghc
         - $HOME/.cabal
         - $HOME/.stack
         - .stack-work
+        #- /nix
```
japesinator: Let's just delete this line instead of commenting it out

japesinator:
no more nix!

PR #131, opened 2018-09-04
Added checks to detect fatal errors of contracts

### LINKED ISSUE #26: fail on selfdestruct
i tried fuzzing a contract that anyone can "kill".
echidna didn't fail.
not sure if this is by design. i think it would make sense to have echidna fail if it manages to  `selfdestruct` a contract. (maybe make this the default but add an option to allow it.)

contract `self_destruct.sol`:

```solidity
contract SelfDestruct {
  function kill(address _to) public {
      selfdestruct(_to);
  }
  function echidna_alwaystrue() public pure returns (bool) {
      return true;
  }
}
```

run:

```
$ echidna-test self_destruct.sol
self_destruct.sol:1:1: Warning: Source file does not specify required compiler version!Consider adding "pragma solidity ^0.4.21;"
contract SelfDestruct {
^ (Relevant source part starts here and spans across multiple lines).
━━━ self_destruct.sol ━━━
  ✓ "echidna_alwaystrue" passed 10000 tests.
  ✓ 1 succeeded.
```

commit: 83b8fd5e8f57c825c98370efa802922da412dada

ggrieco-tob:
Some contracts fail with fatal errors like self destruct or they will run out of stack when they are running in Echidna. This patch will make any property to automatically fail in these edge cases. It should fix #26.

### DIFF
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -115,7 +115,7 @@ instance Show (VMState v) where
 
 newtype VMAction (v :: * -> *) = 
   Call SolCall
- 
+
 instance Show (VMAction v) where
   show (Call c) = displayAbiCall c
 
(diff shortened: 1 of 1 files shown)


ggrieco-tob:
Superseded by #133

PR #133, opened 2018-09-09
Complete Echidna refactor to remove the use of Hedgehog properties, simplify code and add missing features

ggrieco-tob:
This PR contains a complete refactoring of the Echidna code to remove the use of Hedgehog properties, simplify code and include some missing features. It was started during the [EthBerlin](https://ethberlin.com/) . This PR is **not** ready to be merged, it should be carefully reviewed before merging it. The main features of this refactoring are:

* Testing in parallel several properties, once we generated and executed transactions. If it founds a counter-example for a property, it removes that property from the list of properties to test and continue with the others. This allows a massive speed-up if we test a large amount of properties.
* Easy generation and usage of complete sequences of Ethereum transactions to fuzz, instead of individual calls. This allows to simplify and improve coverage guided testing.

Some interesting side effects of this refactoring: 

* Removed internal Hedgehog modules. We also no longer need MVars to update the coverage tracking.
* Should be easier to understand and modify by other Haskell programmers (since it no longer requires to understand how Hedgehog executes the unit tests).
* More control on the allocated memory, since we can easily call the garbage collection at any point of the generation, execution or shrinking.
* More control on the shrinking: we only implemented a simple global approach to minimize values, but we  should be able to remake the code from Hedgehog to restore delta debugging (I believe it's using that).
* More control on the output of echida: we can easily add or remove stuff since the hedgehog UI is no longer interfering. 

We also implemented some missing features:

* Events emitted  by the contracts are printed if Echidna founds  a counter-example for a property.
* Detection of killed and looping contracts. It implements a different type of property fail, a fatal failure. When something really bad happen with the the contract (e.g. it self-destructed), we can say that the checking of all properties will fail. So we stop the execution and show the sequences of transactions that produce the fatal failure.
* Support for sending transactions with random senders (specified in the config file).
* Support for sending ether to the constructor of a contract (specified in the config)
* Support for sending random amounts of ether to payable functions (specified in the config file).
* Support for avoiding using certain functions (specified in the config file).
* Automatic collection of distinctive traces for manual inspection (specified in the config file).

Please not that Hedgehog is not removed because the value generation and resizing is certainly useful, but we should be able to delete the internal runner module.


ggrieco-tob:
Good news:
  * echidna-test is more or less feature complete (even with JSON config support without extra code)
  * Test are [passing](https://travis-ci.org/trailofbits/echidna/jobs/430551708) for echidna-test (other binaries are disabled)
  * No more Internal modules :raised_hands: . 

This PR will also allow us to implement a different type of property fail: a **fatal failure**. When something really bad happen with the the contract (e.g. it self-destructed), we can say that the checking of all properties will fail. So we can stop the execution and show the sequences of transactions that produce the fatal failure. This is not implemented yet, but it can be easily added into the code.

japesinator:
Superseded by #157

PR #141, opened 2018-10-18
Fuzz contract constructor arguments

andremarianiello:
In order to have constructor information, I had to update to hevm-0.21. This required bumping the stack resolver to have a new enough version of the time package that was compatible with hevm, and removing some version constraint (though I didn't try very hard to have a tight bound, I was happy to let stack handle the version compatibilities).

I moved VM initialization from before TestableContract construction to after. Now it happens right before executing each generated sequence of SolCalls, and if the contract reverts, that test is skipped, just like what was being done when a random function call caused a revert.

### DIFF
--- a/.travis/install-secp256k1.sh
+++ b/.travis/install-secp256k1.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+set -eux
+
+gitRef="1086fda4c1975d0cad8d3cad96794a64ec12dca4"
+
+curl -LO "https://github.com/bitcoin-core/secp256k1/archive/$gitRef.zip"
+
+unzip "$gitRef.zip"
+
+cd "secp256k1-$gitRef"
+
+./autogen.sh
+# hevm needs reecovery module
+# don't install shared library so linker must used static
+# enable pic so static library can link against dynamic correctly
+./configure --enable-module-recovery --enable-shared=no --with-pic
+sudo make install
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,25 +1,28 @@
-resolver: lts-8.15
+resolver: lts-12.16
 
 packages:
 - '.'
 
 extra-deps:
-- git: https://github.com/dapphub/hevm.git
-  commit: 40b4ce9d38ed711ffd6065e697807100b3f0a3c3
-- brick-0.18
-- config-ini-0.2.1.1
-- data-clist-0.1.2.0
 - ghci-pretty-0.0.2
-- hedgehog-0.6
 - HSH-2.1.3
 - ipprint-0.6
-- megaparsec-6.4.0
-- optparse-applicative-0.13.2.0
-- parser-combinators-0.4.0
-- restless-git-0.5.0
+- restless-git-0.7
 - rosezipper-0.2
 - sr-extra-1.46.3.2
 - temporary-1.2.1.1
 - tree-view-0.5
-- vty-5.16
-- word-wrap-0.4.1
+- s-cargot-0.1.4.0
+- Unixutils-1.54.1
+- readline-1.0.3.0
+- hevm-0.21
+- multiset-0.3.4.1
+- text-format-0.3.2
+
+nix:
+  packages:
+  - bzip2
+  - readline
+  - zlib
+  - ncurses
+  - secp256k1
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 # Adapted from https://github.com/ChrisPenner/haskell-stack-travis-ci
-language: nix
+language: minimal
 sudo: true
 
 cache:
@@ -13,31 +13,28 @@ matrix:
   fast_finish: true
   include:
     # Add build targets here
-    - env: GHCVER=8.2.1 CACHE_NAME=8.2.1 BUILD_BINARY=1
-      compiler: ": #stack 8.2.1"
-    - env: GHCVER=8.2.1 CACHE_NAME=8.2.1-osx BUILD_BINARY=1
-           LDFLAGS=-L/usr/local/opt/readline/lib CFLAGS=-I/usr/local/opt/readline/include
+    - env: CACHE_NAME=linux BUILD_BINARY=1
+    - env: CACHE_NAME=osx BUILD_BINARY=1
+           LDFLAGS=-L/usr/local/opt/readline/lib
+           CFLAGS=-I/usr/local/opt/readline/include
+           STACK_OPTS="--extra-include-dirs=/usr/local/opt/readline/include --extra-lib-dirs=/usr/local/opt/readline/lib"
       os: osx
-      compiler: ": #stack 8.2.1"
 
 install:
  - unset CC
  - export PATH=$HOME/.local/bin:$PATH
  - ./.travis/install-ghr.sh
  - ./.travis/install-stack.sh
+ - ./.travis/install-secp256k1.sh
 
 script:
  - echo "$(stack ghc -- --version) [$(stack ghc -- --print-project-git-commit-id 2> /dev/null || echo '?')]"
  - GHC_OPTIONS="-Werror"
- - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then 
-   stack install readline --extra-include-dirs=/usr/local/opt/readline/include
-                          --extra-lib-dirs=/usr/local/opt/readline/lib;
-   fi
- - |
-   set -ex
+ - set -ex
+ - STACK_OPTS="$STACK_OPTS --extra-include-dirs=/usr/local/include --extra-lib-dirs=/usr/local/lib"
    # Run tests
-   stack --no-terminal test --ghc-options="$GHC_OPTIONS"
-   set +ex
+ - stack --no-terminal test --ghc-options="$GHC_OPTIONS" $STACK_OPTS
+ - set +ex
 
 after_success:
  - |
--- a/README.md
+++ b/README.md
@@ -13,28 +13,47 @@ It supports relatively sophisticated grammar-based fuzzing campaigns to falsify 
 ## Installation
 
 [stack](https://www.haskellstack.org/) is highly recommended to install echidna.
-If you are a particularly opinionated experienced Haskell user, cabal or hpack should work, but they are neither officially supported nor tested. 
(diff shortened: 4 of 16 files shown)


.travis/install-secp256k1.sh
```diff
@@ -0,0 +1,18 @@
...
+unzip "$gitRef.zip"
+
+cd "secp256k1-$gitRef"
+
+./autogen.sh
+# hevm needs reecovery module
+# don't install shared library so linker must used static
+# enable pic so static library can link against dynamic correctly
+./configure --enable-module-recovery --enable-shared=no --with-pic
+sudo make install
```
ggrieco-tob: README instructions should be updated explain how the installation process should be performed. Keep in mind that we don't want to recommend people to install libraries using `sudo make install` since it's very difficult to un-install them later (for travis scripts, it's ok, we don't care).

stack.yaml
```diff
@@ -1,25 +1,28 @@
...
-- vty-5.16
-- word-wrap-0.4.1
+- s-cargot-0.1.4.0
+- Unixutils-1.54.1
+- readline-1.0.3.0
+- hevm-0.21
+- multiset-0.3.4.1
+- text-format-0.3.2
+
+nix:
```
ggrieco-tob: I don't understand why this section section is necessary if the travis installation does not require NIX. Please remove this if it is useless. Also, update the README to make sure the instructions are correct.

gustavo-grieco (changes requested):
This is the first part of the review regarding the use of the new HEVM.

stack.yaml
andremarianiello: It is not necessary for CI, but it is necessary if someone wants to use stack's nix integration. It let's someone build echidna using `stack build --nix` instead of relying on system libraries that may or may not exist. In the case of libsecp256k1, it almost certainly would not exist. This list of packages also serves as documentation of what system deps echidna needs. I think we should keep this here so I can put `stack build --nix` as one option for building (I will also make sure that the steps for building without nix are documented as well). Does this sound reasonable to you?

.travis/install-secp256k1.sh
andremarianiello: I added a bunch of instructions to the README to cover this. Take a look and let me know if you want me to make any changes.

stack.yaml
andremarianiello: bump

ggrieco-tob: While it sounds reasonable to me, we are not currently supporting nix to build echidna. We would love to do it, but we a have a few reasons to avoid it. If we decided to accept this change, we *should* keep updated this list of packages, in case someone want to use it. But since we are not using nix, that's why I asked you to delete that section. If you want, feel free to support an installation using nix in your own branch. We could even link your branch from our README, in case someone is interested in using nix.

ggrieco-tob:
Can you add a simple example in the `solidity` folder to quickly test this feature? (we want to have a small solidity contract test suite for all the echidna features in the near future).

ggrieco-tob:
It seems that the contract argument fuzzing is not well integrated into the coverage mode. Once echidna tests and runs certain constructor paramters, they are not properly saved into the `outdir` directory (only the function calls are saved). Without this, other echidna instances won't be able to replicate a test case and the "distributed mode" will be quite ineffective. Can you fix this?

stack.yaml
asymmetric: What's the advantage of using `stack --nix` rather than just adding a nix derivation for echidna to nixpkgs?

Also, IIRC all hackage packages are automatically built on nixpkgs.

asymmetric: @andremarianiello once echidna is on hackage (the Haskell package repository) it **should** be picked up within a week or so by the nix infrastructure automatically.

There might be build issues and then we'd have to look into it, but I think pushing to Hackage is the first step.

asymmetric: But now I realize that the above tool only works with packages using Cabal, not Stack. There's [stack2nix](https://github.com/input-output-hk/stack2nix) for that, but the last time I tried packaging echidna with it, it was broken. :man_shrugging:

andremarianiello: @ggrieco-tob Can expand on

> we are not currently supporting nix to build echidna. We would love to do it, but we a have a few reasons to avoid it.

andremarianiello:
> Can you add a simple example in the `solidity` folder to quickly test this feature? (we want to have a small solidity contract test suite for all the echidna features in the near future).

Done

andremarianiello:
> It seems that the contract argument fuzzing is not well integrated into the coverage mode. Once echidna tests and runs certain constructor paramters, they are not properly saved into the `outdir` directory (only the function calls are saved). Without this, other echidna instances won't be able to replicate a test case and the "distributed mode" will be quite ineffective. Can you fix this?

I now store the used arguments in the json that is output to coverage files. I think will allow faithful reproduction of the test case, as requested. I also included the constructor args as part of the "unique key" for the set of tests. If you would rather the CoverageInfo remain the only member of that key, I can revert that change.

ggrieco-tob:
Great!, I will review the new changes during this week.

andremarianiello:
@ggrieco-tob Any more suggestions?

ggrieco-tob:
Sorry for the delay!, I had trouble merging your last commits (maybe because of the forced-pushes?). Anyway, it is fixed now so I'm reviewing your latest changes..

ggrieco-tob:
I'm testing your latest commits with a modification of the `solidity/coverage/single.sol` file:

```
contract C {
  bool state = true;

  function C(uint x) {
     require(x <= 1024);
  }

  function f(uint x, uint y, uint z) {
    require(x == 12);
    require(y == 8);
    require(z == 0);
    state = false;
    return; 
  }

  function echidna_state() returns (bool) {
    return state;
  }

}
```

If you run this file with new code, you should get:

```
$ echidna-test single.sol --config test.yaml 
...
"Analyzing contract: single.sol:C"
1000000
Found 0 new testcases!
Found -6175 new testcases!
...
```

It takes a long time to find the a way to make that trivial property false. However, if you use the original test, it founds the counter-example almost immediately. I assuming there is some issues with the new code in the generation/fuzzing of the transactions, once echidna finds a way to avoid the revert in the constructor. Let's fix this before merging.

Also, can you fix the report of the negative numbers in number of test cases? :smile:

andremarianiello:
@ggrieco-tob Thanks for taking a look; I'll investigate.

andremarianiello:
@ggrieco-tob I was using the constructor args AND the coverage info for identifying unique code paths. This was causing the number of unique test cases to blow up, and was unnecessary. I changed it back to using only the coverage info. Also, I commit the change you made to the single.sol contract.

I could not reproduce the negative test cases output you got, so you might have to show me how you did that.

I also added constructor args to the output for failed properties.

ggrieco-tob:
The issue with the negative number of test cases is fixed! However, I'm testing your latest commits with a modification of the `solidity/coverage/multi.sol` file and I think I found a regression:

```
contract C {
  bool state1 = false;
  bool state2 = false;
  bool state3 = false;
  
  function f(uint x) {
    require(x == 12);
    state1 = true;
  }

  function g(uint y) {
    require(state1);
    require(y == 8);
    state2 = true;    
  }
  
  function h(uint z) {
    require(state2);
    require(z == 0);
    state3 = true;
  }
  
  function ul(uint) {
    uint x = 0;
    return;
  }
  
  function echidna_state3() returns (bool) {
    return (!state3);
  }

}
```

To run this example:

```
$ mkdir -p outdir
$ rm outdir/* 
$ echidna-test multi.sol --config test.yaml
```

Using the original `dev-no-hedgehog` branch, you get:

```
"Analyzing contract: multi.sol:C"
1000000
Found 0 new testcases!
Found 0 new testcases!
Failed property "echidna_state3"
Original input:
h(0) by a329c0648769a73afac7f9381e08fb43dbea70
g(8) by a329c0648769a73afac7f9381e08fb43dbea70
ul(69886639306845688416592989777230678738175469504079973247) by a329c0648769a73afac7f9381e08fb43dbea70
f(12) by a329c0648769a73afac7f9381e08fb43dbea70
Original events:
[]
Reduced input:
h(0) by a329c0648769a73afac7f9381e08fb43dbea70
g(8) by a329c0648769a73afac7f9381e08fb43dbea70
f(12) by a329c0648769a73afac7f9381e08fb43dbea70
Reduced events:
[]
```

but using your PR:

```
"Analyzing contract: multi.sol:C"
1000000
Found 0 new testcases!
Found 0 new testcases!
Failed property "echidna_state3"
Constructor Arguments:
Original input:
h(23192810032488445957108697628195474228326257) by a329c0648769a73afac7f9381e08fb43dbea70
g(8) by a329c0648769a73afac7f9381e08fb43dbea70
f(12) by a329c0648769a73afac7f9381e08fb43dbea70
Original events:
[]
Reduced input:
h(23192810032488445957108697628195474228326257) by a329c0648769a73afac7f9381e08fb43dbea70
g(8) by a329c0648769a73afac7f9381e08fb43dbea70
f(12) by a329c0648769a73afac7f9381e08fb43dbea70
Reduced events:
[]
```

which is clearly not the correct result. Can you fix this regression?

andremarianiello:
Good catch. The original call sequence `cs` was being reported, not the mutated call sequence `cs'` which was actually triggering the property failure. Pushed the fix.

ggrieco-tob:
Great work!, the bug seems to be fixed. A few small remaining issues: If you create a simple contract like this one:

```
contract C {
    function C(uint x) {
        assert(x == 0);
    }
    function echidna_fail() returns (bool) {
        return false;
    } 
}
```

Then echidna will complain that there are no functions to execute (which is no longer the case in this branch).

Additionally, can you print the constructor call using the same convention as the calls when we print the original and the minimized ones? For instance:

```
Original input:
C(8994384938493) by a329c0648769a73afac7f9381e08fb43dbea70
f(0) by a329c0648769a73afac7f9381e08fb43dbea70
Original events:
[]
Reduced input:
C(42) by a329c0648769a73afac7f9381e08fb43dbea70
f(0) by a329c0648769a73afac7f9381e08fb43dbea70
Reduced events:
[]
```

Also, please verify that the constructor can be minimized and add one example of a small contract in the solidity folder.

Finally, do not forget to fix the warnings, otherwise, the CI tests will fail.

japesinator:
Closing, outdated

PR #153, opened 2018-11-21
automagically read json files

japesinator:


japesinator:
Closing this because it's really old and I think I'd prefer to do this explicitly regardless

PR #162, opened 2019-02-08
Testing and general improvements

incertia:
Basically adds a unit testing suite to the Stack project and has a few code improvements as well.


ggrieco-tob:
A few task before merging this:

1. Solve the conflicts.
2. Move the `true.sol` to `examples/solidity/basic` and update `Spec.hs`. 
3. Build and test the docker container (just in case).

src/Main.hs
```diff
@@ -37,10 +38,10 @@ opts = info (options <**> helper) $ fullDesc
...
-            (v,a,ts) <- loadSolidity f (pack <$> c)
-            let r = v ^. state . contract
-            let w = World (cfg ^. sConf . sender) [(r, a)]
-            let ts' = zip ts $ repeat r
-            ui v w ts' >> pure ()
+          Campaign r _ <- flip runReaderT (cfg & cConf %~ (if cov then \k -> k {knownCoverage = Just mempty}
+                                                                  else id)) $ do
+            (v, w, ts) <- loadTesting f (pack <$> c)
+            ui v w ts
+          if any (\case {Passed -> False; _ -> True;}) (snd <$> r)
```
japesinator: `Passed` doesn't have any args, so we can just derive `Eq` and do `== Passed`

src/Main.hs
```diff
@@ -37,10 +38,10 @@ opts = info (options <**> helper) $ fullDesc
...
+          then exitWith $ ExitFailure 1
```
japesinator: Nit: indent again

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+        let tr = findtest' c "echidna_all_sender"
+        assertBool "echidna_all_sender unsolved" $ solved tr
+        let sol = solve tr
+        assertBool "solution has length /= 3" $ length sol == 3
+        let calls = view call <$> sol
+        assertBool "s1 not in solution" $ any (\case {Left ("s1", _) -> True; _ -> False;}) calls
+        assertBool "s2 not in solution" $ any (\case {Left ("s2", _) -> True; _ -> False;}) calls
+        assertBool "s3 not in solution" $ any (\case {Left ("s3", _) -> True; _ -> False;}) calls
+  ]
+  where c2   = "./examples/solidity/basic/flags.sol"
```
japesinator: Let's `cd` here rather than typing this giant prefix each time

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+
+testContract' :: FilePath -> (Campaign -> HU.Assertion) -> HU.Assertion
+testContract' = flip testContract (Left defaultConfig)
+
+solved :: TestState -> Bool
+solved (Large _ _) = True
+solved (Solved _)  = True
+solved _           = False
+
+passed :: TestState -> Bool
```
japesinator: Use the `Eq` instance trick described above

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+passed Passed = True
+passed _      = False
+
+solve :: TestState -> [Tx]
+solve (Large _ s) = s
+solve (Solved s) = s
+solve _ = undefined
```
japesinator: We shouldn't have `undefined` in code. Use `MonadError` if you need to throw, or a sum type otherwise

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+
+findtest :: Campaign -> Text -> Maybe TestState
+findtest = findtest'' . (^. tests)
```
japesinator: Use `view`

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+
+findtest' :: Campaign -> Text -> TestState
+findtest' = (fromJust .) . findtest
```
japesinator: I'm not sure I like these names/this organization of logic. Let's consolidate this into one function, and avoid partial stuff (`fromJust`) when we can. IMO, it's even better to throw our own `error` b/c that's much more transparent to non-Haskell-familiar users

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+
+findtest'' :: [(SolTest, TestState)] -> Text -> Maybe TestState
+findtest'' [] _ = Nothing
+findtest'' ((st, ts):xs) t = if t == fst st then Just ts else findtest'' xs t
```
japesinator: Reimplement with `find`

lib/Echidna/Solidity.hs
```diff
@@ -63,27 +64,34 @@ data SolConf = SolConf { _contractAddr :: Addr   -- ^ Contract address to use
...
 contracts :: (MonadIO m, MonadThrow m, MonadReader x m, Has SolConf x) => FilePath -> m [SolcContract]
-contracts fp = view (hasLens . solcArgs) >>= liftIO . solc >>= (\case
-  Nothing -> throwM CompileFailure
-  Just m  -> pure . toList $ fst m) where
-    solc a = readSolc =<< writeSystemTempFile "" =<< readProcess "solc" (usual <> words a) ""
-    usual = ["--combined-json=bin-runtime,bin,srcmap,srcmap-runtime,abi,ast", fp]
-    
+contracts fp = do
+  a <- view (hasLens . solcArgs)
+  stfu <- view (hasLens . quiet)
```
japesinator: Let's avoid `stfu` as a variable name, both here and below.

lib/Echidna/Solidity.hs
```diff
@@ -63,27 +64,34 @@ data SolConf = SolConf { _contractAddr :: Addr   -- ^ Contract address to use
...
+  pure (a, stfu) >>= liftIO . solc >>= (\case
+    Nothing -> throwM CompileFailure
+    Just m  -> pure . toList $ fst m) where
+      solc (a, stfu) = do
+        stderr <- if stfu
+                  then UseHandle <$> openFile "/dev/null" WriteMode
```
japesinator: Indent

japesinator: (alternatively, just move `then` up a line, indent `else` to match

lib/Echidna/Solidity.hs
```diff
@@ -63,27 +64,34 @@ data SolConf = SolConf { _contractAddr :: Addr   -- ^ Contract address to use
...
-contracts fp = view (hasLens . solcArgs) >>= liftIO . solc >>= (\case
-  Nothing -> throwM CompileFailure
-  Just m  -> pure . toList $ fst m) where
-    solc a = readSolc =<< writeSystemTempFile "" =<< readProcess "solc" (usual <> words a) ""
-    usual = ["--combined-json=bin-runtime,bin,srcmap,srcmap-runtime,abi,ast", fp]
-    
+contracts fp = do
+  a <- view (hasLens . solcArgs)
+  stfu <- view (hasLens . quiet)
+  pure (a, stfu) >>= liftIO . solc >>= (\case
```
japesinator: Can we pass the `Handle` straight to `solc`? I think that makes our intent much more clear. It could also let us avoid binding an intermediate variable here.

lib/Echidna/Solidity.hs
```diff
@@ -8,7 +8,7 @@ module Echidna.Solidity where
 
 import Control.Lens
 import Control.Exception          (Exception)
-import Control.Monad              (liftM2, mapM_)
+import Control.Monad              (liftM2, mapM_, when)
```
japesinator: Good call on `when`!

lib/Echidna/Solidity.hs
```diff
@@ -107,3 +115,16 @@ loadSolidity fp name = let ensure (l, e) = if null l then throwM e else pure ()
     case find (not . null . snd) tests of
       (Just (t,_)) -> throwM $ TestArgsFound t
       Nothing      -> return (loaded, funs, fst <$> tests)
+
+-- | Basically loadSolidity, but prepares the results to be passed directly into
+-- a testing function.
+loadTesting :: (MonadIO m, MonadThrow m, MonadReader x m, Has SolConf x)
```
japesinator: Can we rename for clarity? Maybe `loadSolTests` or something

lib/Echidna/Solidity.hs
```diff
@@ -63,27 +64,34 @@ data SolConf = SolConf { _contractAddr :: Addr   -- ^ Contract address to use
...
 -- throwing errors if necessary.
 selected :: (MonadIO m, MonadThrow m, MonadReader x m, Has SolConf x) => FilePath -> Maybe Text -> m SolcContract
 selected fp name = do cs <- contracts fp
                       c <- choose cs $ ((pack fp <> ":") <>) <$> name
-                      liftIO $ if isNothing name && length cs > 1
-                        then putStrLn "Multiple contracts found in file, only analyzing the first"
-                        else pure ()
-                      liftIO . putStrLn $ "Analyzing contract: " <> unpack (c ^. contractName)
+                      stfu <- view (hasLens . quiet)
+                      liftIO $ when (isNothing name && length cs > 1) $ putStrLn "Multiple contracts found in file, only analyzing the first"
```
japesinator: Let's get these two `when`s in one `liftIO`, and split lines like this to be a bit shorter. Also, general note, I think we should prefer `f . g $ h x` style (relevant a few places in this PR)

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+        assertBool "echidna_true somehow did not pass" $ passed tr
+  , testCase "Simple Flags Example" $ testContract' c2 $
+      \c -> do
+        let t1 = findtest' c "echidna_alwaystrue"
+            t2 = findtest' c "echidna_sometimesfalse"
+        assertBool "echidna_alwaystrue did not pass" $ passed t1
+        assertBool "echidna_sometimesfalse unsolved" $ solved t2
+        let sol = solve t2
+        assertBool "solution has length /= 2" $ length sol == 2
+        -- the following check will fail with significantly higher probability
```
japesinator: Rather than checking in commented code, I think we should get this out of the file and into a github issue, where it's easier to manage

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+        let sol = solve tr
+        assertBool "solution has length /= 3" $ length sol == 3
+        let calls = view call <$> sol
+        assertBool "s1 not in solution" $ any (\case {Left ("s1", _) -> True; _ -> False;}) calls
+        assertBool "s2 not in solution" $ any (\case {Left ("s2", _) -> True; _ -> False;}) calls
+        assertBool "s3 not in solution" $ any (\case {Left ("s3", _) -> True; _ -> False;}) calls
+  ]
+  where c2   = "./examples/solidity/basic/flags.sol"
+        c3   = "./examples/solidity/basic/revert.sol"
+        --c4   = "./examples/solidity/basic/payable.sol"
```
japesinator: Same note w.r.t. checking in commented code

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+        let tr = findtest' c "echidna_test"
+        assertBool "echidna_test unsolved" $ solved tr -}
+  , testCase "Multisender" $ testContract c6 (Right cfg6) $
+      \c -> do
+        let tr = findtest' c "echidna_all_sender"
+        assertBool "echidna_all_sender unsolved" $ solved tr
+        let sol = solve tr
+        assertBool "solution has length /= 3" $ length sol == 3
+        let calls = view call <$> sol
+        assertBool "s1 not in solution" $ any (\case {Left ("s1", _) -> True; _ -> False;}) calls
```
japesinator: Let's make this pattern into a helper function rather than duplicating this logic.

src/test/Spec.hs
```diff
@@ -0,0 +1,120 @@
...
+        --   case call2 of
+        --        Left ("set1", [AbiInt _ 0]) -> True
+        --        _                           -> False
+  , testCase "Optimal Solve" $ testContract' c3 $
+      \c -> do
+        let tr = findtest' c "echidna_revert"
+        assertBool "echidna_revert unsolved" $ solved tr
+        let sol = solve tr
+        assertBool "solution has length /= 1" $ length sol == 1
+        let sol' = head sol
```
japesinator: I'm worried the partial function here will obfuscate the reason for future failures from people trying to contribute to echidna. Let's instead check for the empty list.

lib/Echidna/Test.hs
```diff
@@ -52,6 +52,8 @@ checkETest (f, a) = asks getter >>= \(TestConf p s) -> do
 -- still solves that test.
 shrinkSeq :: (MonadRandom m, MonadReader x m, Has TestConf x, MonadState y m, Has VM y, MonadThrow m)
           => SolTest -> [Tx] -> m [Tx]
-shrinkSeq t xs = shorten >>= mapM shrinkTx >>= ap (fmap . flip bool xs) check where
+shrinkSeq t xs = sequence [shorten, shrunk, shrunk'] >>= uniform >>= ap (fmap . flip bool xs) check where
```
japesinator: We can simplify this to just `[shorten, shrunk]` since most IRL uses have small sequences

lib/Echidna/Campaign.hs
```diff
@@ -77,8 +77,9 @@ updateTest v (Just (v', xs)) (n, t) = view (hasLens . to testLimit) >>= \tl -> (
   _                   -> snd <$> updateTest v Nothing (n,t)
 updateTest v Nothing (n, t) = view (hasLens . to shrinkLimit) >>= \sl -> (n,) <$> case t of
   Large i x | i >= sl -> pure $ Solved x
-  Large i x           -> if any canShrinkTx x then Large (i + 1) <$> evalStateT (shrinkSeq n x) v
-                                              else pure $ Solved x
+  Large i x           -> if length x > 1 || any canShrinkTx x
+                         then Large (i + 1) <$> evalStateT (shrinkSeq n x) v
```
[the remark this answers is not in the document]
ggrieco-tob: Change the space between `x` and `->` will break the alignment of the guards (` i >= tl` and `i >= sl`:

```
updateTest v (Just (v', xs)) (n, t) = view (hasLens . to testLimit) >>= \tl -> (n,) <$> case t of
  Open i    | i >= tl -> pure Passed
  Open i              -> catch (evalStateT (checkETest n) v' <&> bool (Large (-1) xs) (Open (i + 1)))
                               (pure . Failed)
  _                   -> snd <$> updateTest v Nothing (n,t)
updateTest v Nothing (n, t) = view (hasLens . to shrinkLimit) >>= \sl -> (n,) <$> case t of
  Large i x | i >= sl -> pure $ Solved x
  Large i x           -> if length x > 1 || any canShrinkTx x
                         then Large (i + 1) <$> evalStateT (shrinkSeq n x) v
                         else pure $ Solved x
  _                   -> pure t
```

If you don't care, we can change it.

lib/Echidna/Solidity.hs
```diff
@@ -63,27 +64,34 @@ data SolConf = SolConf { _contractAddr :: Addr   -- ^ Contract address to use
                        , _sender       :: [Addr] -- ^ Sender addresses to use
                        , _prefix       :: Text   -- ^ Function name prefix used to denote tests
                        , _solcArgs     :: String -- ^ Args to pass to @solc@
+                       , _quiet        :: Bool   -- ^ Option to suppress some output and stderr of @solc@
```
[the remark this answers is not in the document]
ggrieco-tob: Done.

src/Main.hs
ggrieco-tob: We can't use `deriving Eq` in `TestState`, because it uses `ExecException`, which uses `Error` (in HEVM), which uses a GADT: `Query` (in HEVM). :sob:

src/test/Spec.hs
ggrieco-tob: Done.

src/test/Spec.hs
ggrieco-tob: Done.

lib/Echidna/Solidity.hs
ggrieco-tob: Done.

.travis/install-solc.sh
```diff
@@ -0,0 +1,28 @@
...
+set -eux
+
+mkdir -p ~/.local/bin;
+
+travis_retry() {
+  cmd=$*
+  $cmd || (sleep 2 && $cmd) || (sleep 10 && $cmd)
+}
+
+#fetch_stack_osx() {
```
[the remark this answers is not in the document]
ggrieco-tob: Done.

.travis/install-libsecp256k1.sh
```diff
@@ -2,8 +2,6 @@
 
 set -eux
 
-if ls /usr/local/lib | grep -q libsecp256k1; then exit 0; fi
```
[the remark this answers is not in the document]
ggrieco-tob: Done.

lib/Echidna/Campaign.hs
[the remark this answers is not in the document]
incertia: i'm fairly certain he means the `then` and `else` after the `if`, which i can get on right now

src/test/Spec.hs
incertia: since this is just code for unit testing purposes, the idea is to use this to obtain the transaction list after determining that a test case has actually been solved. maybe we can do the following?

```haskell
solve :: TestState -> Maybe [Tx]
solve (Large _ s) = Just s
solve (Solved s) = Just s
solve _ = Nothing
```

it would still bloat the testing code though, and we would need to put `fromJust` in a bunch of places.

lib/Echidna/Solidity.hs
incertia: done

lib/Echidna/Test.hs
incertia: done

src/test/Spec.hs
ggrieco-tob: Done.

lib/Echidna/Solidity.hs
ggrieco-tob: Done (@incertia)

lib/Echidna/Test.hs
```diff
@@ -52,6 +52,7 @@ checkETest (f, a) = asks getter >>= \(TestConf p s) -> do
 -- still solves that test.
 shrinkSeq :: (MonadRandom m, MonadReader x m, Has TestConf x, MonadState y m, Has VM y, MonadThrow m)
           => SolTest -> [Tx] -> m [Tx]
-shrinkSeq t xs = shorten >>= mapM shrinkTx >>= ap (fmap . flip bool xs) check where
+shrinkSeq t xs = sequence [shorten, shrunk] >>= uniform >>= ap (fmap . flip bool xs) check where
```
japesinator: This line is so pretty now

incertia: `ap` is still black magic

src/test/Spec.hs
```diff
@@ -74,12 +73,13 @@ solidityTests = testGroup "Solidity-HUnit"
         assertBool "s2 not in solution" $ any (\case {Left ("s2", _) -> True; _ -> False;}) calls
         assertBool "s3 not in solution" $ any (\case {Left ("s3", _) -> True; _ -> False;}) calls
   ]
-  where c2   = "./examples/solidity/basic/flags.sol"
-        c3   = "./examples/solidity/basic/revert.sol"
-        --c4   = "./examples/solidity/basic/payable.sol"
-        c5   = "./examples/solidity/basic/true.sol"
-        c6   = "./examples/solidity/basic/multisender.sol"
-        cfg6 = "./examples/solidity/basic/multisender.yaml"
+  where cd   = "./examples/solidity/basic/"
```
incertia: instead of doing `cd ++ "whatever"` here we could do `main = setCurrentDirectory "./examples/solidity/basic" >> defaultMain tests`

src/test/Spec.hs
ggrieco-tob: Done (@japesinator)

src/test/Spec.hs
ggrieco-tob: Done (@japesinator)

src/test/Spec.hs
ggrieco-tob: Done (@japesinator)

src/test/Spec.hs
ggrieco-tob: Done (@japesinator)

src/test/Spec.hs
ggrieco-tob: Done (@japesinator)

lib/Echidna/Campaign.hs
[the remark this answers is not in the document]
ggrieco-tob: Done (@incertia )

src/test/Spec.hs
ggrieco-tob: Done (@japesinator )

src/test/Spec.hs
ggrieco-tob: Done (@japesinator)

lib/Echidna/ABI.hs
```diff
@@ -189,7 +189,7 @@ incBits x = (x .|.) . fromIntegral <$> getRandomR (bounds x)
 -- | Given an 'AbiValue', generate a random \"smaller\" (simpler) value of the same 'AbiType'.
 shrinkAbiValue :: MonadRandom m => AbiValue -> m AbiValue
 shrinkAbiValue (AbiUInt n m) = AbiUInt n <$> dropBits m
-shrinkAbiValue (AbiInt n m)  = AbiInt n  <$> (if m > 0 then dropBits else incBits) m
+shrinkAbiValue (AbiInt n m)  = AbiInt n  <$> (case m `compare` 0 of {GT -> dropBits; EQ -> pure; LT -> incBits}) m
```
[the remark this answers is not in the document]
ggrieco-tob: Done (@japesinator)

src/test/Spec.hs
ggrieco-tob: Done (@japesinator)

src/Main.hs
ggrieco-tob: Done. (@japesinator)

lib/Echidna/Solidity.hs
ggrieco-tob: Done. (@japesinator)

lib/Echidna/Solidity.hs
ggrieco-tob: Done.

PR #165, opened 2019-02-09
Implementation of UI selection [WIP]

ggrieco-tob:


ggrieco-tob:
This is no longer required, since we have #175.

PR #166, opened 2019-02-11
Enable the use of dictionaries to generate transactions [WIP]

ggrieco-tob:


ggrieco-tob:
This is no longer necessary, since we have #181.

PR #177, opened 2019-02-21
Support for solc 0.5.x

ggrieco-tob:
Echidna was already working fine with Solidity 0.5.4. This PR fixes the testing contracts and adds the use of newer solidity version in the CI scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,17 +36,14 @@ script:
  - STACK_OPTS="$STACK_OPTS --extra-include-dirs=/usr/local/include --extra-lib-dirs=/usr/local/lib"
  - |
    set -ex
-   # Run tests using solc 0.4.25
-   export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib"
-   cp ~/.local/bin/solc-0.4.25 ~/.local/bin/solc || true
-   stack --no-terminal test --ghc-options="$GHC_OPTIONS" $STACK_OPTS
-   set +ex
- - |
-   set -ex
-   # Run tests using solc 0.5.4
-   export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib"
-   cp ~/.local/bin/solc-0.5.4 ~/.local/bin/solc || true
-   stack --no-terminal test --ghc-options="$GHC_OPTIONS" $STACK_OPTS
+   run_tests_using () {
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -36,10 +36,19 @@ script:
...
  - |
    set -ex
-   # Run tests
+   # Run tests using solc 0.4.25
    export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib"
+   cp ~/.local/bin/solc-0.4.25 ~/.local/bin/solc || true
    stack --no-terminal test --ghc-options="$GHC_OPTIONS" $STACK_OPTS
    set +ex
+ - |
+   set -ex
```
japesinator: Rather than duplicating set +ex and export, let's just make a function that takes a binary, symbolic links it into $PATH as solc, and then runs stack tests. Then we can just run that function twice

examples/solidity/bad/nobytecode.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.24;
+//pragma solidity ^0.4.24;
```
japesinator: We can just delete these

.travis.yml
ggrieco-tob: Another option is to implement a new config option: `solcBin` to select `solc-0.4.25` and `solc-0.5.4` during the compilation of a contract.

ggrieco-tob: Done (but untested, let's wait for travis..)

examples/solidity/bad/nobytecode.sol
ggrieco-tob: Done.

.travis.yml
ggrieco-tob: Works as expected.

.travis.yml
```diff
@@ -36,11 +36,17 @@ script:
  - STACK_OPTS="$STACK_OPTS --extra-include-dirs=/usr/local/include --extra-lib-dirs=/usr/local/lib"
  - |
    set -ex
-   # Run tests
-   export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib"
-   stack --no-terminal test --ghc-options="$GHC_OPTIONS" $STACK_OPTS
+   run_tests_using () {
+     export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib"
```
japesinator: nit: let's move this outside the function

PR #178, opened 2019-02-21
Fixed use of contractAddr to define the address of the contract

ggrieco-tob:


japesinator:
Instead of reimplementing `vmForEthrunCreation`, let's just do `blank & state . caller .~ addr & env . origin .~ addr` to be a lot cleaner. We can modify `setupTx` for this too

ggrieco-tob:
It's easier to close this and open #182 .

PR #179, opened 2019-02-24
Show the senders when printing a list of transactions

ggrieco-tob:
This small PR enables to show where the transactions are coming from when Echidna finds a counter-example. This is particularly useful when you are using multiple senders. For instance, using the `multisender.sol` example:

```
$ echidna-test examples/solidity/basic/multisender.sol --config examples/solidity/basic/multisender.yaml
...
echidna_all_sender: failed!💥  
  Call sequence, shrinking (4825/5000):
    s3() from 0x0000000000000000000000000000000000000003
    s2() from 0x0000000000000000000000000000000000000002
    s1() from 0x0000000000000000000000000000000000000001
```

### DIFF
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -11,6 +11,7 @@ import Control.Monad.Reader (ReaderT(..))
 import Data.Has (Has(..))
 import Data.Aeson
 import EVM (result)
+import EVM.Types (Addr)
 
 import qualified Data.ByteString as BS
 import qualified Data.Yaml as Y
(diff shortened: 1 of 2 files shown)


japesinator:
I'm not sure I think we should keep this on by default, since most users only have one sender. I'm open to making it configurable, but I think we should think more about the UX.

ggrieco-tob:
Uhm, any suggestion on how to keep the count of sender addresses ? For instance:

```
-- | Number of different addresses per role
type Cardinality = Role -> String
```

PR #181, opened 2019-02-26
Add dictionary support, preserving existing API where possible

japesinator:


src/Main.hs
```diff
@@ -34,9 +32,10 @@ opts = info (options <**> helper) $ fullDesc
...
-          Campaign r _ <- runReaderT (loadSolTests f (pack <$> c) >>= \(v,w,ts) -> ui v w ts) $
-                            cfg & cConf %~ if cov then \k -> k {knownCoverage = Just mempty} else id
-          if any (/= Passed) $ snd <$> r then exitWith $ ExitFailure 1
-                                         else exitSuccess
+main = do Options f c conf <- execParser opts
+          cfg <- maybe (pure defaultConfig) parseConfig conf
+          res <- flip runReaderT cfg $ do
+            cs       <- contracts f
+            (v,w,ts) <- loadSpecified (pack . (f ++) . (':' :) <$> c) cs >>= prepareForTest
+            view tests <$> ui v w ts (Just $ mkGenDict 0.15 (extractConstants cs) [])
```
ggrieco-tob: Should we make that `0.15` a config parameter? I don't see a lot of users tweaking that, but it could be useful for debugging ..

examples/solidity/coverage/single.sol
```diff
@@ -1,14 +1,14 @@
 contract C {
```
ggrieco-tob: I think echidna is having trouble with this example. Is it because of the `require` that cause a revert?

src/Main.hs
japesinator: Now that we're trying to do releases, I'm gonna try to keep stuff like that out of config until we need to put it in there

ggrieco-tob:
I'm getting an error with this: 

```
contract Constants {
  bool found = false;

  function find(bytes32 s) public {
    if (s == "test") {found = true;}
  }

  function echidna_found() public view returns (bool) {
    return(!found);
  }
}
```

```
echidna-test: Data.ByteString.index: index too large: 4, length = 4
CallStack (from HasCallStack):
  error, called at libraries/bytestring/Data/ByteString.hs:1877:23 in bytestring-0.10.8.2:Data.ByteString
```

examples/solidity/coverage/single.sol
japesinator: I think the problem is both `mutateCall` and `shrinkTx` modify all the different args at once. We need something that mutates/shrinks less aggressively

ggrieco-tob:
Weird bug, if echidna finds two constants, the smallest one is never used in the generation:

```
contract Constants {
  bool found = false;

  function find(int i) public {
    if (i == 1337) {found = true;}
    if (i == 133700000000) {found = false;}
  }

  function echidna_found() public view returns (bool) {
    return(!found);
  }
}
```

emavgl:
As pointed out in #183 Echidna is not able to pass the require in the function transfer.

`config.yaml`:

```
sender: ["0x00a329c0648769a73afac7f9381e08fb43dbea70", "0x00a329c0648769a73afac7f9381e08fb43dbea60", "0x00a329c0648769a73afac7f9381e08fb43dbea50"]
```

and the `source file`:

```javascript
contract TEST {
  mapping(address => uint256) internal balances;

  address private alice = 0x00A329C0648769a73aFAc7F9381E08fB43DbeA50;
  address private carl = 0x00A329C0648769a73aFAc7F9381E08fB43DbeA77;
  uint totalSupply = 100;

  constructor() public {
    balances[alice] = totalSupply;
  }

  function transfer(address _to, uint _value) public returns (bool) {
    require(_value <= balances[msg.sender]);
    balances[msg.sender] -= _value;
    balances[_to] += _value;
  }

    return balances[alice] == totalSupply;
  }
}
```

Result

```
 ┌─────────Echidna─────────┐                                                    
   echidna_test: passed! 🎉│                                                                               └─────────────────────────┘                                                    
Campaign complete, C-c or esc to print report 
```

ggrieco-tob:
I think I fixed the bug in the dictionary generation. Echidna is now passing the `single.sol` and `multi.sol` examples :tada: . @emavgl please retest and let us know the result.

emavgl:
> I think I fixed the bug in the dictionary generation. Echidna is now passing the `single.sol` and `multi.sol` examples tada . @emavgl please retest and let us know the result.

Ok, even with `totalSupply = 100` now is working.

```
Analyzing contract: new_test.sol:TEST
echidna_test: failed!💥  
  Call sequence, shrinking (819/5000):
    transfer(244267d07eba8bd7ed4d3bccb8a6873f0ad11d58,4)
```

Where does the address `244267d07eba8bd7ed4d3bccb8a6873f0ad11d58` come from? It is not one of senders and neither one of the hardcoded addresses. How can I enforce to use a given set of addresses as possible actual parameters (previous behaviour with the configuration field `addrList`)?.

ggrieco-tob:
> > I think I fixed the bug in the dictionary generation. Echidna is now passing the `single.sol` and `multi.sol` examples tada . @emavgl please retest and let us know the result.
> 
> Ok, even with `totalSupply = 100` now is working.
> 
> ```
> Analyzing contract: new_test.sol:TEST
> echidna_test: failed!💥  
>   Call sequence, shrinking (819/5000):
>     transfer(244267d07eba8bd7ed4d3bccb8a6873f0ad11d58,4)
> ```

Great!

> Where does the address `244267d07eba8bd7ed4d3bccb8a6873f0ad11d58` come from? It is not one of senders and neither one of the hardcoded addresses. 

It's just a random address. Echidna uses constants extracted from the contract **and** randomly generated values in the generation of transactions. In this case, if `alice` transfer to any address, the property will fail. So maybe you want:

```
    return balances[alice] <= totalSupply;
  }
```

> How can I enforce to use a given set of addresses as possible actual parameters (previous behaviour with the configuration field `addrList`)?.

You can't directly. If you want echidna to use a particular address, add it in a constant somewhere. Otherwise, use solidity to create a function like this:

```
  function transferToX(uint _value) public returns (bool) {
    return transfer(X, _value);
  }
```

lib/Echidna/Campaign.hs
```diff
@@ -121,27 +130,39 @@ evalSeq v e = go [] where
...
                (y:ys) -> e y >> go (y:r) ys
 
+-- | Execute a transaction, capturing the PC and codehash of each instruction executed, saving the
+-- transaction if it finds new coverage.
+execTxOptC :: (MonadState x m, Has Campaign x, Has VM x, MonadThrow m) => Tx -> m VMResult
+execTxOptC t = do
+  og  <- hasLens . coverage <<.= mempty
+  res <- execTxWith vmExcept (usingCoverage $ pointCoverage (hasLens . coverage)) t
+  hasLens . coverage %= unionWith union og
+  grew <- (== GT) . comparing coveragePoints og <$> use (hasLens . coverage)
```
ggrieco-tob: I did a few experiments with different contracts and I found that `grew` was `False` all the time. I'm investigating this..

ggrieco-tob: I believe the correct code should be: 
```
grew <- (== LT) . comparing coveragePoints og <$> use (hasLens . coverage)
```

japesinator: oops, good call

ggrieco-tob:
> I'm getting an error with this:
> 
> ```
> contract Constants {
>   bool found = false;
> 
>   function find(bytes32 s) public {
>     if (s == "test") {found = true;}
>   }
> 
>   function echidna_found() public view returns (bool) {
>     return(!found);
>   }
> }
> ```
> ```
> echidna-test: Data.ByteString.index: index too large: 4, length = 4
> CallStack (from HasCallStack):
>   error, called at libraries/bytestring/Data/ByteString.hs:1877:23 in bytestring-0.10.8.2:Data.ByteString
> ```

This should be solved.

PR #182, opened 2019-02-27
Fixed use of contractAddr to define the address of the contract

ggrieco-tob:


lib/Echidna/Transaction.hs
```diff
@@ -120,6 +120,7 @@ setupTx (Tx c s r v) = S.state . runState . zoom hasLens . sequence_ $
   , env . origin .= s, state . caller .= s, state . callvalue .= v, setup] where
     setup = case c of
       Left cd  -> loadContract r >> state . calldata .= encode cd
-      Right bc -> assign (env . contracts . ix r) (initialContract bc) >> loadContract r
+      Right bc -> sequence_ [ env . contracts . ix r .= initialContract bc,
```
japesinator: Don't these two lines do the same thing?

ggrieco-tob: Fixed!

PR #185, opened 2019-03-05
Fix contractAddr to use its configuration file

ggrieco-tob:
The `testContract` function can be used to test nonexistent functions (e.g. `echidna_addr`) and it won't report an error.

### DIFF
--- a/src/test/Spec.hs
+++ b/src/test/Spec.hs
@@ -66,7 +66,7 @@ integrationTests = testGroup "Solidity Integration Testing"
       , ("echidna_all_sender didn't shrink optimally",     solvedLen 3        "echidna_all_sender")
       ] ++ (["s1", "s2", "s3"] <&> \n ->
         ("echidna_all_sender solved without " ++ unpack n, solvedWith (n, []) "echidna_all_sender"))
-  , testContract "basic/contractAddr.sol"        Nothing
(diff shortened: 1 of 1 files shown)


japesinator:
We should just add an `open` function instead of `not . solved` I think. Also nit, can we preserve alignment of the configs?

src/test/Spec.hs
```diff
@@ -66,8 +66,10 @@ integrationTests = testGroup "Solidity Integration Testing"
       , ("echidna_all_sender didn't shrink optimally",     solvedLen 3        "echidna_all_sender")
       ] ++ (["s1", "s2", "s3"] <&> \n ->
         ("echidna_all_sender solved without " ++ unpack n, solvedWith (n, []) "echidna_all_sender"))
-  , testContract "basic/contractAddr.sol"        Nothing
-      [ ("echidna_addr failed",                  not . solved "echidna_addr") ]
+  , testContract "basic/contractAddr.sol" Nothing
+      [ ("echidna_address failed",                         solved "echidna_address") ]
+  , testContract "basic/contractAddr.sol" (Just "basic/contractAddr.yaml")
+      [ ("echidna_address failed",                         passed "echidna_address") ]
```
ggrieco-tob: I'm not sure if these two tests should have the same name?

incertia: it would require another argument to `testContract`

PR #192, opened 2019-03-21
Emit and set the seed for repeatable testing

incertia:
[this](https://github.com/crytic/echidna/commit/317a25ac7b00ba44554d0b4387787fa9d9d8c346#diff-034a47a58d6131377508c0edab99f1e4R129) is a bit sketchy but it seems fine because the `StdGen` is never looked at in `campaignStatus`.


lib/Echidna/UI.hs
```diff
@@ -112,11 +111,11 @@ monitor cleanup = let
 
 -- | Heuristic check that we're in a sensible terminal (not a pipe)
 isTerminal :: MonadIO m => m Bool
-isTerminal = liftIO $ queryTerminal (Fd 0)
+isTerminal = liftIO $ (&&) <$> queryTerminal (Fd 0) <*> queryTerminal (Fd 1)
```
japesinator: when do we need this check?

lib/Echidna/Campaign.hs
```diff
@@ -134,17 +145,18 @@ callseq v w ql = replicateM ql (evalStateT genTxM w) >>= \is -> use hasLens >>=
...
          => StateT Campaign m a -- ^ Callback to run after each state update (for instrumentation)
          -> VM                  -- ^ Initial VM state
          -> World               -- ^ Initial world state
          -> [SolTest]           -- ^ Tests to evaluate
          -> m Campaign
-campaign u v w ts = view (hasLens . to knownCoverage) >>= \c ->
-  execStateT runCampaign (Campaign ((,Open (-1)) <$> ts) c) where
-    step        = runUpdate (updateTest v Nothing) >> u >> runCampaign
+campaign u v w ts = view (hasLens . to knownCoverage) >>= \c -> do
+  g <- view (hasLens . to (fromMaybe (mkStdGen 0) . seed))
```
japesinator: do we wanna always use the same random? shouldn't we consider `newStdGen` instead?

lib/Echidna/Config.hs
```diff
@@ -60,10 +62,15 @@ instance FromJSON EConfig where
                 psender  <- v .:? "psender"  .!= 0x00a329c0648769a73afac7f9381e08fb43dbea70
                 let good = if reverts then (`elem` [ResTrue, ResRevert]) else (== ResTrue)
                 return $ TestConf (good . maybe ResOther classifyRes . view result) (const psender)
-        cc = CampaignConf <$> v .:? "testLimit"   .!= 10000
-                          <*> v .:? "seqLen"      .!= 10
-                          <*> v .:? "shrinkLimit" .!= 5000
-                          <*> pure Nothing
+        cc = do
+          s <- v .:? "seed"
+          let s' :: Maybe StdGen
```
japesinator: do we need the type annotation/auxiliary variable here?

lib/Echidna/UI.hs
incertia: piping to another program or file will still cause the ui to run, e.g. `stack run echidna-test -- solidity.sol --config cfg 1>output.log 2>/dev/null` and users will need to `C-c` out of echidna without seeing the UI.

lib/Echidna/Campaign.hs
incertia: this should never happen when running `echidna-test`. `getStdGen` is inserted into `EConfig` if a seed is not specified in the config.

lib/Echidna/Config.hs
incertia: fixed

lib/Echidna/Campaign.hs
japesinator: we're already in a `MonadThrow` though, there's no reason for this to be silent

src/Main.hs
```diff
@@ -35,8 +37,10 @@ opts = info (options <**> helper) $ fullDesc
 
 main :: IO ()
 main = do (Options f c cov conf) <- execParser opts
+          g <- getStdGen
           cfg          <- maybe (pure defaultConfig) parseConfig conf
-          Campaign r _ <- runReaderT (loadSolTests f (pack <$> c) >>= \(v,w,ts) -> ui v w ts) $
-                            cfg & cConf %~ if cov then \k -> k {knownCoverage = Just mempty} else id
+          Campaign r _ <- runReaderT (loadSolTests f (pack <$> c) >>= (\(v,w,ts) -> ui v w ts)) $
+                            cfg & cConf %~ (if cov then \k -> k {knownCoverage = Just mempty} else id)
+                                & cConf %~ \x -> if isJust (seed x) then x else x { seed = Just g }
```
japesinator: we can use something like `x {seed = seed x $> g}`

lib/Echidna/UI.hs
```diff
@@ -113,17 +113,33 @@ monitor cleanup = let
...
 
 -- | Set up and run an Echidna 'Campaign' while drawing the dashboard, then print 'Campaign' status
 -- once done.
+-- <<<<<<< HEAD
```
ggrieco-tob: Please remove all the merge comments (like this one). After that, this PR looks good to merge.

incertia: should have just been pushed

lib/Echidna/Campaign.hs
```diff
@@ -12,29 +12,37 @@ module Echidna.Campaign where
...
+import System.Random (mkStdGen)
 
 import Echidna.ABI
 import Echidna.Exec
 import Echidna.Test
 import Echidna.Transaction
 
+instance MonadThrow m => MonadThrow (RandT g m) where
+  throwM = lift . throwM
+instance MonadCatch m => MonadCatch (RandT g m) where
```
japesinator: Nit, let's add a newline above

lib/Echidna/Campaign.hs
```diff
@@ -46,6 +54,7 @@ data CampaignConf = CampaignConf { testLimit     :: Int
                                  , knownCoverage :: Maybe (Map W256 (Set Int))
                                    -- ^ If applicable, initially known coverage. If this is 'Nothing',
                                    -- Echidna won't collect coverage information (and will go faster)
+                                 , seed          :: Maybe Int
```
japesinator: Please add a comment documenting

lib/Echidna/Campaign.hs
```diff
@@ -167,11 +176,12 @@ campaign :: ( MonadCatch m, MonadRandom m, MonadReader x m, Has TestConf x, Has
          -> [SolTest]           -- ^ Tests to evaluate
          -> Maybe GenDict       -- ^ Optional generation dictionary
          -> m Campaign
-campaign u v w ts d = let d' = fromMaybe mempty d in fmap (fromMaybe mempty) (view (hasLens . to knownCoverage))
-  >>= \c -> execStateT runCampaign (Campaign ((,Open (-1)) <$> ts) c d') where
-    step        = runUpdate (updateTest v Nothing) >> u >> runCampaign
+campaign u v w ts d = let d' = fromMaybe mempty d in fmap (fromMaybe mempty) (view (hasLens . to knownCoverage)) >>= \c -> do
+  g <- view (hasLens . to (maybe (mkStdGen 0) mkStdGen . seed))
```
japesinator: I don't think we should just silently set the seed to zero here. Maybe this could take a `StdGen` instead?

lib/Echidna/Config.hs
```diff
@@ -59,14 +60,16 @@ instance FromJSON EConfig where
                           <*> v .:? "seqLen"      .!= 100
                           <*> v .:? "shrinkLimit" .!= 5000
                           <*> pure Nothing
+                          <*> v .:? "seed"
         names = const $ const mempty :: Names
-        ppc = cc <&> \c x -> runReader (ppCampaign x) (c, names)
-        style :: Y.Parser (Campaign -> String)
+        ppc = cc <&> \c x _ -> runReader (ppCampaign x) (c, names)
+        style :: Y.Parser (Campaign -> Maybe Int -> String)
```
japesinator: IMO, we should always know what the seed is here. I'm not sure why we need to handle the "Maybe" case

lib/Echidna/Config.hs
```diff
@@ -59,14 +60,16 @@ instance FromJSON EConfig where
...
         style = v .:? "format" >>= \case (Nothing :: Maybe String) -> ppc
                                          (Just "text")             -> ppc
-                                         (Just "json")             -> pure $ unpack . encode
-                                         (Just "none")             -> pure $ const ""
-                                         _                         -> M.fail
+                                         (Just "json")             -> pure . flip $ \g ->
```
japesinator: Nit: we shouldn't need to `flip` a lambda function, and I think that makes this code less clear

lib/Echidna/UI.hs
```diff
@@ -126,10 +126,11 @@ ui :: ( MonadCatch m, MonadRandom m, MonadReader x m, MonadUnliftIO m
    -> m Campaign
 ui v w ts d = let xfer e = use hasLens >>= \c -> isDone c >>= ($ e c) . bool id forever in do
   s <- (&&) <$> isTerminal <*> view (hasLens . dashboard)
+  g <- view (hasLens . to seed)
   c <- if s then do bc <- liftIO $ newBChan 100
                     t <- forkIO $ campaign (xfer $ liftIO . writeBChan bc) v w ts d >> pure ()
                     a <- monitor (killThread t)
-                    liftIO (customMain (mkVty defaultConfig) (Just bc) a mempty)
+                    liftIO (customMain (mkVty defaultConfig) (Just bc) a $ Campaign mempty mempty (fromJust d))
```
japesinator: Can we avoid using `fromJust`?

lib/Echidna/Campaign.hs
japesinator: Also, I don't think this should be a `Maybe Int`. I think that I'd prefer it just be a StdGen that we build intelligently at parse time. This seems like a logical mismatch as-is

lib/Echidna/UI.hs
incertia: the `fromJust` is for the `Maybe GenDict` parameter and was not my change

lib/Echidna/Config.hs
incertia: fixed

lib/Echidna/Campaign.hs
incertia: fixed

lib/Echidna/Config.hs
incertia: Changing the order of the arguments of `_finished` will avoid the `flip`, but the `flip` here makes the code more readable as we will avoid an application of the massive chain of functions to the second parameter.

incertia:
superceded by #254

PR #212, opened 2019-04-12
User defined addresses can hold and receive ether

### LINKED ISSUE #180: VM failed for unhandled reason, Query <EVM.Query: fetch contract ... >
https://solidity.readthedocs.io/en/v0.5.3/solidity-by-example.html

```
pragma solidity >=0.4.22 <0.6.0;

contract SimpleAuction {
    // Parameters of the auction. Times are either
    // absolute unix timestamps (seconds since 1970-01-01)
    // or time periods in seconds.
    address payable public beneficiary;
    uint public auctionEndTime;

    // Current state of the auction.
    address public highestBidder;
    uint public highestBid;

    // Allowed withdrawals of previous bids
    mapping(address => uint) pendingReturns;

    // Set to true at the end, disallows any change.
    // By default initialized to `false`.
    bool ended;

    // Events that will be emitted on changes.
    event HighestBidIncreased(address bidder, uint amount);
    event AuctionEnded(address winner, uint amount);

    // The following is a so-called natspec comment,
    // recognizable by the three slashes.
    // It will be shown when the user is asked to
    // confirm a transaction.

    /// Create a simple auction with `_biddingTime`
    /// seconds bidding time on behalf of the
    /// beneficiary address `_beneficiary`.
    constructor(
        uint _biddingTime,
        address payable _beneficiary
    ) public {
        beneficiary = _beneficiary;
        auctionEndTime = now + _biddingTime;
    }

    /// Bid on the auction with the value sent
    /// together with this transaction.
    /// The value will only be refunded if the
    /// auction is not won.
    function bid() public payable {
        // No arguments are necessary, all
        // information is already part of
        // the transaction. The keyword payable
        // is required for the function to
        // be able to receive Ether.

        // Revert the call if the bidding
        // period is over.
        require(
            now <= auctionEndTime,
            "Auction already ended."
        );

        // If the bid is not higher, send the
        // money back.
        require(
            msg.value > highestBid,
            "There already is a higher bid."
        );

        if (highestBid != 0) {
            // Sending back the money by simply using
            // highestBidder.send(highestBid) is a security risk
            // because it could execute an untrusted contract.
            // It is always safer to let the recipients
            // withdraw their money themselves.
            pendingReturns[highestBidder] += highestBid;
        }
        highestBidder = msg.sender;
        highestBid = msg.value;
        emit HighestBidIncreased(msg.sender, msg.value);
    }

    /// Withdraw a bid that was overbid.
    function withdraw() public returns (bool) {
        uint amount = pendingReturns[msg.sender];
        if (amount > 0) {
            // It is important to set this to zero because the recipient
            // can call this function again as part of the receiving call
            // before `send` returns.
            pendingReturns[msg.sender] = 0;

            if (!msg.sender.send(amount)) {
                // No need to call throw here, just reset the amount owing
                pendingReturns[msg.sender] = amount;
                return false;
            }
        }
        return true;
    }

    /// End the auction and send the highest bid
    /// to the beneficiary.
    function auctionEnd() public {
        // It is a good guideline to structure functions that interact
        // with other contracts (i.e. they call functions or send Ether)
        // into three phases:
        // 1. checking conditions
        // 2. performing actions (potentially changing conditions)
        // 3. interacting with other contracts
        // If these phases are mixed up, the other contract could call
        // back into the current contract and modify the state or cause
        // effects (ether payout) to be performed multiple times.
        // If functions called internally include interaction with external
        // contracts, they also have to be considered interaction with
        // external contracts.

        // 1. Conditions
        require(now >= auctionEndTime, "Auction not yet ended.");
        require(!ended, "auctionEnd has already been called.");

        // 2. Effects
        ended = true;
        emit AuctionEnded(highestBidder, highestBid);

        // 3. Interaction
        beneficiary.transfer(highestBid);
    }

    function echidna_test () external returns (bool) {
    	return true;
    }
}
```

No need for config, just `echidna-test SimpleAuction.sol` and you'll see the hang between 10-50 iterations (usually about 10).  My solc is:

`Version: 0.5.4+commit.9549d8ff.Darwin.appleclang`

ggrieco-tob:
This PR enable `deployer` and `sender` addresses to receive hold and ether. Also, a Query HEVM error is no longer fatal error: it reverts the transaction. It should fix #180.

### DIFF
--- a/examples/solidity/basic/balance.sol
+++ b/examples/solidity/basic/balance.sol
@@ -1,0 +1,11 @@
+contract C {
+  constructor() public {
+      //msg.sender.transfer(0);
+  }
+
+  function f() public {}
+  function echidna_balance() public returns (bool) { 
+      return address(msg.sender) == address(0x1) && msg.sender.balance == 42; 
+  }
+
+}
--- a/examples/solidity/basic/balance.sol
+++ b/examples/solidity/basic/balance.sol
@@ -1,11 +1,11 @@
 contract C {
   constructor() public {
-      //msg.sender.transfer(0);
+      msg.sender.transfer(0); // this contract has no ether, but this call to transfer should not fail
   }
 
   function f() public {}
   function echidna_balance() public returns (bool) { 
-      return address(msg.sender) == address(0x1) && msg.sender.balance == 42; 
+      return address(msg.sender) == address(0x42) && msg.sender.balance == 42; 
   }
 
 }
--- a/lib/Echidna/Solidity.hs
+++ b/lib/Echidna/Solidity.hs
@@ -34,10 +34,12 @@ import Echidna.Exec        (execTx)
 import Echidna.Transaction (Tx(..), World(..))
 
 import EVM hiding (contracts)
+import qualified EVM (contracts)
 import EVM.ABI      (AbiValue(..))
 import EVM.Exec     (vmForEthrunCreation)
 import EVM.Solidity
 import EVM.Types    (Addr)
+import EVM.Concrete (w256)
 
 import qualified Data.ByteString     as BS
 import qualified Data.HashMap.Strict as M
@@ -70,12 +72,13 @@ instance Show SolException where
 instance Exception SolException
 
 -- | Configuration for loading Solidity for Echidna testing.
-data SolConf = SolConf { _contractAddr :: Addr   -- ^ Contract address to use
-                       , _deployer     :: Addr   -- ^ Contract deployer address to use
-                       , _sender       :: [Addr] -- ^ Sender addresses to use
-                       , _prefix       :: Text   -- ^ Function name prefix used to denote tests
-                       , _solcArgs     :: String -- ^ Args to pass to @solc@
-                       , _quiet        :: Bool   -- ^ Suppress @solc@ output, errors, and warnings
+data SolConf = SolConf { _contractAddr    :: Addr    -- ^ Contract address to use
+                       , _deployer        :: Addr    -- ^ Contract deployer address to use
+                       , _sender          :: [Addr]  -- ^ Sender addresses to use
+                       , _initialBalance  :: Integer -- ^ Initial balance of deployer and senders
+                       , _prefix          :: Text    -- ^ Function name prefix used to denote tests
+                       , _solcArgs        :: String  -- ^ Args to pass to @solc@
+                       , _quiet           :: Bool    -- ^ Suppress @solc@ output, errors, and warnings
                        }
 makeLenses ''SolConf
 
@@ -94,6 +97,12 @@ contracts fp = do
         readSolc =<< writeSystemTempFile ""
                  =<< readCreateProcess (proc "solc" $ usual <> words a) {std_err = stderr} ""
 
+populateAddresses :: [Addr] -> Integer -> VM -> VM
+populateAddresses []     _ vm = vm
+populateAddresses (a:as) b vm = populateAddresses as b (vm & set (env . EVM.contracts . at a) (Just account))
+                                 where account = initialContract mempty & set nonce 1 & set balance (w256 $ fromInteger b)
+
+
 -- | Given an optional contract name and a list of 'SolcContract's, try to load the specified
 -- contract, or, if not provided, the first contract in the list, into a 'VM' usable for Echidna
 -- testing and extract an ABI and list of tests. Throws exceptions if anything returned doesn't look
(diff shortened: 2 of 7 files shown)


lib/Echidna/Solidity.hs
```diff
@@ -94,6 +97,12 @@ contracts fp = do
         readSolc =<< writeSystemTempFile ""
                  =<< readCreateProcess (proc "solc" $ usual <> words a) {std_err = stderr} ""
 
+populateAddresses :: [Addr] -> Integer -> VM -> VM
+populateAddresses []     _ vm = vm
+populateAddresses (a:as) b vm = populateAddresses as b (vm & set (env . EVM.contracts . at a) (Just account))
+                                 where account = initialContract mempty & set nonce 1 & set balance (w256 $ fromInteger b)
+
+
```
japesinator: nit: extra line

lib/Echidna/Solidity.hs
```diff
@@ -70,12 +72,13 @@ instance Show SolException where
...
-data SolConf = SolConf { _contractAddr :: Addr   -- ^ Contract address to use
-                       , _deployer     :: Addr   -- ^ Contract deployer address to use
-                       , _sender       :: [Addr] -- ^ Sender addresses to use
-                       , _prefix       :: Text   -- ^ Function name prefix used to denote tests
-                       , _solcArgs     :: String -- ^ Args to pass to @solc@
-                       , _quiet        :: Bool   -- ^ Suppress @solc@ output, errors, and warnings
+data SolConf = SolConf { _contractAddr    :: Addr    -- ^ Contract address to use
+                       , _deployer        :: Addr    -- ^ Contract deployer address to use
+                       , _sender          :: [Addr]  -- ^ Sender addresses to use
+                       , _initialBalance  :: Integer -- ^ Initial balance of deployer and senders
```
japesinator: is this used?

lib/Echidna/Solidity.hs
```diff
@@ -111,9 +120,9 @@ loadSpecified name cs = let ensure l e = if l == mempty then throwM e else pure
     unless q . putStrLn $ "Analyzing contract: " <> unpack (c ^. contractName)
 
   -- Local variables
-  (SolConf ca d _ pref _ _) <- view hasLens
+  (SolConf ca d ads b pref _ _) <- view hasLens
   let bc = c ^. creationCode
-      blank = vmForEthrunCreation bc
+      blank = populateAddresses (ads++[d]) b (vmForEthrunCreation bc)
```
japesinator: `ads |> b`

lib/Echidna/Solidity.hs
```diff
@@ -94,6 +97,12 @@ contracts fp = do
         readSolc =<< writeSystemTempFile ""
                  =<< readCreateProcess (proc "solc" $ usual <> words a) {std_err = stderr} ""
 
+populateAddresses :: [Addr] -> Integer -> VM -> VM
+populateAddresses []     _ vm = vm
+populateAddresses (a:as) b vm = populateAddresses as b (vm & set (env . EVM.contracts . at a) (Just account))
+                                 where account = initialContract mempty & set nonce 1 & set balance (w256 $ fromInteger b)
```
japesinator: can you align to two space indent?

src/test/Spec.hs
```diff
@@ -101,6 +101,9 @@ integrationTests = testGroup "Solidity Integration Testing"
       [ ("echidna_state failed",                  not . solved "echidna_state") ]
   , testContract "coverage/multi.sol"        Nothing
       [ ("echidna_state3 failed",                  not . solved "echidna_state3") ]
+  , testContract "basic/balance.sol"        (Just "basic/balance.yaml")
+      [ ("echidna_balance failed",                  passed       "echidna_balance") ]
```
japesinator: can we fix the indent on `passed` (and also `not . solved` on L103)?

src/test/Spec.hs
```diff
@@ -101,6 +101,9 @@ integrationTests = testGroup "Solidity Integration Testing"
...
+
```
japesinator: nit: extra line

examples/solidity/basic/balance.sol
```diff
@@ -0,0 +1,11 @@
+contract C {
+  constructor() public {
+      //msg.sender.transfer(0);
```
japesinator: let's either remove this, include it, or document why it's commented out

lib/Echidna/Solidity.hs

src/test/Spec.hs
ggrieco-tob: Fixed.

examples/solidity/basic/balance.sol
ggrieco-tob: Fixed.

lib/Echidna/Solidity.hs

src/test/Spec.hs
ggrieco-tob: Fixed.

lib/Echidna/Solidity.hs

PR #215, opened 2019-04-17
Added support to configure gas using testMaxGas and propMaxGas

ggrieco-tob:


agroce:
works for me, kills mutants, avoids memory explosions, BTW

PR #221, opened 2019-04-25
Use prefix++'_revert' as name prefix for functions that should revert

### LINKED ISSUE #214: Support properties that SHOULD revert
If property name has _reverts in it, consider result false/failed unless the transaction reverts.

Distinct from config-level allowing all props to revert or return true:

- per property avoids danger of accidental vacuity

- forces, vs. allows, revert (can simulate with return false at config level of course)

ggrieco-tob:
This fixes #214.


lib/Echidna/Exec.hs
```diff
@@ -63,11 +63,10 @@ vmExcept e = throwM $ case VMFailure e of {Illegal -> IllegalExec e; _ -> Unknow
 -- | Given an error handler, an execution function, and a transaction, execute that transaction
 -- using the given execution strategy, handling errors with the given handler.
 execTxWith :: (MonadState x m, Has VM x) => (Error -> m ()) -> m VMResult -> Tx -> m VMResult
-execTxWith h m t = do og <- get
-                      setupTx t
+execTxWith h m t = do setupTx t
                       res <- m
                       case (res, isRight $ t ^. call) of
-                        (Reversion,   _)         -> put og
+                        (Reversion,   _)         -> return ()
```
japesinator: I'm not sure this is correct. I think that if we make any state changes before encountering a revert, this will keep them. We can maybe do `put og >> result .= Just Reversion` to be more correct instead?

lib/Echidna/Test.hs
```diff
@@ -11,7 +11,7 @@ import Data.Bool (bool)
 import Data.Foldable (traverse_)
 import Data.Has (Has(..))
 import Data.Text (Text)
-import EVM
+import EVM (VMResult(..), VM)
```
japesinator: good call!

lib/Echidna/Test.hs
```diff
@@ -36,14 +36,16 @@ data CallRes = ResFalse | ResTrue | ResRevert | ResOther deriving (Eq, Show)
 classifyRes :: VMResult -> CallRes
 classifyRes (VMSuccess b) | b == encodeAbiValue (AbiBool True)  = ResTrue
                           | b == encodeAbiValue (AbiBool False) = ResFalse
+                          | otherwise                           = error "invalid result"
```
japesinator: We've been classifying these as `ResOther` so far. Do we really want to start throwing instead?

lib/Echidna/Exec.hs
ggrieco-tob: You are right. I think I fixed it. Also, @agroce should be aware of this, since echidna could execute incorrectly some code before a revert. For instance this will fail:

```
contract C { 
  bool state = true;

  function f() public {
    state = false;
    revert();
  }

 function echidna_state() public returns (bool) {
    return state;
 }
}
```

Make sure you fix it and re-run your experiments.

lib/Echidna/Test.hs
ggrieco-tob: I believe this will never happen.

japesinator: I think it could, if someone writes a test like, really wrong (with inline asm for instance)

ggrieco-tob: Fair enough. It returns `ResOther` now.

PR #227, opened 2019-05-14
Support to execute the 'init_function' function after the constructor

ggrieco-tob:
This PR will make Echidna to execute the `init_function` function (if available) after the constructor but before any other function. This function could be still re-executed, so the user is responsible to ensure that its re-execution will not cause any issue.


incertia:
we can probably close this and rethink how we want to do it

ggrieco-tob:
Fair enough.

PR #251, opened 2019-06-26
Timeout for a campaign

ggrieco-tob:
Added a timeout (in second) for a campaign. Use the keyword `timeout`. By default is `-1`, that means no timeout.

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -175,7 +175,7 @@ campaign u v w ts d = let d' = fromMaybe mempty d in fmap (fromMaybe mempty) (vi
     step        = runUpdate (updateTest v Nothing) >> u >> runCampaign
     runCampaign = use (hasLens . tests . to (fmap snd)) >>= update
     update c      = (liftIO getCPUTime) >>= \t -> view hasLens >>= \(CampaignConf tl q sl _ tm) ->
-      if | t > 0 && t >= tm                              -> u
+      if | tm > 0 && t >= tm                              -> u
          | any (\case Open  n   -> n < tl; _ -> False) c -> callseq v w q >> step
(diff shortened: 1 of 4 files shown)


lib/Echidna/Campaign.hs
```diff
@@ -46,6 +47,8 @@ data CampaignConf = CampaignConf { testLimit     :: Int
                                  , knownCoverage :: Maybe (Map W256 (Set Int))
                                    -- ^ If applicable, initially known coverage. If this is 'Nothing',
                                    -- Echidna won't collect coverage information (and will go faster)
+                                 , timeout       :: Integer
+                                   -- ^ Timeout in picoseconds
```
incertia: I don't think timeout should be in picoseconds. milliseconds or seconds seems more normal. Granted, the config option is seconds but I think that it should be reflected in `CampaignConf`.

lib/Echidna/Config.hs
```diff
@@ -56,10 +57,16 @@ instance FromJSON EConfig where
                 fprefix <- v .:? "prefix"  .!= "echidna_"
                 let goal fname = if (fprefix <> "revert_") `isPrefixOf` fname then ResRevert else ResTrue
                 return $ TestConf (\fname -> (== goal fname)  . maybe ResOther classifyRes . view result) (const psender)
+
+        timeoutp :: Y.Parser Integer
+        timeoutp = v .:? "timeout" >>= \case (Nothing :: Maybe Integer) -> pure (-1) 
+                                             (Just n) -> pure (n*1e12)
```
incertia: I think `timeoutp = ((*1e12) <$> v .:? "timeout") .!= -1` looks better here. Also I'm surprised the `\case` doesn't automatically type to `Maybe Integer` due to the `n*1e12`.

PR #258, opened 2019-07-04
Fixed input generation using seeds and its tests

ggrieco-tob:


src/test/Spec.hs
```diff
@@ -79,9 +81,16 @@ seedTests =
...
+                      cfg = defaultConfig' & cConf %~ \x -> x { seed = Just s } in 
+                  genTxs fp cfg s
+
+genTxs :: FilePath -> EConfig -> Int -> IO Tx
+genTxs fp c s =
+  flip runReaderT c $ do
+    liftIO $ setStdGen $ mkStdGen s
+    (_,w,_) <- loadSolTests fp Nothing
+    cs  <- contracts fp
+    evalStateT genTxM (w, mkGenDict 0.15 (extractConstants cs) [] s)
```
incertia: The point of reproducible seeds is also so that the random-ness no longer comes from `IO`, as seen [here](https://github.com/crytic/echidna/blob/master/lib/Echidna/Campaign.hs#L179).

japesinator:
Superseded by #259

PR #262, opened 2019-07-12
Scrape return values for generation constants

japesinator:


lib/Echidna/ABI.hs
```diff
@@ -81,7 +81,9 @@ data GenDict = GenDict { _pSynthA    :: Float
                          -- ^ Whole calls to use, sorted by type
                        , _defSeed    :: Int
                          -- ^ Default seed to use if one is not provided in EConfig
-                       } deriving (Show)
```
ggrieco-tob: No more `Show` ?

src/Main.hs
```diff
@@ -39,5 +39,5 @@ main = do Options f c conf <- execParser opts
           cpg <- flip runReaderT cfg $ do
             cs       <- contracts f
             (v,w,ts) <- loadSpecified (pack . (f ++) . (':' :) <$> c) cs >>= prepareForTest
-            ui v w ts (Just $ mkGenDict 0.15 (extractConstants cs) [] g)
+            ui v w ts (Just $ mkGenDict 0.15 (extractConstants cs) [] g (returnTypes cs))
```
ggrieco-tob: At some point we need to start adding everything extracted from the contracts into some more suitable type..

lib/Echidna/Campaign.hs
```diff
@@ -157,13 +166,18 @@ execTxOptC t = do
 -- | Given an initial 'VM' and 'World' state and a number of calls to generate, generate that many calls,
 -- constantly checking if we've solved any tests or can shrink known solves. Update coverage as a result
```
ggrieco-tob: Can you update the comments here and/or add comments in the new lines of `callseq`. This function is starting to be very difficult to understand..

lib/Echidna/ABI.hs
japesinator: There's no good way to show the return typing function, and we never use that instance (I added it to debug something way back when and forgot to remove it)

src/Main.hs
japesinator: Yes definitely! I think that escapes the scope of this PR though

PR #265, opened 2019-07-12
make multisender the default, preserving minimal ui

japesinator:


lib/Echidna/Config.hs
```diff
@@ -73,21 +73,25 @@ instance FromJSON EConfig where
...
-                "none"             -> pure . const . const $ ""
+                "none"             -> pure $ \_ _ -> ""
                 _                  -> pure $ \_ _ -> M.fail
                   "unrecognized ui type (should be text, json, or none)" in
     EConfig <$> cc
             <*> pure names
             <*> (SolConf <$> v .:? "contractAddr"   .!= 0x00a329c0648769a73afac7f9381e08fb43dbea72
                          <*> v .:? "deployer"       .!= 0x00a329c0648769a73afac7f9381e08fb43dbea70
-                         <*> v .:? "sender"         .!= [0x00a329c0648769a73afac7f9381e08fb43dbea70]
+                         <*> v .:? "sender"         .!= [ 0x00a329c0648769a73afac7f9381e08fb43dbea70
```
ggrieco-tob: I suggest to make these addresses:

(1) smaller, so they are easy to ready and
(2) very different between each other, so they are easier to read.

PR #279, opened 2019-07-28
Added support for crytic arguments

dbogs425:

PR #280, opened 2019-07-30
Added support for crytic arguments

dbogs425:
I apologize for the additional PR, I needed to re-commit to make ensure my email was associated for signing the CLA.  I made the requested changes and tested, let me know if there's any remaining issues -- thanks for the help!

Old PR: [https://github.com/crytic/echidna/pull/279](https://github.com/crytic/echidna/pull/279)


incertia:
@ggrieco-tob unless crytic can handle multiple `--solc-args` directives, this would be a regression, as `echidna` passes its own `--solc-args` as well.

incertia:
After further internal discussion, we decided that we will keep `solcArgs` for legacy support and then deprecate it after a few releases. I'll be merging this now.

PR #284, opened 2019-08-12
Introduce ABIv2 support into Echidna

incertia:
A lot of the code as it currently stands is pulled directly from HEVM with enough modification to hack it to work with ABIv2 tuples.

What works:
* Tuples classified as "static" by the ABI
* Tuples classified as "dynamic" by the ABI (probably)
* Shrinking of tuples
* Decoding `solc` output to the correct tuple type for use in the ABI
* Call sequence generation, or the hard part of what Echidna does
* Storage of return values as constants for callseq generation

What doesn't work:
* UI is broken for super long lines introduced by massive tuples

What should change:
* JSON parsing is super wonky
* Cleanup of `ABIv2.hs`

tl;dr it basically works.

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE DeriveAnyClass #-}
-{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -115,10 +114,10 @@ deriving instance Hashable AbiType
 
 -- | Construct a 'GenDict' from some dictionaries, a 'Float', a default seed, and a typing rule for
 -- return values
-mkGenDict :: Float       -- ^ Percentage of time to mutate instead of synthesize. Should be in [0,1]
-          -> [AbiValue]  -- ^ A list of 'AbiValue' constants to use during dictionary-based generation
-          -> [SolCall]   -- ^ A list of complete 'SolCall's to mutate
-          -> Int         -- ^ A default seed
+mkGenDict :: Float      -- ^ Percentage of time to mutate instead of synthesize. Should be in [0,1]
+          -> [AbiValue] -- ^ A list of 'AbiValue' constants to use during dictionary-based generation
+          -> [SolCall]  -- ^ A list of complete 'SolCall's to mutate
+          -> Int        -- ^ A default seed
           -> (Text -> Maybe AbiType)
           -- ^ A return value typing rule
           -> GenDict
@@ -208,16 +207,16 @@ mutateV t v = traverse mutateAbiValue =<< changeSize where
 
 -- | Check if an 'AbiValue' is as \"small\" (trivial) as possible (using ad-hoc heuristics).
 canShrinkAbiValue :: AbiValue -> Bool
-canShrinkAbiValue (AbiUInt _ 0) = False
-canShrinkAbiValue (AbiInt  _ 0) = False
-canShrinkAbiValue (AbiBool b) = b
-canShrinkAbiValue (AbiBytes _ b)       = BS.any (/= 0) b
-canShrinkAbiValue (AbiBytesDynamic "") = False
-canShrinkAbiValue (AbiString "")       = False
+canShrinkAbiValue (AbiUInt _ 0)         = False
+canShrinkAbiValue (AbiInt  _ 0)         = False
+canShrinkAbiValue (AbiBool b)           = b
+canShrinkAbiValue (AbiBytes _ b)        = BS.any (/= 0) b
+canShrinkAbiValue (AbiBytesDynamic "")  = False
+canShrinkAbiValue (AbiString "")        = False
 canShrinkAbiValue (AbiArray _ _ l)      = any canShrinkAbiValue l
 canShrinkAbiValue (AbiArrayDynamic _ l) = l /= mempty
-canShrinkAbiValue (AbiTuple v) = any canShrinkAbiValue v
-canShrinkAbiValue _ = True
+canShrinkAbiValue (AbiTuple v)          = any canShrinkAbiValue v
+canShrinkAbiValue _                     = True
 
 bounds :: forall a. (Bounded a, Integral a) => a -> (Integer, Integer)
 bounds = const (fromIntegral (0 :: a), fromIntegral (maxBound :: a))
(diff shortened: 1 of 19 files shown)


lib/Echidna/ABI.hs
```diff
@@ -224,15 +229,19 @@ shrinkInt x | x == -1   = pure 0
...
 shrinkAbiValue (AbiBool _)    = pure $ AbiBool False
 shrinkAbiValue (AbiBytes n b)      = AbiBytes n <$> addNulls b
 shrinkAbiValue (AbiBytesDynamic b) = fmap AbiBytesDynamic $ addNulls =<< shrinkBS b
 shrinkAbiValue (AbiString b)       = fmap AbiString       $ addNulls =<< shrinkBS b
-shrinkAbiValue (AbiArray n t l)      = AbiArray n t <$> traverse shrinkAbiValue l
+shrinkAbiValue (AbiArray n t l)    = AbiArray n t <$> traverse shrinkAbiValue l
 shrinkAbiValue (AbiArrayDynamic t l) = fmap (AbiArrayDynamic t) $ traverse shrinkAbiValue =<< shrinkV l
+shrinkAbiValue (AbiTuple v)   = AbiTuple <$> traverse shrinkAbiValue' v
+  where shrinkAbiValue' x = do
+          f <- uniform [return, shrinkAbiValue]
```
japesinator: `liftM3 bool (pure x) (shrinkAbiValue x) getRandom`

japesinator: let's also use this for array shrinking! it's probably useful there too

lib/Echidna/ABI.hs
```diff
@@ -277,8 +287,9 @@ genAbiValueM = genWithDict (view constants) $ \case
   AbiStringType       -> liftM2 (\n -> AbiString       . BS.pack . take n)
                                 (getRandomR (1, 32)) getRandoms
   (AbiArrayDynamicType t) -> fmap (AbiArrayDynamic t) $ getRandomR (1, 32)
-                             >>= flip V.replicateM (genAbiValueM t)
+                              >>= flip V.replicateM (genAbiValueM t)
```
japesinator: nit: deindent

lib/Echidna/Campaign.hs
```diff
@@ -64,9 +64,9 @@ data CampaignConf = CampaignConf { testLimit     :: Int
 
 -- | State of a particular Echidna test. N.B.: \"Solved\" means a falsifying call sequence was found.
 data TestState = Open Int             -- ^ Maybe solvable, tracking attempts already made
-               | Large Int [Tx]       -- ^ Solved, maybe shrinable, tracking shrinks tried + best solve
+               | Large Int [Tx]      -- ^ Solved, maybe shrinable, tracking shrinks tried + best solve
```
japesinator: nit: fix indent

lib/Echidna/Campaign.hs
```diff
@@ -88,7 +88,7 @@ instance ToJSON TestState where
 -- | The state of a fuzzing campaign.
 data Campaign = Campaign { _tests    :: [(SolTest, TestState)] -- ^ Tests being evaluated
                          , _coverage :: Map W256 (Set Int)     -- ^ Coverage captured (NOTE: we don't always record this)
-                         , _genDict  :: GenDict                -- ^ Generation dictionary
+                         , _genDict  :: GenDict               -- ^ Generation dictionary
```
japesinator: indent

lib/Echidna/Solidity.hs
```diff
@@ -7,7 +7,7 @@
 
 module Echidna.Solidity where
 
-import Control.Lens
+import Control.Lens hiding        (Indexed)
```
japesinator: why hide this?

lib/Echidna/Solidity.hs
```diff
@@ -29,15 +29,16 @@ import Data.Text.Read             (decimal)
 import System.Process             (StdStream(..), readCreateProcess, proc, std_err)
 import System.IO                  (openFile, IOMode(..))
 
-import Echidna.ABI         (SolSignature)
+import Echidna.ABI         (SolSignature, SolSignature)
+import Echidna.ABIv2       (AbiType(..), AbiValue(..), readSolc, SolcContract,
```
japesinator: we can probably just import this entirely

lib/Echidna/Solidity.hs
```diff
@@ -161,12 +162,11 @@ loadSpecified name cs = let ensure l e = if l == mempty then throwM e else pure
   -- Make sure everything is ready to use, then ship it
   mapM_ (uncurry ensure) $ [(abi, NoFuncs), (funs, OnlyTests)]
                         ++ if ch then [] else [(tests, NoTests)] -- ABI checks
-  ensure bc (NoBytecode $ c ^. contractName)                     -- Bytecode check
+  ensure bc (NoBytecode $ c ^. contractName)                    -- Bytecode check
```
japesinator: indent

lib/Echidna/Transaction.hs
```diff
@@ -63,38 +63,38 @@ makeLenses 'TxConf
...
-
 -- | A contract is just an address with an ABI (for our purposes).
 type ContractA = (Addr, [SolSignature])
 
 -- | The world is made our of humans with an address, and contracts with an address + ABI.
 data World = World { _senders   :: [Addr]
                    , _receivers :: [ContractA]
                    }
-
+  deriving (Show)
```
japesinator: are we using this instance? if so, we should probably make it somewhat more clear by writing it explicitly

lib/Echidna/Transaction.hs
```diff
@@ -63,38 +63,38 @@ makeLenses 'TxConf
...
 makeLenses ''World
 
 -- | Given generators for an origin, destination, value, and function call, generate a call
 -- transaction. Note: This doesn't generate @CREATE@s because I don't know how to do that at random.
 genTxWith :: (MonadRandom m, MonadState x m, Has World x, MonadThrow m) 
           => ([Addr] -> m Addr)                       -- ^ Sender generator
-          -> ([ContractA] -> m ContractA)             -- ^ Receiver generator
+          -> ([ContractA] -> m ContractA)           -- ^ Receiver generator
```
japesinator: indent

lib/Echidna/Solidity.hs
incertia: there's an `Indexed` from HEVM that needs using somewhere.

incertia: but that was from when part of the code lived in Solidity.hs, i can get rid of this now

lib/Echidna/Transaction.hs
incertia: we don't use this instance but it probably doesn't hurt to have right?

japesinator: if we're going to ship an instance, it should be a good instance. I don't think the auto-derived one is good

lib/Echidna/ABI.hs
```diff
@@ -249,6 +256,7 @@ mutateAbiValue (AbiBytesDynamic b)   = AbiBytesDynamic   <$> mutateBS b
 mutateAbiValue (AbiString b)         = AbiString         <$> mutateBS b
 mutateAbiValue (AbiArray n t l)      = AbiArray n t      <$> traverse mutateAbiValue l
 mutateAbiValue (AbiArrayDynamic t l) = AbiArrayDynamic t <$> mutateV t l
+mutateAbiValue (AbiTuple v)  = AbiTuple <$> traverse mutateAbiValue v
```
japesinator: nit: indent

lib/Echidna/ABI.hs
```diff
@@ -279,6 +287,7 @@ genAbiValueM = genWithDict (view constants) $ \case
   (AbiArrayDynamicType t) -> fmap (AbiArrayDynamic t) $ getRandomR (1, 32)
                              >>= flip V.replicateM (genAbiValueM t)
   (AbiArrayType n t)      -> AbiArray n t <$> V.replicateM n (genAbiValueM t)
+  (AbiTupleType v)    -> AbiTuple <$> traverse genAbiValueM v
```
japesinator: indent

lib/Echidna/ABI.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
```
japesinator: is this used?

incertia: this can get deleted

PR #295, opened 2019-09-11
Merge contracts created by function calls into the address pool

incertia:
`GenDict` is also updated to internally use `HashSet`.


lib/Echidna/Campaign.hs
```diff
@@ -180,22 +181,31 @@ callseq :: ( MonadCatch m, MonadRandom m, MonadReader x m, MonadState y m
 callseq v w ql = do
   -- First, we figure out whether we need to execute with or without coverage optimization, and pick
   -- our execution function appropriately
+  let old = v ^. env . EVM.contracts
+  --hasLens . genDict %= gaddConstants (AbiAddress . addressWord160 <$> traceShow addrs addrs)
```
ggrieco-tob: @incertia just remove this line, and this PR should be ready to mege

src/test/Spec.hs
```diff
@@ -193,19 +199,23 @@ integrationTests = testGroup "Solidity Integration Testing"
...
   flip runReaderT c $ do
     g <- getRandom
-    (v,w,ts) <- loadSolTests fp Nothing
-    cs  <- contracts fp
-    ads <- addresses
-    campaign (pure ()) v w ts (Just $ mkGenDict 0.15 (extractConstants cs ++ NE.toList ads) [] g (returnTypes cs))
+    (v,w,ts) <- loadSolTests fp n
+    cs  <- Echidna.Solidity.contracts fp
+    ads <- NE.toList <$> addresses
+    let ads' = AbiAddress . addressWord160 <$> v ^. env . EVM.contracts . to keys
```
ggrieco-tob: It seems this is used in several places of the code. Maybe we can have a function for this?

lib/Echidna/Campaign.hs
```diff
@@ -187,21 +188,29 @@ callseq v w ql = do
...
   -- We then run each call sequentially. This gives us the result of each call, plus a new state
   (res, s) <- runStateT (evalSeq v ef is) (v, ca)
+  let new = s ^. _1 . env . EVM.contracts
+      diff = keys $ new \\ old
+      diffs = H.fromList [(AbiAddressType, S.fromList $ AbiAddress . addressWord160 <$> diff)]
   -- Save the global campaign state (also vm state, but that gets reset before it's used)
   hasLens .= snd s
   -- Now we try to parse the return values as solidity constants, and add then to the 'GenDict'
-  modifying (hasLens . constants) . H.unionWith (++) . parse res =<< use (hasLens . rTypes) where
+  types <- use $ hasLens . rTypes
```
ggrieco-tob: Can you add a few comments in `callseq` to explain how the address collection works?

src/test/Spec.hs
incertia: let's sort this out after this PR

PR #297, opened 2019-09-13
Use corpusDir option to save/load whole calls discovered during a fuzzing campaign

ggrieco-tob:


ggrieco-tob:
Superseded by #353

PR #299, opened 2019-09-15
Fix default.yaml config

arcz:
I've discovered that `basic/default.yaml` config won't parse due to `solcLibs` being string. I've changed that to the proper type and created a test to have it in sync, since it seems to be a sample config mentioned in README.

I've added `basic/config.yaml` to tests as well but it looks unused, should this be removed? Rest of configs are already parsed with their sol tests.

Additionally, the range option also is unused but parser accepts all the additional keys as long as the required are present. Maybe it would be good to force the parser to reject a config with extra keys?

### DIFF
--- a/examples/solidity/basic/config.yaml
+++ b/examples/solidity/basic/config.yaml
@@ -1,2 +1,1 @@
 testLimit: 100000000
-range: 10
--- a/examples/solidity/basic/default.yaml
+++ b/examples/solidity/basic/default.yaml
@@ -28,7 +28,7 @@ balanceContract: 0
 #solcArgs allows special Args to solc
 solcArgs: ""
 #solcLibs is solc libraries
-solcLibs: ""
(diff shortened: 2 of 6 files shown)


ggrieco-tob:
Thanks for the cleaning up the .yaml files and adding the parse tests. We will internally discuss this on Monday and merge this next week.

ggrieco-tob:
PS: try to avoid `force-pushed` commits :smiley:

agroce:
Rejecting extra keys might cause some problems, but warning seems like a reasonable thing to do.

incertia:
We can definitely warn on unused keys. I will draft a PR that lifts all the "aeson key tracking" stuff into a `StateT` so the resulting code remains mostly the same.

incertia:
> I will draft a PR
#301 now does this.

Issue #208, opened 2019-04-11
Echidna may be ignoring coverage configuration


agroce:
Runs with

`coverage: false`

in the .yaml config file seem to be identical to runs without it.  They show instruction count, which means coverage is being measured (which is quite costly).  Problem for benchmarking as well.

incertia:
solved by #300

PR #300, opened 2019-09-17
Fix coverage optimization and configuration

arcz:
This PR fixes 2 things:
1. Wrong execution function choice in `callseq` which resulted in coverage being collected all the time. I believe it could be due to a bit confusing `bool` function. It picked in fact `execTxOptC` when `knownCoverage` was `Nothing`. I've replaced it with a more explicit condition.
2. The `coverage: true` JSON config was not being respected. I've added correct parsing with some tests. This solves: https://github.com/crytic/echidna/issues/208

The test suite is running now in around 100s on my machine vs around 140s before the fix. Also `echidna-test` should be faster by default from now on.

### DIFF
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -16,6 +16,7 @@ import Data.ByteString.Lazy.Char8 (unpack)
 import Data.Has (Has(..))
 import Data.Aeson
 import Data.Aeson.Lens
+import Data.Functor ((<&>))
 import Data.Text (isPrefixOf)
 import EVM (result)
 import EVM.Concrete (Word(..), Whiff(..))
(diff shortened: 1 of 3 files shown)


lib/Echidna/Config.hs
```diff
@@ -68,10 +69,12 @@ instance FromJSON EConfig where
         getWord s d = C Dull . fromIntegral <$> v .:? s .!= (d :: Integer)
         xc = liftM4 TxConf (getWord "propMaxGas" 8000030) (getWord "testMaxGas" 0xffffffff)
                            (getWord "maxTimeDelay" 604800)     (getWord "maxBlockDelay" 60480)
+        coverageParser = v .:? "coverage" <&> \case Just True -> Just mempty
```
incertia: I think the wording `coverageParser` itself can be a bit confusing. I would probably rename it to just `coverage` or `cov` for short.

arcz: Good point, I was also not sure about this name, changed to `cov`. Now I think it could be even put inline but it is a minor difference.

incertia:
Merging early because the added tests have passed and no other tested functionality was modified.

PR #301, opened 2019-09-18
Warn on unused config keys

incertia:
Strategy: lift `EConfig` into `EConfig2`, which is JSON parsed with a `StateT` that tracks the used JSON keys. This helps avoid error when adding/removing keys to the `EConfig` parser. The `FromJSON EConfig` instance is modified to pull from `EConfig2` and `parseConfig` now returns `EConfig2` instead.


lib/Echidna/Config.hs
```diff
@@ -40,6 +46,11 @@ data EConfig = EConfig { _cConf :: CampaignConf
                        }
 makeLenses ''EConfig
 
+data EConfig2 = EConfig2 { _econfig :: EConfig
```
ggrieco-tob: Can we have a better name for `Econfig2`, maybe `EconfigInternal` or something like that?

src/test/Spec.hs
```diff
@@ -38,10 +39,15 @@ configTests :: TestTree
 configTests = testGroup "Configuration tests" $
   [ testCase file $ void $ parseConfig file | file <- files ] ++
   [ testCase "parse \"coverage: true\"" $ do
-      config <- parseConfig "coverage/test.yaml"
+      config <- _econfig <$> parseConfig "coverage/test.yaml"
       assertCoverage config $ Just mempty
   , testCase "coverage disabled by default" $
       assertCoverage defaultConfig Nothing
+  , testCase "defaults.yaml" $ do
```
incertia: this test case should fix our woes with people forgetting to modify `default.yaml`

src/test/Spec.hs
```diff
@@ -38,10 +39,15 @@ configTests :: TestTree
...
+      EConfigWithUsage _ bad unset <- parseConfig "basic/default.yaml"
+      assertBool ("unused options: " ++ show bad) $ null bad
+      let unset' = unset `HS.difference` HS.fromList ["seed"]
+      assertBool ("unset options: " ++ show unset') $ null unset'
```
incertia: this test case should fix our woes with people forgetting to modify `default.yaml`

src/Main.hs
```diff
@@ -44,7 +46,8 @@ opts = info (helper <*> versionOption <*> options) $ fullDesc
 main :: IO ()
 main = do Options f c conf <- execParser opts
           g   <- getRandom
-          cfg <- maybe (pure defaultConfig) parseConfig conf
+          EConfigWithUsage cfg ks _ <- maybe (pure (EConfigWithUsage defaultConfig (fromList []) (fromList []))) parseConfig conf
+          mapM_ (hPutStrLn stderr . ("Warning: unused option: " ++) . unpack) ks
```
japesinator: can we silence this if `quiet` is enabled?

src/Main.hs
```diff
@@ -44,7 +46,8 @@ opts = info (helper <*> versionOption <*> options) $ fullDesc
 main :: IO ()
 main = do Options f c conf <- execParser opts
           g   <- getRandom
-          cfg <- maybe (pure defaultConfig) parseConfig conf
+          EConfigWithUsage cfg ks _ <- maybe (pure (EConfigWithUsage defaultConfig (fromList []) (fromList []))) parseConfig conf
```
japesinator: maybe s/`fromList []`/`mempty`

src/test/Spec.hs
```diff
@@ -38,10 +39,15 @@ configTests :: TestTree
...
   [ testCase "parse \"coverage: true\"" $ do
-      config <- parseConfig "coverage/test.yaml"
+      config <- _econfig <$> parseConfig "coverage/test.yaml"
       assertCoverage config $ Just mempty
   , testCase "coverage disabled by default" $
       assertCoverage defaultConfig Nothing
+  , testCase "defaults.yaml" $ do
+      EConfigWithUsage _ bad unset <- parseConfig "basic/default.yaml"
+      assertBool ("unused options: " ++ show bad) $ null bad
+      let unset' = unset `HS.difference` HS.fromList ["seed"]
```
japesinator: `HS.singleton`

japesinator: or if you wanna be fancy, I think `Control.Lens.sans` might be nicer

lib/Echidna/Config.hs
```diff
@@ -61,63 +76,74 @@ instance Has UIConf EConfig where
...
-                         <*> v .:? "solcLibs"       .!= []
-                         <*> v .:? "quiet"          .!= False
-                         <*> v .:? "checkAsserts"   .!= False)
-            <*> tc
-            <*> xc
-            <*> (UIConf <$> v .:? "dashboard" .!= True <*> v .:? "timeout" <*> style)
-  parseJSON _ = parseJSON (Object mempty)
+  parseJSON = fmap _econfig . parseJSON
+
+instance FromJSON EConfigWithUsage where
```
japesinator: this needs some comments explaining the architecture

Issue #304, opened 2019-09-19
Echidna generates only transactions with tx.gasprice == 0


ggrieco-tob:
Echidna will never find the inputs to make this property fail:

```solidity
contract C {
  bool state = true;
  function f() public {
    if (tx.gasprice > 0)
      state = false;

  }

  function echidna_state() public returns (bool) {
    return state;
  }

}
```

ggrieco-tob:
This is already fixed in #307.

PR #307, opened 2019-09-30
Fuzz tx.gasprice

arcz:
This is an attempt to solve https://github.com/crytic/echidna/issues/304. The test from the issue is included. I'm not entirely sure about the default for `maxGasprice`.

I didn't implement properly the `spliceTxs` part as the function is unused.

UI preview (examples/solidity/basic/gasprice.sol):
```
┌──────────Echidna──────────┐
│  echidna_state: failed!💥  │
│    Call sequence:         │
│      f() Gas price: 0x1   │
│                           │
└───────────────────────────┘
Campaign complete, C-c or esc to print report
```

### DIFF
--- a/examples/solidity/basic/gasprice.sol
+++ b/examples/solidity/basic/gasprice.sol
@@ -1,0 +1,12 @@
+contract C {
+  bool state = true;
+
+  function f() public {
+    if (tx.gasprice > 0)
+      state = false;
+  }
+
+  function echidna_state() public returns (bool) {
+    return state;
+  }
+}
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -8,7 +8,7 @@
 module Echidna.Config where
 
 import Control.Lens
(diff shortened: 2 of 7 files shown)


ggrieco-tob:
The code looks fine, but you include a bunch of stylistic changes that make the review more difficult than expected. Can we revert the unrelated changes?

arcz:
@ggrieco-tob Sure, I've reduced the diff considerably. Hope it is easier to review now. I can put back the nonsemantic changes in a separate PR if that is desirable.

ggrieco-tob:
It looks good now!, please make a separated PR with the style changes that you propose. Include a description of each one and why you think they are useful.

ggrieco-tob:
Regarding `maxGasprice`, take a look to [here](https://ethgasstation.info/) to look for sensible values.

arcz:
I've set 100 Gwei but it looks like half of this should be enough, or even less, what do you think?

PR #309, opened 2019-10-05
Basic msg.value fuzzing

ggrieco-tob:
This PR will make any transfer to have a positive `msg.value` but only 1/100 of the time.

### DIFF
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -8,7 +8,7 @@
 module Echidna.Config where
 
 import Control.Lens
-import Control.Monad (liftM2, liftM4)
+import Control.Monad (liftM2, liftM5)
 import Control.Monad.Catch (MonadThrow)
 import Control.Monad.IO.Class (MonadIO(..))
 import Control.Monad.Reader (Reader, ReaderT(..), runReader)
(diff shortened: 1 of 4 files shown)


lib/Echidna/Config.hs
```diff
@@ -67,8 +67,9 @@ instance FromJSON EConfig where
                 return $ TestConf (\fname -> (== goal fname)  . maybe ResOther classifyRes . view result)
                                   (const psender)
         getWord s d = C Dull . fromIntegral <$> v .:? s .!= (d :: Integer)
-        xc = liftM4 TxConf (getWord "propMaxGas" 8000030) (getWord "testMaxGas" 0xffffffff)
-                           (getWord "maxTimeDelay" 604800)     (getWord "maxBlockDelay" 60480)
+        xc = liftM5 TxConf (getWord "propMaxGas" 8000030) (getWord "testMaxGas" 0xffffffff)
+                           (getWord "maxTimeDelay" 604800)     (getWord "maxBlockDelay" 60480) (getWord "maxValue" 0xffffffff)
```
incertia: there's no reason to have all these spaces here, maybe we should make this 1 arg per line with applicative instead now?

lib/Echidna/Transaction.hs
```diff
@@ -109,12 +111,17 @@ genTx = use (hasLens :: Lens' y World) >>= evalStateT genTxM . (defaultDict,)
...
-genTxM = view hasLens >>= \(TxConf _ g t b) -> genTxWith
+genTxM = view hasLens >>= \(TxConf _ g t b v) -> genTxWith
   (rElem "sender list") (rElem "recipient list")                             -- src and dst
   (const $ genInteractionsM . snd)                                           -- call itself
-  (pure g) (\_ _ _ -> pure 0) (level <$> liftM2 (,) (inRange t) (inRange b)) -- gas, value, delay
+  (pure g) (genValue v) (level <$> liftM2 (,) (inRange t) (inRange b)) -- gas, value, delay
      where inRange hi = w256 . fromIntegral <$> getRandomR (0 :: Integer, fromIntegral hi)
 
+genValue :: (MonadRandom m, MonadState x m, Has World x, MonadThrow m) => Word -> Addr -> ContractA -> SolCall -> m Word
+genValue v _ _ _ = w256 . fromIntegral <$> oneIn 100 (return 0) (getRandomR (0 :: Integer, toInteger v))
```
incertia: i don't like hard coding this to 100

ggrieco-tob: This can be another parameter in a campaign if you want. A good default value could be determined using a benchmark of contracts.

ggrieco-tob:
For some reason, this branch will not solve this example:

```solidity
contract B {
    constructor() payable public {}
}

contract A {
    B b;
    function f() payable public {
        if (msg.value <= 0) {
            revert();
        }
        b = (new B).value(msg.value)();
        if (address(b) == address(0)) {
            revert();
        }

    }

      return address(b) == address(0);
    }

}
```

It is unclear if it is a bug in Echidna or in hevm.. :disappointed:

ggrieco-tob:
This PR won't be merged as it is, since we are looking for better options using a slither printer to obtain the list of payable functions.

Issue #308, opened 2019-10-04
Echidna shrinker (too) often fails to remove un-needed gas price


agroce:
Example contract:

```
contract TEST {
  uint256 constant A = 999120;
  uint256 n = 0;

  function f (uint256 newn) external {
    n = newn;
  }

  function echidna_checkIt() external view returns (bool) {
    return (n < A);
  }
}
```

Obviously, you don't need a non-zero delay or any sender changes here.  But running with completely default echidna settings, 17% of the time, echidna reports a non-zero Time or Block delay in the found sequence.

There are no bogus sender values that didn't shrink away, but those can be produced using this contract plus assertion checking, and likely arise from same shrinker inadequacies:

```
contract TEST {
  uint256 constant A = 999120;
  uint256 n = 0;

  function f (uint256 newn) external {
    if (newn >= 0) {
      n = newn;
    }
  }

  function checkIt() external view returns (bool) {
    assert (n < A);
    return true;
  }
  
  function crytic_notrans() external view returns (bool) {
    return false;
  } 
}
```

agroce:
This is moderately important, in that I think crytic users especially may be very confused if a sequence for a property violation includes a time delay or changed sender when this is actually totally irrelevant.  I could see this being a big frustration to understanding a bug (in contract or in a property) if you aren't savvy to echidna internals

agroce:
Does #310 fix this?  Ready to close?

arcz:
@agroce yes, this should be fixed

agroce:
I did a pretty thorough check, it's fixed.

agroce:
Reopening this, with small shrink limits, or detection near timeout, gas prices often stick around, even though 99% of failures don't need gas price changes.  We should probably just deterministically try 0 gas price & delay etc. at the start of reduction, since it usually will work and is cheap

incertia:
are we able to close this now?

agroce:
I haven't tested it out, is everything in master, so I can see if it fails to get rid of gas/delay/sender more than feels right?

agroce:
(or if @arcz tested for this on this contract or similar, we can just close, of course!  I don't know if this was a driving force in his work...)

arcz:
Yes, getting to 0 is faster now.

PR #310, opened 2019-10-06
Improve transaction shrinking

arcz:
I came up with this PR after investigating https://github.com/crytic/echidna/issues/308.

The problem is that all components of a transaction are shrunk at the same time.

In the example from the issue, extracted constants (hence the mentioned probability) were already producing a minimal value for ABI calls and further shrinking of this value no longer made the property to fail. This caused the other shrinkable components (time and block delay) to halt. The problem will generally occur when shrinkage of a components flips a property back to success but there are other shrinkable components.

The fix I've made is to randomly choose a component to shrink so that their shrinking is independent.

### DIFF
--- a/lib/Echidna/Transaction.hs
+++ b/lib/Echidna/Transaction.hs
@@ -14,7 +14,7 @@ import Prelude hiding (Word)
 import Control.Lens
 import Control.Monad (join, liftM2, liftM3, liftM5)
 import Control.Monad.Catch (MonadThrow)
-import Control.Monad.Random.Strict (MonadRandom, getRandomR)
+import Control.Monad.Random.Strict (MonadRandom, getRandomR, uniform)
 import Control.Monad.Reader.Class (MonadReader)
 import Control.Monad.State.Strict (MonadState, State, evalStateT, runState)
 import Data.Aeson (ToJSON(..), object)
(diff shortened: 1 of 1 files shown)


ggrieco-tob:
We discuss this internally and arrived to the same conclusion. This looks like a clean PR, great work!

agroce:
Does this up # of allowed shrinking runs, since for things with lots of shrinkable components, this should take longer on average...

ggrieco-tob:
Maybe we should combine both approaches to shrinking?

arcz:
I'm not sure how to measure the impact of this change on number of runs, I doubt it is as trivial as 3x. We still have the `shrinkLimit` config option to increase and we can bump the default if we see the shrinking is generally taking more steps.

lib/Echidna/Transaction.hs
```diff
@@ -124,10 +124,14 @@ canShrinkTx _                                = True
 -- | Given a 'Transaction', generate a random \"smaller\" 'Transaction', preserving origin,
 -- destination, value, and call signature.
 shrinkTx :: MonadRandom m => Tx -> m Tx
-shrinkTx (Tx c s d g (C _ v) (C _ t, C _ b)) = let
+shrinkTx tx'@(Tx c _ _ _ (C _ v) (C _ t, C _ b)) = let
   c' = either (fmap Left . shrinkAbiCall) (fmap Right . pure) c
   lower x = w256 . fromIntegral <$> getRandomR (0 :: Integer, fromIntegral x) in
-    liftM5 Tx c' (pure s) (pure d) (pure g) (lower v) <*> fmap level (liftM2 (,) (lower t) (lower b))
+  sequence
```
japesinator: we can maybe rewrite this a bit cleaner: `uniform [set call =<< c', set value...] <*> pure tx'`

arcz: I've cleaned this up a bit

PR #316, opened 2019-10-20
Abort testing if constructor requires arguments

ggrieco-tob:
This small PR will be useful to warn users about arguments in the contract constructor.


lib/Echidna/Solidity.hs
```diff
@@ -67,6 +68,8 @@ instance Show SolException where
     NoFuncs              -> "ABI is empty, are you sure your constructor is right?"
     NoTests              -> "No tests found in ABI"
     OnlyTests            -> "Only tests and no public functions found in ABI"
+    (ConstructorArgs s)  -> "Constructor arguments are requiered: " ++ s
```
incertia: s/requiered/required/

PR #319, opened 2019-10-25
Automatic corpus collection

ggrieco-tob:
The new config keyword `corpusDir` allows to define a directory that will be used to save all the sequences of transactions that generated new coverage.


ggrieco-tob:
Superseded by #353

PR #320, opened 2019-10-25
Implement configurable timeouts

japesinator:


lib/Echidna/UI.hs
```diff
@@ -139,15 +148,22 @@ ui :: ( MonadCatch m, MonadRandom m, MonadReader x m, MonadUnliftIO m
...
+                  getSeed = view $ hasLens . to seed . non (d' ^. defSeed) in do
+  bc <- liftIO $ newBChan 100
+  t    <- forkIO (void $ campaign (xfer bc) v w ts d)
+  dash <- liftM2 (&&) isTerminal $ view (hasLens . dashboard)
+  done <- views hasLens $ \cc -> flip runReader (cc :: CampaignConf) . isDone
+  app  <- customMain (mkVty defaultConfig) (Just bc) <$> monitor
+  time <- views (hasLens . maxTime) . maybe (fmap Just) $ timeout . (* 1000000)
+  res  <-  liftIO . time $ if dash
+    then fst <$> app (defaultCampaign, Uninitialized)
+    else let go = readBChan bc >>= \c -> if done c then pure c else go in go
```
arcz: I'm wondering how much CPU time is consumed by the loop here. Maybe we could put the main thread to sleep while waiting for worker to signal completion/timeout? Here is an example how to wake up a thread with `MVar` https://stackoverflow.com/questions/28169297/how-to-force-main-thread-to-wait-for-all-its-child-threads-finish-in-haskell, however this might require a bigger change.

japesinator:
Re: `testLimit`, I think it's still probably the best option for most people. I don't think most dashboard users will ever use the timeout feature, and most runs should still be ~minutes.

Re: the loop just keeping the BChan empty, the problem is getting campaign state when the campaign thread is killed without warning. We need to push recent campaigns to some kind of shared channel/mvar type thing, or use richer signaling than `timeout` to keep state synchronized. IMO at that point, we should probably change how `campaign`'s instrumentation works too (we probably don't need to call `xfer` every callseq, and that's probably way more of an impact). If you have something specific you'd like to see before we merge, I'm happy to impl that, or we can merge as-is and then upgrade later

arcz:
Oh I missed that `readBChan` will block, so that shouldn't be a problem. I think we can go ahead and merge and improve from there once we figure out the thread matters.

Issue #225, opened 2019-05-10
Support early termination on property failure


agroce:
Critical for easy effective use of mutation analysis.  Handled via ad hoc scripts watching dashboard output and killing, now, which is a mess

agroce:
Done in #323

PR #323, opened 2019-11-14
Allow user to request termination as soon as any property fails

agroce:
Address #225 -- useful for mutation, potentially for long-running crytic tests?

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -115,8 +115,8 @@ defaultCampaign = Campaign mempty mempty defaultDict
 -- | Given a 'Campaign', checks if we can attempt any solves or shrinks without exceeding
 -- the limits defined in our 'CampaignConf'.
 isDone :: (MonadReader x m, Has CampaignConf x) => Campaign -> m Bool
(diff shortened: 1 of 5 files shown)


ggrieco-tob:
Don't forget to update the `default.yaml` file.

lib/Echidna/Campaign.hs
```diff
@@ -114,7 +116,8 @@ defaultCampaign = Campaign mempty mempty defaultDict
 -- the limits defined in our 'CampaignConf'.
 isDone :: (MonadReader x m, Has CampaignConf x) => Campaign -> m Bool
 isDone (view tests -> ts) = view (hasLens . to (liftM2 (,) testLimit shrinkLimit)) <&> \(tl, sl) ->
-  all (\case Open i -> i >= tl; Large i _ -> i >= sl; _ -> True) $ snd <$> ts
+  (stopOnFail && (any (\case Solved _ -> True; Failed _ -> True; _ -> False) $ snd <$> ts)) ||
```
japesinator: why not `if stopOnFail then any else all`

PR #335, opened 2020-01-06
Improved error handling when crytic-compile is missing or returns an error

### LINKED ISSUE #334: Improve missing crytic-compile error message
If `crytic-compile` is not installed, echidna-test fails and prints:
```
echidna-test: crytic-compile: readCreateProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

It would be nice to have a more user-friendly message reported, like 
```
crytic-compile is missing. To install it, run:
   pip install crytic-compile
```

ggrieco-tob:
This small PR fixes #334 and also allows to properly show the compilation failure error when crytic-compile fails to compile a contract.

### DIFF
--- a/lib/Echidna/Solidity.hs
+++ b/lib/Echidna/Solidity.hs
@@ -26,8 +26,10 @@ import Data.Monoid                ((<>))
 import Data.Text                  (Text, isPrefixOf, isSuffixOf, append)
 import Data.Text.Lens             (unpacked)
 import Data.Text.Read             (decimal)
-import System.Process             (StdStream(..), readCreateProcess, proc, std_err)
+import System.Process             (StdStream(..), readCreateProcessWithExitCode, proc, std_err)
 import System.IO                  (openFile, IOMode(..))
+import System.Exit                (ExitCode(..))
+import System.Directory           (findExecutable)
 
 import Echidna.ABI         (SolSignature)
 import Echidna.Exec        (execTx)
@@ -58,6 +60,7 @@ data SolException = BadAddr Addr
                   | NoTests
                   | OnlyTests
                   | ConstructorArgs String
+                  | NoCryticCompile
 
 instance Show SolException where
   show = \case
@@ -71,6 +74,7 @@ instance Show SolException where
     NoTests              -> "No tests found in ABI"
     OnlyTests            -> "Only tests and no public functions found in ABI"
     (ConstructorArgs s)  -> "Constructor arguments are required: " ++ s
+    NoCryticCompile      -> "crytic-compile not installed. To install it, run:\n   pip install crytic-compile"
 
 
 instance Exception SolException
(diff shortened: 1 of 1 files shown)


lib/Echidna/Solidity.hs
```diff
@@ -106,8 +112,11 @@ contracts fp = let usual = ["--solc-disable-warnings", "--export-format", "solc"
                   (\sa -> if null sa then [] else ["--solc-args", sa])
   maybe (throwM CompileFailure) (pure . toList . fst) =<< liftIO (do
     stderr <- if q then UseHandle <$> openFile "/dev/null" WriteMode else pure Inherit
-    _ <- readCreateProcess (proc "crytic-compile" $ (c ++ solargs) |> fp) {std_err = stderr} ""
-    readSolc "crytic-export/combined_solc.json")
+    (ec, _, _) <- readCreateProcessWithExitCode (proc "crytic-compile" $ (c ++ solargs) |> fp) {std_err = stderr} ""
```
arcz: We can use the full path returned by `findExecutable`

lib/Echidna/Solidity.hs
```diff
@@ -71,6 +74,7 @@ instance Show SolException where
     NoTests              -> "No tests found in ABI"
     OnlyTests            -> "Only tests and no public functions found in ABI"
     (ConstructorArgs s)  -> "Constructor arguments are required: " ++ s
+    NoCryticCompile      -> "crytic-compile not installed. To install it, run:\n   pip install crytic-compile"
```
arcz: should we add that Echidna expects `crytic-compile` to be on PATH to be crystal clear?

lib/Echidna/Solidity.hs
incertia: I am in favor of this

lib/Echidna/Solidity.hs
incertia: Historically we've said that in the releases but I agree it would be good to also add it to the README

ggrieco-tob: `pip install crytic-compile` will add crytic-compile in your path (if you use `--user`, which is better IMHO, then you should add `~/.local/bin` into your `PATH`).

lib/Echidna/Solidity.hs
ggrieco-tob: We could use the path returned from the `findExecutable` call, however, if it finds path, then `readCreateProcessWithExitCode` should not fail (with code 127).

arcz: It still might fail due to a race

ggrieco-tob: Fair enough. This should be implemented already. Please take a look.

lib/Echidna/Solidity.hs
ggrieco-tob: Improved error message to ask for crytic-compile in PATH.

PR #336, opened 2020-01-06
Minor shrinking improvements

arcz:
This includes a few shrinking improvements:

1. `AbiInt` and `AbiUInt` shrinking is no longer stuck when there is a last bit on a high position
2. `addChars` function was occasionally adding one byte due to off by one error
3. `0` has a preference when shrinking `value`, `gasprice` and `delay`. Should solve https://github.com/crytic/echidna/issues/308#issuecomment-557922200

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -14,7 +14,6 @@ import Control.Monad.Catch (MonadThrow(..))
 import Control.Monad.State.Class (MonadState, gets)
 import Control.Monad.State (evalStateT)
 import Control.Monad.Random.Strict (MonadRandom, getRandom, getRandoms, getRandomR, uniformMay)
-import Data.Bits (Bits(..))
 import Data.Bool (bool)
 import Data.ByteString (ByteString)
 import Data.Foldable (toList)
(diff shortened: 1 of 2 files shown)


lib/Echidna/Transaction.hs
```diff
@@ -131,7 +131,9 @@ canShrinkTx _                                = True
 shrinkTx :: MonadRandom m => Tx -> m Tx
 shrinkTx tx'@(Tx c _ _ _ gp (C _ v) (C _ t, C _ b)) = let
   c' = either (fmap Left . shrinkAbiCall) (fmap Right . pure) c
+  lower 0 = pure $ w256 0
   lower x = w256 . fromIntegral <$> getRandomR (0 :: Integer, fromIntegral x)
+              >>= \r -> uniform [0, r] -- try 0 quicker
```
incertia: maybe we want to be explicit about `uniform` wherever we use `lower`. pinging @japesinator for this one.

lib/Echidna/ABI.hs
```diff
@@ -210,13 +209,8 @@ canShrinkAbiValue (AbiArrayDynamic _ l) = l /= mempty
...
 
-bounds :: forall a. (Bounded a, Integral a) => a -> (Integer, Integer)
-bounds = const (fromIntegral (0 :: a), fromIntegral (maxBound :: a))
-
--- | Given a bitvector representation of an integer type, randomly change bits, shrinking it towards 0
-shrinkInt :: forall a m. (Bits a, Bounded a, Integral a, MonadRandom m) => a -> m a
-shrinkInt x | x == -1   = pure 0
-            | otherwise = (if x < 0 then (.|.) else (.&.)) x . fromIntegral <$> getRandomR (bounds x)
+shrinkInt :: (Integral a, MonadRandom m) => a -> m a
+shrinkInt x = fromIntegral <$> getRandomR (0, toInteger x)
```
incertia: maybe we also want to keep the "bitwise shrinking". i kinda like the idea of being able to switch off bits.

arcz: I can bring it back if there is a good use case for it. Not sure if it is worth complicating the code though. The random smaller number turned out to be simpler and more effective strategy.

lib/Echidna/Transaction.hs
arcz: I've just put it here because all calls to lower should use this strategy, it is a local function.

PR #340, opened 2020-01-14
Split echidna core types into several smaller modules

incertia:
This lets us import types more easily and avoid circular dependencies


incertia:
this is being superseded by several smaller pull requests

PR #345, opened 2020-01-21
Decouple UI updates from fuzzing

arcz:
This PR introduces a few changes:
1. `campaign` function is now instrumented with a simple `IORef` which gets updated on each state step.
2. Ticking thread pushes UI updates at regular intervals of time (100ms), state is fetched from the `IORef`.
3. Without dashboard, main thread now fully sleeps until a campaign is done or timed out.
4. Simplified timeout detection, UI doesn't need to be reinitialized.

Motivation for those changes:
1. Decouple UI updates from fuzzing progress. This lays foundation for an interactive timer to be shown in the UI.
2. Fix the bug when UI goes away and reappears to show the timeout information.

### DIFF
--- a/lib/Echidna/UI.hs
+++ b/lib/Echidna/UI.hs
@@ -79,15 +79,16 @@ ui v w ts d = do
   let getSeed = view $ hasLens . to seed . non (d' ^. defSeed)
   bc <- liftIO $ newBChan 100
   ref <- liftIO $ newIORef defaultCampaign
-  let pushUpdate x = readIORef ref >>= writeBChan bc . x
+  let updateRef = use hasLens >>= liftIO . atomicWriteIORef ref
+  let updateUI x = readIORef ref >>= writeBChan bc . x
   waitForMe <- liftIO newEmptyMVar
   ticker <- liftIO $ forkIO -- update UI every 100ms
-    (void $ forever $ threadDelay 100000 >> pushUpdate CampaignUpdated)
+    (void $ forever $ threadDelay 100000 >> updateUI CampaignUpdated)
   timeoutSeconds <- (* 1000000) . fromMaybe (-1) <$> view (hasLens . maxTime)
   _ <- forkFinally -- run worker
-    (void $ timeout timeoutSeconds (campaign ref v w ts d) >>= \case
-       Nothing -> liftIO $ pushUpdate CampaignTimedout
-       Just _ -> liftIO $ pushUpdate CampaignUpdated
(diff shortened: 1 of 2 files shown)


lib/Echidna/UI.hs
```diff
@@ -148,22 +154,28 @@ ui :: ( MonadCatch m, MonadRandom m, MonadReader x m, MonadUnliftIO m
...
-ui v w ts d = let xfer bc = use hasLens >>= liftIO . writeBChan bc
-                  d' = fromMaybe defaultDict d
-                  getSeed = view $ hasLens . to seed . non (d' ^. defSeed) in do
+ui v w ts d = do
+  let d' = fromMaybe defaultDict d
+  let getSeed = view $ hasLens . to seed . non (d' ^. defSeed)
   bc <- liftIO $ newBChan 100
-  t    <- forkIO (void $ campaign (xfer bc) v w ts d)
+  ref <- liftIO $ newIORef defaultCampaign
+  let updateRef = use hasLens >>= liftIO . atomicWriteIORef ref
```
japesinator: Why are we tracking campaign state in both an IORef and a BChan? Can we simplify to one or the other?

lib/Echidna/UI.hs
```diff
@@ -148,22 +154,28 @@ ui :: ( MonadCatch m, MonadRandom m, MonadReader x m, MonadUnliftIO m
...
+  let updateUI x = readIORef ref >>= writeBChan bc . x
+  waitForMe <- liftIO newEmptyMVar
+  ticker <- liftIO $ forkIO -- update UI every 100ms
+    (void $ forever $ threadDelay 100000 >> updateUI CampaignUpdated)
+  timeoutSeconds <- (* 1000000) . fromMaybe (-1) <$> view (hasLens . maxTime)
+  _ <- forkFinally -- run worker
+    (void $ timeout timeoutSeconds (campaign updateRef v w ts d) >>= \case
+       Nothing -> liftIO $ updateUI CampaignTimedout
+       Just _ -> liftIO $ updateUI CampaignUpdated
```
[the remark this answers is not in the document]
arcz: this won't work for the call of `updateUI` above - `(void $ forever $ threadDelay 100000 >> updateUI CampaignUpdated)`

lib/Echidna/UI.hs
arcz: IORef is used to sample a value of campaign at a particular point in time. This value is then pushed as a UI update through a BChan. The state is rather tracked in IORef, BChan acts as an event queue for Brick UI. Using both is what allows for controlled UI updates / decoupling.

PR #347, opened 2020-01-27
Adjust testsuite parameters

incertia:
In an effort to make the testsuite run faster, we turn down `testLimit` by a factor of `5` and `shrinkLimit` by a factor of `2`. This should reduce wasted effort on shrinking unshrinkable results (the shrinker) is pretty efficient and continuous fuzzing of properties that are always `true`, as our test contracts are not that complex.

Merge this after #346.

### DIFF
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -10,7 +10,6 @@ module Echidna.Exec where
 import Control.Lens
 import Control.Monad.Catch (Exception, MonadThrow(..))
 import Control.Monad.State.Strict (MonadState, execState)
-import Data.Either (isRight)
 import Data.Has (Has(..))
 import Data.Map.Strict (Map)
 import Data.Maybe (fromMaybe)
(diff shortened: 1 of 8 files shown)


lib/Echidna/Exec.hs
```diff
@@ -70,12 +69,12 @@ execTxWith h m t = do (og :: VM) <- use hasLens
...
-                                                       hasLens . state . calldata .= cd
-                                                       hasLens . result ?= f
-                        (VMFailure x, _)         -> h x
-                        (VMSuccess bc, True)     -> (hasLens %=) . execState $ do
+                      case (res, t ^. call) of
+                        (f@Reversion, _)            -> do hasLens .= og
+                                                          hasLens . state . calldata .= cd
+                                                          hasLens . result ?= f
+                        (VMFailure x, _)            -> h x
+                        (VMSuccess bc, SolCreate _) -> (hasLens %=) . execState $ do
```
arcz: I know it's been like that before but is this case exhaustive?

incertia: the next case is a catch-all

arcz: thanks, I've missed that 👍

PR #348, opened 2020-01-28
Update addressarrayexample for blog post(s)

agroce:
This adds reverts_ properties and the mutants (we don't need them now, but will in the future)


arcz:
what about adding some README to `examples/solidity/addressarrayutils` explaining why this particular code is included and instructions to run?

agroce:
Good idea.  With link to blog post when that's ready, but something in the meantime

agroce:
Yeah, let's readme when we can link to blog + crytic.

PR #350, opened 2020-02-03
Fix indefinite MVar blocking

arcz:
We don't need the ticker thread when dashboard (UI) is disabled. It probably caused a silent exception causing dropping a reference to the `waitForMe` MVar which in turn resulted that main thread could not move forward and crashed with the error: `thread blocked indefinitely in an MVar operation`.

### DIFF
--- a/lib/Echidna/UI.hs
+++ b/lib/Echidna/UI.hs
@@ -80,12 +80,12 @@ ui v w ts d = do
   bc <- liftIO $ newBChan 100
   ref <- liftIO $ newIORef defaultCampaign
   dash <- liftM2 (&&) isTerminal $ view (hasLens . dashboard)
+  timeoutSeconds <- (* 1000000) . fromMaybe (-1) <$> view (hasLens . maxTime)
   let updateRef = use hasLens >>= liftIO . atomicWriteIORef ref
   let updateUI e = when dash $ readIORef ref >>= writeBChan bc . e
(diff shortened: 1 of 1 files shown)


lib/Echidna/UI.hs
```diff
@@ -79,19 +79,19 @@ ui v w ts d = do
...
+  dash <- liftM2 (&&) isTerminal $ view (hasLens . dashboard)
   let updateRef = use hasLens >>= liftIO . atomicWriteIORef ref
-  let updateUI x = readIORef ref >>= writeBChan bc . x
+  let updateUI e = when dash $ readIORef ref >>= writeBChan bc . e
   waitForMe <- liftIO newEmptyMVar
-  ticker <- liftIO $ forkIO -- update UI every 100ms
-    (void $ forever $ threadDelay 100000 >> updateUI CampaignUpdated)
+  ticker <- liftIO $ forkIO -- update UI every 100ms, instant exit when no UI
+    (when dash $ forever $ threadDelay 100000 >> updateUI CampaignUpdated)
   timeoutSeconds <- (* 1000000) . fromMaybe (-1) <$> view (hasLens . maxTime)
```
incertia: can we maybe move this line too

arcz: after dash?

incertia: yeah sure

PR #351, opened 2020-02-03
some compiler failure messages

incertia:
title

### DIFF
--- a/lib/Echidna/Solidity.hs
+++ b/lib/Echidna/Solidity.hs
@@ -54,7 +54,7 @@ import qualified Data.Text           as T
 -- | Things that can go wrong trying to load a Solidity file for Echidna testing. Read the 'Show'
 -- instance for more detailed explanations.
 data SolException = BadAddr Addr
-                  | CompileFailure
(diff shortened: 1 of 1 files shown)


lib/Echidna/Solidity.hs
```diff
@@ -120,12 +120,12 @@ contracts fp = let usual = ["--solc-disable-warnings", "--export-format", "solc"
                   (\sa -> if null sa then [] else ["--solc-args", sa])
         fps = toList fp
         compileOne :: (MonadIO m, MonadThrow m, MonadReader x m, Has SolConf x) => FilePath -> m [SolcContract]
-        compileOne x = maybe (throwM CompileFailure) (pure . toList . fst) =<< liftIO (do
+        compileOne x = maybe (throwM $ CompileFailure "" "could not read combined_solc.json") (pure . toList . fst) =<< liftIO (do
```
arcz: I think this might be a bit misleading since this is not the actual stderr. Maybe a new data constructor to match on for `SolException`?

Issue #286, opened 2019-08-20
Feature request: finding worst-case execution gas with echidna


montyly:
We could use echidna to provide a summary of the worst-case execution gas of each function (not just the properties).

That would help to find out-of-gas issues, and to profile which functions need to be optimized.

yxliang01:
@montyly Just a question (possibly out of scope), how can echidna a fuzzer find the worst-case execution gas? I believe this is going to be an estimation. Would you like to comment on how close such estimation you are thinking it's going to be? Thanks.

montyly:
Yes it's going to be an estimation and not a guarantee.

The precision will depend on the complexity of the contracts, and the capacity for echidna to explore the states, so it's hard to say. The goal here would not be to ensure a bound on the gas consumption but to find bad gas usage with a best approach effort, in a minimal amount of time.

We can start with a naive version which just fuzz randomly all the functions.Then we can probably tune the exploration with some heuristics (ex: by locating targets with Slither)

yxliang01:
Just curious, how are you planning to deal with paths containing external contract calling while the external contract is not accessible(e.g. its address can dynamically be set)?

montyly:
Well, I guess it's not a problem related to the gas estimation, but a general concern for echidna exploration?

But in that case, you can just fuzz all the addresses parameters with all known addresses (including all the contracts created)

ggrieco-tob:
Fixed in #355 (this needs some documentation, but I will do that soon).

PR #355, opened 2020-02-12
Gas Worst-case Estimation Feature (#286)

agroce:
Will address https://github.com/crytic/echidna/issues/286

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -178,7 +178,7 @@ updateGasInfo gi [] _ = gi
 updateGasInfo gi ((t@(Tx (SolCall((f, _))) _ _ _ _ _ _), (_, used')):ts) tseq =
   let mused = Data.Map.lookup f gi
   in  case mused of Nothing -> updateGasInfo (insert f (used', t:tseq) gi) ts (t:tseq)
-                    Just (used, _) | used' > used -> updateGasInfo (insert f (used', t:tseq) gi) ts (t:tseq)
+                    Just (used, _) | used' >= used -> updateGasInfo (insert f (used', t:tseq) gi) ts (t:tseq)
                     _ -> updateGasInfo gi ts (t:tseq)
 updateGasInfo gi ((t, _):ts) tseq = updateGasInfo gi ts (t:tseq)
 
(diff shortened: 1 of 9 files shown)


lib/Echidna/UI/Report.hs
```diff
@@ -47,6 +48,11 @@ ppCoverage s | s == mempty = Nothing
              | otherwise   = Just $ "Unique instructions: " ++ show (coveragePoints s)
                                  ++ "\nUnique codehashes: " ++ show (length s)
 
+-- | Pretty-print the gas usage information a 'Campaign' has obtained.
+ppGasInfo :: Map Text (Int, [Tx]) -> Maybe String
+ppGasInfo s | s == mempty = Nothing
+            | otherwise   = Just $ foldr (\t1 t2 -> t1 ++ "\n" ++ t2) "" $ map show $ toList s
```
arcz: you can also consider `intercalate "\n"` instead of `foldr`

agroce:
Current status is that this works but there are a few major outstanding issues, before it's ready for people.  Then we probably need to think about test generation, to hit the corner cases, since this is often nastier than just finding property violations.

- The call sequences need a shrinker.  They are very hard to read, unless very short and easy to find (the boring cases).  ABI offers some general shrinking code, so I think just adapting the test shrinker in Test.hs to this case will work fairly easily

- The JSON has the sequence, but for some reason I don't get, drops the actual gas values.  Figure out why and include those.

agroce:
- not sure about the function in SolCall being a good map key in multicontract...

agroce:
@ggrieco-tob this is ready for review, I think, assuming we're punting the shrinker for a second PR.  The only thing not checking out ok is the one codeclimate issue that looks to pre-exist the PR.

lib/Echidna/Campaign.hs
```diff
@@ -166,9 +174,38 @@ evalSeq v e = go [] where
     case xs of []     -> pure []
                (y:ys) -> e y >>= \a -> ((y, a) :) <$> go (y:r) ys
 
+-- | Given a call sequence that produces Tx with gas >= g for f, try to randomly generate
+-- a smaller one that achieves at least that gas usage
+shrinkGasSeq :: ( MonadRandom m, MonadReader x m, MonadThrow m
```
ggrieco-tob: This one is unused, right?

agroce: Yes, that's for the future gas-based shrinking

Issue #341, opened 2020-01-15
Feature request: filtering the functions to be called 


montyly:
Right now Echidna will call all the functions from the contract's ABI. It would be nice to be able to restrict the exploration, either through a whitelist or a blacklist.

For the API, I think we could use the configuration file, with two new fields: `functionsToBeCalledWhitelist`,  `functionsToBeCalledBlacklist` (or with better names).

incertia:
can we just make these a single JSON object which defaults to blacklist

```yaml
filter:
  blacklist: true
  functions: []
```

ggrieco-tob:
Fixed. I'm writing some examples/documentation for this feature.

PR #356, opened 2020-02-17
Filtering the functions to be called

ggrieco-tob:
This PR adds two new config options:

* `filterBlacklist` to set to blacklist or whitelist functions (`false` by default)
* `filterFunctions`: A list of strings with function names either to blacklist or whitelist, according to the previous config option. (`[]` by default).

This will fix issue #341.

### DIFF
--- a/examples/solidity/basic/default.yaml
+++ b/examples/solidity/basic/default.yaml
@@ -60,3 +60,5 @@ maxBlockDelay: 60480
 #maximum number of blocks elapsed between generated txs; default is expected increment in one week
 # timeout:
 #campaign timeout (in seconds)
+ignoreMethods: []
+includeMethods: []
--- a/examples/solidity/basic/default.yaml
+++ b/examples/solidity/basic/default.yaml
@@ -60,5 +60,7 @@ maxBlockDelay: 60480
 #maximum number of blocks elapsed between generated txs; default is expected increment in one week
 # timeout:
 #campaign timeout (in seconds)
-ignoreMethods: []
-includeMethods: []
+# list of methods to filter
+filterFunctions: []
+# by default, blacklist methods in filterFunctions
+filterBlacklist: true
--- a/examples/solidity/basic/default.yaml
+++ b/examples/solidity/basic/default.yaml
@@ -68,3 +68,4 @@ filterFunctions: []
 filterBlacklist: true
 #directory to save the corpus; by default is disabled  
 corpusDir: null
+
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -147,7 +147,9 @@ instance FromJSON EConfigWithUsage where
                                  <*> v ..:? "quiet"           ..!= False
                                  <*> v ..:? "initialize"      ..!= Nothing
                                  <*> v ..:? "multi-abi"       ..!= False
-                                 <*> v ..:? "checkAsserts"    ..!= False)
+                                 <*> v ..:? "checkAsserts"    ..!= False
+                                 <*> v ..:? "ignoreMethods"   ..!= []
+                                 <*> v ..:? "includeMethods"  ..!= [])
                     <*> tc
                     <*> xc
                     <*> (UIConf <$> v ..:? "dashboard" ..!= True <*> v ..:? "timeout" <*> style)
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -15,6 +15,7 @@ import Control.Monad.IO.Class (MonadIO(..))
 import Control.Monad.Reader (Reader, ReaderT(..), runReader)
 import Control.Monad.State (StateT(..), runStateT)
 import Control.Monad.Trans (lift)
+import Data.Bool (bool)
 import Data.ByteString.Lazy.Char8 (unpack)
 import Data.Aeson
 import Data.Aeson.Lens
@@ -148,8 +149,7 @@ instance FromJSON EConfigWithUsage where
                                  <*> v ..:? "initialize"      ..!= Nothing
                                  <*> v ..:? "multi-abi"       ..!= False
                                  <*> v ..:? "checkAsserts"    ..!= False
-                                 <*> v ..:? "ignoreMethods"   ..!= []
-                                 <*> v ..:? "includeMethods"  ..!= [])
+                                 <*> (bool Whitelist Blacklist <$> v ..:? "filterBlacklist" ..!= True <*> v ..:? "filterFunctions" ..!= []))
                     <*> tc
                     <*> xc
                     <*> (UIConf <$> v ..:? "dashboard" ..!= True <*> v ..:? "timeout" <*> style)
--- a/lib/Echidna/Solidity.hs
+++ b/lib/Echidna/Solidity.hs
@@ -163,10 +163,16 @@ linkLibraries [] = ""
 linkLibraries ls = "--libraries " ++
   iconcatMap (\i x -> concat [x, ":", show $ addrLibrary + toEnum i, ","]) ls
 
-filterMethods :: MonadThrow m => Filter -> [(Text, [AbiType])] -> m [(Text, [AbiType])]
+filterMethods :: MonadThrow m => Filter -> NE.NonEmpty SolSignature -> m (NE.NonEmpty SolSignature)
 filterMethods (Whitelist [])  _ = throwM InvalidMethodFilters 
-filterMethods (Whitelist ic) ms = return $ filter (\(m, _) -> T.unpack m `elem` ic) ms
-filterMethods (Blacklist ig) ms = return $ filter (\(m, _) -> T.unpack m `notElem` ig) ms
+filterMethods (Whitelist ic) ms = return $ NE.fromList $ NE.filter (\(m, _) -> T.unpack m `elem` ic) ms
+filterMethods (Blacklist ig) ms = return $ NE.fromList $ NE.filter (\(m, _) -> T.unpack m `notElem` ig) ms
+
+fallback :: SolSignature
+fallback = ("",[])
+
+abiOf :: Text -> SolcContract -> NE.NonEmpty SolSignature
+abiOf pref cc = NE.fromList $ fallback : filter (not . isPrefixOf pref . fst) (elems (cc ^. abiMap) <&> \m -> (m ^. methodName, m ^.. methodInputs . traverse . _2))
 
 -- | Given an optional contract name and a list of 'SolcContract's, try to load the specified
 -- contract, or, if not provided, the first contract in the list, into a 'VM' usable for Echidna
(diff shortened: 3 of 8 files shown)


lib/Echidna/Solidity.hs
```diff
@@ -98,6 +100,8 @@ data SolConf = SolConf { _contractAddr    :: Addr             -- ^ Contract addr
                        , _initialize      :: Maybe FilePath   -- ^ Initialize world with Etheno txns
                        , _multiAbi        :: Bool             -- ^ Whether or not to use the multi-abi mode
                        , _checkAsserts    :: Bool             -- ^ Test if we can cause assertions to fail
+                       , _ignoreMethods   :: [String]         -- ^ List of method to avoid calling during a campaign
```
arcz: What about having something like this here instead:
```haskell
data MethodFilter = Whitelist NonEmpty | Blacklist NonEmpty
_methodFilter :: Maybe MethodFilter
```
and parse the config to this type, having a bit stronger semantics on type level.

ggrieco-tob: If I understand correctly, this type should allow json files with either whitelist or blacklist keywords in a config file. If you include both of them, the JSON parsing in Haskell should fail? It seems to be this JSON library was not designed for this kind of semantic parsing, where some keywords are excluded if certain other keyword are present.

incertia:
wait i accidentally commented on the issue instead

can we just make these a single JSON object which defaults to blacklist

```yaml
filter:
  blacklist: true
  functions: []
```

examples/solidity/basic/default.yaml
```diff
@@ -60,3 +60,5 @@ maxBlockDelay: 60480
 #maximum number of blocks elapsed between generated txs; default is expected increment in one week
 # timeout:
 #campaign timeout (in seconds)
+ignoreMethods: []
+includeMethods: []
```
incertia: can we just make these a single JSON object which defaults to blacklist

```yaml
filter:
  blacklist: true
  functions: []
```

lib/Echidna/Config.hs
```diff
@@ -147,7 +147,9 @@ instance FromJSON EConfigWithUsage where
                                  <*> v ..:? "quiet"           ..!= False
                                  <*> v ..:? "initialize"      ..!= Nothing
                                  <*> v ..:? "multi-abi"       ..!= False
-                                 <*> v ..:? "checkAsserts"    ..!= False)
+                                 <*> v ..:? "checkAsserts"    ..!= False
+                                 <*> v ..:? "ignoreMethods"   ..!= []
+                                 <*> v ..:? "includeMethods"  ..!= [])
```
incertia: single json

lib/Echidna/Solidity.hs
incertia: single JSON object makes this super nice

lib/Echidna/Solidity.hs
```diff
@@ -158,6 +162,13 @@ linkLibraries [] = ""
 linkLibraries ls = "--libraries " ++
   iconcatMap (\i x -> concat [x, ":", show $ addrLibrary + toEnum i, ","]) ls
 
+filterMethods :: MonadThrow m => [String] -> [String] -> [(Text, [AbiType])] -> m [(Text, [AbiType])]
```
incertia: use artur's `MethodFilter` thing

ggrieco-tob:
@incertia @arcz  
Using this type:

```haskell
data Filter = Whitelist (NE.NonEmpty String) | Blacklist (NE.NonEmpty String)
```

and the automatic generation of JSON instances:

```haskell
$(deriveJSON defaultOptions ''Filter)
```

echidna asked me to format this option as:

```yaml
methodFilter:
  tag: "Whitelist"
  contents: ["f"]
```
Using a different tag than `"Whitelist"` or `"Blacklist"`, or an empty list results in an error. This approach is not very fancy (since it forces you to use `tag` and `contents`), but minimizes the amount of code to maintain)

examples/solidity/basic/default.yaml
incertia: scratch that, just do
```yaml
filterBlacklist: true
filterFunctions: []
```

incertia:
using [this](https://github.com/crytic/echidna/pull/356#discussion_r389004017) config format

```haskell
data MethodFilter = Blacklist [Text] | Whitelist [Text]
_methodFilter :: MethodFilter
```
-- ...
method <- bool Whitelist Blacklist <$> v ..:? "filterBlacklist" ..:! True
functions <- v ..:? "filterFunctions" ..:! []
-- method functions :: MethodFilter
-- (bool Whitelist Blacklist <$> v ..:? "filterBlacklist" ..:! True <*> v ..:? "filterFunctions" ..:! []) :: Applicative f => f MethodFilter
-- as well if we are constructing the result via `Applicative`
```

**Comment by ggrieco-tob** [comment]
@incertia Re-wrote this PR based on your suggestion. Please take a look, let me know it's what you expect so I can finish with the testing.

File: lib/Echidna/Solidity.hs [hunk 5]
```diff
@@ -158,6 +163,17 @@ linkLibraries [] = ""
 linkLibraries ls = "--libraries " ++
   iconcatMap (\i x -> concat [x, ":", show $ addrLibrary + toEnum i, ","]) ls
 
+filterMethods :: MonadThrow m => Filter -> NE.NonEmpty SolSignature -> m (NE.NonEmpty SolSignature)
+filterMethods (Whitelist [])  _ = throwM InvalidMethodFilters 
+filterMethods (Whitelist ic) ms = return $ NE.fromList $ NE.filter (\(m, _) -> T.unpack m `elem` ic) ms
+filterMethods (Blacklist ig) ms = return $ NE.fromList $ NE.filter (\(m, _) -> T.unpack m `notElem` ig) ms
```
incertia: is `NonEmpty` the correct behavior here? maybe we want to also throw `InvalidMethodFilters` if the resulting list is empty

lib/Echidna/Solidity.hs
```diff
@@ -158,6 +163,17 @@ linkLibraries [] = ""
...
+
+fallback :: SolSignature
+fallback = ("",[])
+
+abiOf :: Text -> SolcContract -> NE.NonEmpty SolSignature
+abiOf pref cc = NE.fromList $ fallback : filter (not . isPrefixOf pref . fst) (elems (cc ^. abiMap) <&> \m -> (m ^. methodName, m ^.. methodInputs . traverse . _2))
```
incertia: let's use `NE..|` and be total, i know `fallback :` ensures construction but still

lib/Echidna/Solidity.hs
```diff
@@ -175,17 +191,22 @@ loadSpecified name cs = do
...
-      bc = c ^. creationCode
+  let bc = c ^. creationCode
       abi = liftM2 (,) (view methodName) (fmap snd . view methodInputs) <$> toList (c ^. abiMap)
       con = view constructorInputs c
       (tests, funs) = partition (isPrefixOf pref . fst) abi
-
+      
+
+  -- Filter ABI according to the config options
+  fabiOfc <- filterMethods fs $ abiOf pref c
```
incertia: is this the correct place to do this? how does this affect contracts created afterwards and multi-abi interaction?

lib/Echidna/Solidity.hs
```diff
@@ -158,6 +163,17 @@ linkLibraries [] = ""
 linkLibraries ls = "--libraries " ++
   iconcatMap (\i x -> concat [x, ":", show $ addrLibrary + toEnum i, ","]) ls
 
+filterMethods :: MonadThrow m => Filter -> NE.NonEmpty SolSignature -> m (NE.NonEmpty SolSignature)
+filterMethods (Whitelist [])  _ = throwM InvalidMethodFilters 
+filterMethods (Whitelist ic) ms = return $ NE.fromList $ NE.filter (\(m, _) -> T.unpack m `elem` ic) ms
+filterMethods (Blacklist ig) ms = return $ NE.fromList $ NE.filter (\(m, _) -> T.unpack m `notElem` ig) ms
+
+fallback :: SolSignature
+fallback = ("",[])
```
incertia: maybe this should go in `Echidna.ABI`, where `SolSignature` is defined

lib/Echidna/Solidity.hs
```diff
@@ -232,7 +251,7 @@ loadWithCryticCompile fp name = contracts fp >>= loadSpecified name
 -- for running a 'Campaign' against the tests found.
 prepareForTest :: (MonadReader x m, Has SolConf x)
                => (VM, NE.NonEmpty SolSignature, [Text], M.HashMap BS.ByteString (NE.NonEmpty SolSignature)) -> m (VM, World, [SolTest])
-prepareForTest (v, a, ts, m) = view hasLens <&> \(SolConf _ _ s _ _ _ _ _ _ _ _ _ ch) ->
+prepareForTest (v, a, ts, m) = view hasLens <&> \(SolConf _ _ s _ _ _ _ _ _ _ _ _ ch _ ) ->
```
incertia: `NamedFieldPuns`

lib/Echidna/Solidity.hs
```diff
@@ -198,7 +219,6 @@ loadSpecified name cs = do
 
   -- Make sure everything is ready to use, then ship it
   when (null abi) $ throwM NoFuncs                              -- < ABI checks
-  neFuns <- maybe (throwM OnlyTests) pure (NE.nonEmpty funs)    -- <
```
incertia: this no longer triggers the `OnlyTests` condition

lib/Echidna/Solidity.hs
ggrieco-tob: Filtering methods is not supported in multi-abi mode. We can extend the support to that later, if necessary.

lib/Echidna/Solidity.hs
ggrieco-tob: Good idea. Done!

examples/solidity/basic/default.yaml
```diff
@@ -62,5 +62,9 @@ maxBlockDelay: 60480
 #maximum number of blocks elapsed between generated txs; default is expected increment in one week
 # timeout:
 #campaign timeout (in seconds)
+# list of methods to filter
+filterFunctions: []
+# by default, blacklist methods in filterFunctions
+filterBlacklist: true
 #directory to save the corpus; by default is disabled  
-corpusDir: null
+corpusDir: null
```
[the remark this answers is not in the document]
ggrieco-tob: Done.

lib/Echidna/Solidity.hs

lib/Echidna/Solidity.hs

lib/Echidna/Solidity.hs

lib/Echidna/Solidity.hs
ggrieco-tob: Fallback function will be always there, unless it is filtered. We discussed about handling this better in a future PR.

lib/Echidna/Solidity.hs
```diff
@@ -80,10 +81,13 @@ instance Show SolException where
     OnlyTests            -> "Only tests and no public functions found in ABI"
     (ConstructorArgs s)  -> "Constructor arguments are required: " ++ s
     NoCryticCompile      -> "crytic-compile not installed or not found in PATH. To install it, run:\n   pip install crytic-compile"
+    InvalidMethodFilters -> "Filtering methods whitelisting or blacklisting cannot result empty"
```
incertia: we can make this a bit more descriptive, see slack chat

ggrieco-tob: Done!

PR #305, opened 2019-09-20
Improved coverage detection using complete list of transactions

ggrieco-tob:
This PR include some major new features:
* Coverage detection using complete list of transactions: the complete list of transactions is used in the generation to find new test cases, every time new coverage is found. 
* New mutations modes for multiple transactions, including an optimized mode for contracts that should be tested using single-transaction. Using the list of the sequence of transactions that increased coverage, several generation and mutations are implemented:
  * Concatenation of sequences of transactions.
  * Mutation of one transaction of a sequence.
  * Mutation of all the transactions of a sequence.
  * Shirking of all the transactions of a sequence.
  * Insertion/Removal of some transactions in a sequence.
  * Swapping of transactions in a sequence.
* Improved support for saving and loading multi-transaction coverage. The new config keyword `corpusDir` allows to define an existing directory that will be used to save and load all the transactions that generated new coverage.

PR #357, opened 2020-02-18
Improved array generation using generic mutations

ggrieco-tob:
This PR implements better mutations for arrays (e.g. strings, dynamic arrays, etc). It will allow Echidna to test complex code more effectively, in particular when a contract parses something on-chain. This PR is needed in order to eventually merge #305.


lib/Echidna/Mutator.hs
```diff
@@ -0,0 +1,86 @@
...
+  if n < 0
+  then x : xs
+  else x : replaceAt a xs (n-1)
+
+
+expandRandList :: MonadRandom m => [a] -> m [a]
+expandRandList []   = return []
+expandRandList xs = do k <- getRandomR (0, length xs - 1)
+                       t <- getRandomR (1, max 32 (length xs))
+                       return $ insertAt (xs !! k) xs t
```
arcz: you can use `uniform xs` instead

lib/Echidna/Mutator.hs
```diff
@@ -0,0 +1,86 @@
...
+  return $ maybe xs (`take` (xs ++ fs)) mn
+
+mutateBS :: MonadRandom m => Maybe Int -> [Word8] -> ByteString -> m ByteString
+mutateBS mn fs bs = do
+  f <- listMutators
+  xs <- f $ unpack bs
+  return $ pack $ maybe xs (`take` (xs ++ fs)) mn
+
+
+replaceAt :: a -> [a] -> Int -> [a]
```
arcz: you can use lens to do that `(element index .~ value) list`

lib/Echidna/ABI.hs
```diff
@@ -241,20 +233,37 @@ shrinkAbiCall = traverse $ traverse shrinkAbiValue
...
 -- | Given an 'AbiValue', generate a random \"similar\" value of the same 'AbiType'.
 mutateAbiValue :: MonadRandom m => AbiValue -> m AbiValue
-mutateAbiValue (AbiUInt n x)         = AbiUInt n         <$> mutateNum x
-mutateAbiValue (AbiInt n x)          = AbiInt n          <$> mutateNum x
-mutateAbiValue (AbiAddress _)        = genAbiValue AbiAddressType
+mutateAbiValue (AbiUInt n x)         = getRandomR (0, 9 :: Int) >>=
+                                          \case  
+                                            0 -> (AbiUInt n <$> mutateNum x)
+                                            _ -> return $ AbiUInt n x
+mutateAbiValue (AbiInt n x)          = getRandomR (0, 9 :: Int) >>=
```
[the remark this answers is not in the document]
ggrieco-tob: Yes. Let me add a comment.

lib/Echidna/ABI.hs
```diff
@@ -241,20 +233,37 @@ shrinkAbiCall = traverse $ traverse shrinkAbiValue
...
+                                          xs <- mutateV (Just n) fs (V.toList l) 
+                                          return (AbiArray n t (V.fromList xs))
+
+mutateAbiValue (AbiArrayDynamic t l) = AbiArrayDynamic t . V.fromList <$> mutateV Nothing [] (V.toList l)
 mutateAbiValue (AbiTuple v)          = AbiTuple          <$> traverse mutateAbiValue v
 
 -- | Given a 'SolCall', generate a random \"similar\" call with the same 'SolSignature'.
-mutateAbiCall :: MonadRandom m => SolCall -> m SolCall
-mutateAbiCall = traverse $ traverse mutateAbiValue
+mutateAbiCall :: (MonadState x m, Has GenDict x, MonadRandom m) => SolCall -> m SolCall
```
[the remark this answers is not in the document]
ggrieco-tob: Good catch. Fixed.

lib/Echidna/Mutator.hs
```diff
@@ -0,0 +1,86 @@
...
+  else x : replaceAt a xs (n-1)
+
+
+expandRandList :: MonadRandom m => [a] -> m [a]
+expandRandList []   = return []
+expandRandList xs = do k <- getRandomR (0, length xs - 1)
+                       t <- getRandomR (1, max 32 (length xs))
+                       return $ insertAt (xs !! k) xs t
+
+expandAt :: [a] -> Int -> Int -> [a]
```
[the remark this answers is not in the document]
ggrieco-tob: `expandRandList` should use `expandAt`. Fixed.

lib/Echidna/Mutator.hs
ggrieco-tob: This was modified to requiere only using the index (`k`). I could use `uniform [0 .. length xs - 1]`, which is certainly more elegant that `getRandomR (0, length xs - 1)`, however, I believe using `getRandomR` will be faster at the cost of some code slightly less easier to read.

lib/Echidna/Mutator.hs
ggrieco-tob: > Might be good to actually use vector, there is a lot of random access going on.

The use of list makes easier to write mutation functions to do interesting things like expand or splice.

Regarding the use of lens, I'm not confident enough to re-write this code using that library. If someone wants try it, just commit in this branch.

incertia:
at merge point, can we squash all the `fixes` commits to a single commit?

lib/Echidna/Mutator.hs
```diff
@@ -0,0 +1,86 @@
...
+expandAt (x:xs) i n = case i of 
+                       0   -> replicate n x ++ xs 
+                       _   -> x: expandAt xs (i-1) n
+
+deleteRandList :: MonadRandom m => [a] -> m [a]
+deleteRandList []   = return []
+deleteRandList xs = do k <- getRandomR (0, length xs - 1)
+                       return $ deleteAt k xs
+
+deleteAt :: Int -> [a] -> [a]
```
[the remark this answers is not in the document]
incertia: `[a]` is not `at`-able. can we rewrite using `take` and `drop`?

lib/Echidna/Mutator.hs
```diff
@@ -0,0 +1,86 @@
...
+  return $ pack $ maybe xs (`take` (xs ++ fs)) mn
+
+
+replaceAt :: a -> [a] -> Int -> [a]
+replaceAt _ []     _ = []
+replaceAt a (_:xs) 0 = a:xs
+replaceAt a (x:xs) n =
+  if n < 0
+  then x : xs
+  else x : replaceAt a xs (n-1)
```
incertia: rewrite with `take` and `drop`

lib/Echidna/Mutator.hs
incertia: under the hood all the evm abi stuff is `Vector`. we should try to use `Vector` whenever possible if those arrays are being passed directly into these functions.

lib/Echidna/ABI.hs
```diff
@@ -241,20 +233,37 @@ shrinkAbiCall = traverse $ traverse shrinkAbiValue
...
-mutateAbiValue (AbiInt n x)          = AbiInt n          <$> mutateNum x
-mutateAbiValue (AbiAddress _)        = genAbiValue AbiAddressType
+mutateAbiValue (AbiUInt n x)         = getRandomR (0, 9 :: Int) >>= -- 10% of chance of mutation
+                                          \case  
+                                            0 -> (AbiUInt n <$> mutateNum x)
+                                            _ -> return $ AbiUInt n x
+mutateAbiValue (AbiInt n x)          = getRandomR (0, 9 :: Int) >>= -- 10% of chance of mutation
+                                          \case  
+                                            0 -> (AbiInt n <$> mutateNum x)
+                                            _ -> return $ AbiInt n x
```
incertia: can we make these 10 percents tunable config parameters. if you want to go the `Rational` route the JSON for that is `{ "numerator" : 1, "denominator" : 10 }`. otherwise choose something like `Double`.

lib/Echidna/Mutator.hs
```diff
@@ -0,0 +1,86 @@
+module Echidna.Mutator where
+
+import Control.Monad.Random.Strict (fromList, MonadRandom, getRandomR)
+import Data.Maybe (maybe)
+import Data.ByteString (ByteString, pack, unpack)
+import GHC.Word (Word8)
+import EVM.ABI (AbiValue(..))
+
+listMutators :: MonadRandom m => m ([a] -> m [a])
+listMutators = fromList [(return, 1), (expandRandList, 10), (deleteRandList, 10), (swapRandList, 10)]
```
incertia: can we config these weights as well?

lib/Echidna/Mutator.hs
```diff
@@ -0,0 +1,86 @@
...
+listMutators :: MonadRandom m => m ([a] -> m [a])
+listMutators = fromList [(return, 1), (expandRandList, 10), (deleteRandList, 10), (swapRandList, 10)] 
+ 
+mutateV :: MonadRandom m => Maybe Int -> [AbiValue] -> [AbiValue] -> m [AbiValue]
+mutateV mn fs vs = do
+  f <- listMutators
+  xs <- f vs
+  return $ maybe xs (`take` (xs ++ fs)) mn
+
+mutateBS :: MonadRandom m => Maybe Int -> [Word8] -> ByteString -> m ByteString
```
incertia: document

lib/Echidna/Mutator.hs
```diff
@@ -0,0 +1,86 @@
...
+import Control.Monad.Random.Strict (fromList, MonadRandom, getRandomR)
+import Data.Maybe (maybe)
+import Data.ByteString (ByteString, pack, unpack)
+import GHC.Word (Word8)
+import EVM.ABI (AbiValue(..))
+
+listMutators :: MonadRandom m => m ([a] -> m [a])
+listMutators = fromList [(return, 1), (expandRandList, 10), (deleteRandList, 10), (swapRandList, 10)] 
+ 
+mutateV :: MonadRandom m => Maybe Int -> [AbiValue] -> [AbiValue] -> m [AbiValue]
```
incertia: can we document the new parameters

lib/Echidna/ABI.hs
```diff
@@ -241,20 +233,37 @@ shrinkAbiCall = traverse $ traverse shrinkAbiValue
...
-mutateAbiValue (AbiArray n t l)      = AbiArray n t      <$> traverse mutateAbiValue l
-mutateAbiValue (AbiArrayDynamic t l) = AbiArrayDynamic t <$> mutateV t l
+mutateAbiValue (AbiBytes n b)        = do fs <- replicateM n getRandom
+                                          xs <- mutateBS (Just n) fs b 
+                                          return (AbiBytes n xs)
+
+mutateAbiValue (AbiBytesDynamic b)   = mutateBS Nothing [] b >>= (return . AbiBytesDynamic)
+mutateAbiValue (AbiString b)         = mutateBS Nothing [] b >>= (return . AbiString)
+mutateAbiValue (AbiArray n t l)      = do fs <- replicateM n $ genAbiValue t
+                                          xs <- mutateV (Just n) fs (V.toList l)
```
incertia: doesn't `Just n` accomplish the same as `Nothing`?

lib/Echidna/ABI.hs
```diff
@@ -285,7 +294,7 @@ genAbiValueM = genWithDict (fmap toList . view constants) $ \case
 
 -- | Given a 'SolSignature', generate a random 'SolCalls' with that signature, possibly with a dictionary.
 genAbiCallM :: (MonadState x m, Has GenDict x, MonadRandom m) => SolSignature -> m SolCall
-genAbiCallM = genWithDict (fmap toList . view wholeCalls) (traverse $ traverse genAbiValueM)
+genAbiCallM abi = genWithDict (fmap toList . view wholeCalls) (traverse $ traverse genAbiValueM) abi >>= mutateAbiCall
```
incertia: why are we doing a `mutateAbiCall` right after we `gen`?

ggrieco-tob:
@incertia @arcz I re-wrote the mutators using ListLike. Please take another look to this PR.

lib/Echidna/ABI.hs
ggrieco-tob: `Just n` means the mutated array/string should have size `n`. `Nothing` means the mutated array/string can have any size.

lib/Echidna/ABI.hs
ggrieco-tob: This is an interesting question. I couldn't find an easy way to mutate only the results of the values generated by the dictionaries, but mutating the generated values will perform the same effect. If you have a better way to do it, please let me know.

lib/Echidna/ABI.hs
ggrieco-tob: Increasing these values will make the tests to fail, because magic numbers are going to be (likely) corrupted. I think the user should be allowed to modify these.

lib/Echidna/Mutator.hs
ggrieco-tob: I need to experiment with these parameters (with the exception of the `return` one, which should be small), in order to understand which mutations are more useful. Probably @agroce will be interested on this as well.

PR #364, opened 2020-03-06
Properly handle contract self-destruction

ggrieco-tob:
When a contract self-destructs, Echidna is unable to test any properties. This means that all properties should fail. However, right now, Echidna is not handling this correctly, it [ignores](https://github.com/crytic/echidna/issues/26) the self-destruction in fact. 

This PR stops Echidna, showing that all the properties failed because of a self-destruction.


ggrieco-tob:
Superseded by #515

PR #367, opened 2020-03-07
Small coverage refactoring

ggrieco-tob:
This small refactoring prepares the code for future coverage improvements.  There is no new features here.


incertia:
we should fix the warning (`Data.Map.Map` not used) and then i can reapprove and merge

ggrieco-tob:
It should be fixed now.

Issue #363, opened 2020-03-04
Notion of coverage should properly handle reverting transactions


agroce:
Our current coverage counts instructions covered _during non-reverting transactions_.  In a sense, that's right -- it's not as useful to cover something if it isn't a potentially-blockchain-affecting execution of the code.  But it has a problem in that we lose information.  E.g., in constructs like:

```
if (foo) {
    revert();
}
```

it is impossible to tell if we ever had a case where `foo` held and we ran the `revert()` or not.  This is both not helpful for seeing how well echidna is doing, and makes test suites output in the corpus collection mode less interesting (covering all the reverts seems like a good idea).

Also, we won't keep txs that reverted but covered new code, to try to mutate them to non-reverting txs that cover new code.

agroce:
One solution is to make coverage of an instruction not a simple binary, but a tuple with info on whether it reverted.  Ideally, we'd cover everything both with a revert and without.  Some code can only be covered one value of "did-revert" of course.

ggrieco-tob:
I reviewed the code from the [current PR](https://github.com/crytic/echidna/pull/353) and it seems that coverage during reverts is **not** discarded (but it was on different branch, some time ago). To implement this new improvement, I suggest to use two different coverage maps (for non-reverting and reverting code coverage). After #353 is merged, I can implement this in a new PR.

ggrieco-tob:
In fact, preliminary research indicates that Solidity is not very well optimized and Echidna is already collecting different  instructions for examples like this one:

```
if (foo) {
    revert();
}
```

This is caused by Solidity generating a specif branch for the revert, something that Echidna will already detect.

However, if you run the following contract:

```
contract C {

  function f(bool sel, bool b) {
    if (sel)
      sel = false; // not important
    else
      sel = true;  // not important
    require(b);
  }

}
```

with this config:

```
coverage: true
corpusDir: "corpus"
seqLen: 1
testLimit: 100
cryticArgs: ["--solc", "solc-0.4.21"] 
checkAsserts: true
```

then echidna finds 3 different paths:
* A call to the fallback function.
* `f(false,false)` and `f(true,true)`

Ideally, we want to find:
* A call to the fallback function.
* `f(false,false)`, `f(true,false)`, `f(false,true)` and `f(true,true)`

Right?

agroce:
Seems right, since you can cover everything independent of everything else

ggrieco-tob:
This is already fixed.

PR #368, opened 2020-03-08
Multiple coverage tracking according to the result of a transaction

ggrieco-tob:
This is POC to fix issue #363 . This PR introduces two coverage maps to keep track of transactions when they revert (or not). This will allow a richer notion of coverage.


ggrieco-tob:
Not the best approach. Closing this one.

PR #369, opened 2020-03-09
Multiple coverage tracking according to the result of a transaction

ggrieco-tob:
This PR fixes issue #363 . It introduces a boolean type in the coverage maps to keep track of transactions when they revert (or not). This will allow a richer notion of coverage. 

It is worth mentioning that this PR will not improve the test cases generation by itself, until the code to mutate lists of transactions is merged (this will happen in upcoming PR, after #357 is merged).  

This PR also includes a simple test to check that the amount of lists of transactions in the corpus collected is correct (for a small contract). This tests the code that collect the corpus as well.


lib/Echidna/Campaign.hs
```diff
@@ -219,13 +219,18 @@ execTxOptC :: (MonadState x m, Has Campaign x, Has VM x, MonadThrow m) => Tx ->
 execTxOptC t = do
   og  <- hasLens . coverage <<.= mempty
   res <- execTxWith vmExcept (usingCoverage $ pointCoverage (hasLens . coverage)) t
-  hasLens . coverage %= unionWith union og
+  let vmr = isRevert $ fst res
```
incertia: we should have view patterns for reversions

lib/Echidna/Campaign.hs
```diff
@@ -219,13 +219,18 @@ execTxOptC :: (MonadState x m, Has Campaign x, Has VM x, MonadThrow m) => Tx ->
...
+  -- Update the coverage map with the proper binary according to the vm result 
+  hasLens . coverage %= mapWithKey (\ _ s -> DS.map (\(i,_) -> (i, vmr)) s)
```
incertia: i don't think constantly overwriting is the correct plan here. maybe `modifying _2 (vmr ||)` so revert sets it forever

lib/Echidna/Exec.hs
```diff
@@ -100,7 +100,7 @@ coveragePoints = sum . fmap S.size
 -- | Capture the current PC and codehash. This should identify instructions uniquely (maybe? EVM is weird).
 pointCoverage :: (MonadState x m, Has VM x) => Lens' x CoverageMap -> m ()
 pointCoverage l = use hasLens >>= \v ->
-  l %= M.insertWith (const . S.insert $ v ^. state . pc) (fromMaybe (W256 maxBound) $ h v) mempty where
+  l %= M.insertWith (const . S.insert $ (v ^. state . pc, True)) (fromMaybe (W256 maxBound) $ h v) mempty where
```
incertia: are we sure we want to set `True` by default? see the `or` comment.

lib/Echidna/Campaign.hs
ggrieco-tob: I believe this will be generalized soon with a different value for every type of revert.

ggrieco-tob: Done.

lib/Echidna/Exec.hs
ggrieco-tob: We are using `Success` by default but it can be any value. Later, it will be updated to the result extracted from the Transaction execution.

lib/Echidna/Exec.hs
```diff
@@ -97,10 +97,14 @@ usingCoverage cov = maybe (cov >> liftSH exec1 >> usingCoverage cov) pure =<< us
 coveragePoints :: CoverageMap -> Int
 coveragePoints = sum . fmap S.size
 
+-- | Given good point coverage, count simplified unique points (to report).
+scoveragePoints :: CoverageMap -> Int
+scoveragePoints = sum . fmap (S.size . S.map fst)
```
incertia: maybe we want to add a comment here about how it disregards `TxResults` as a set as it is a bit unclear

lib/Echidna/Transaction.hs
```diff
@@ -66,6 +66,47 @@ data Tx = Tx { _call  :: TxCall       -- | Call
...
+getResult (VMFailure (OutOfGas _ _))         = ErrorOutOfGas
+getResult (VMFailure (BadCheatCode _))       = ErrorBadCheatCode
+getResult (VMFailure StackLimitExceeded)     = ErrorStackLimitExceeded
+getResult (VMFailure IllegalOverflow)        = ErrorIllegalOverflow
+getResult (VMFailure (Query _))              = ErrorQuery
+getResult (VMFailure StateChangeWhileStatic) = ErrorStateChangeWhileStatic
+getResult (VMFailure InvalidMemoryAccess)    = ErrorInvalidMemoryAccess
+getResult (VMFailure CallDepthLimitReached)  = ErrorCallDepthLimitReached
+getResult (VMFailure (MaxCodeSizeExceeded _ _)) = ErrorMaxCodeSizeExceeded
+getResult (VMFailure PrecompileFailure)      = ErrorPrecompileFailure
```
incertia: maybe we can change the `Constructor _ _` pattern to `Constructor{}`

ggrieco-tob: We could, but I suspect we will need to get the information from some of this constructors in the near future to augment the `TxResult` type (and therefore, improving the coverage information). I suggest to open an issue to consider this change in the meantime.

lib/Echidna/Exec.hs
ggrieco-tob: Done.

PR #370, opened 2020-03-10
Basic mutations based on collected corpus

ggrieco-tob:
This small PR introduces the first mutations based on the collected corpora. The mutations are enabled using `coverage: true` and will randomly splice a sequence of transactions from the corpus with randomly generated ones.

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -235,15 +235,17 @@ addToCorpus res = unless (null rtxs) $ hasLens . corpus %= (rtxs:) where
 seqMutators :: (MonadRandom m) => m (Int -> [[Tx]] -> [Tx] -> m [Tx])
 seqMutators = fromList [(cnm, 1), (apm, 1), (prm, 1)]
   where -- Use the generated random transactions
-        cnm _  _         = return
+        cnm _ _          = return
         -- Append a sequence from the corpus with random ones
-        apm ql ctxs gtxs = do rtxs <- rElem $ NE.fromList ctxs
-                              k <- getRandomR (0, length rtxs - 1 )
-                              return $ take ql $ take k rtxs ++ gtxs
+        apm ql ctxs gtxs = do 
+          rtxs <- (rElem . NE.fromList) ctxs
+          k <- getRandomR (0, length rtxs - 1)
+          return . take ql . take k $ rtxs ++ gtxs
         -- Prepend a sequence from the corpus with random ones
-        prm ql ctxs gtxs = do rtxs <- rElem $ NE.fromList ctxs
-                              k <- getRandomR (0, length rtxs - 1 )
-                              return $ take ql $ take k gtxs ++ rtxs
+        prm ql ctxs gtxs = do 
+          rtxs <- (rElem . NE.fromList) ctxs
+          k <- getRandomR (0, length rtxs - 1)
+          return . take ql . take k $ gtxs ++ rtxs
 
 -- | Generate a new sequences of transactions, either using the corpus or with randomly created transactions
 randseq :: ( MonadCatch m, MonadRandom m, MonadReader x m, MonadState y m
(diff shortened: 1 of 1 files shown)


lib/Echidna/Campaign.hs
```diff
@@ -235,14 +236,30 @@ addToCorpus res = unless (null rtxs) $ hasLens . corpus %= (rtxs:) where
...
+        (0, _ ) -> return gtxs
+        -- Append a sequence from the corpus with random ones
+        (1, _ ) -> do rtxs <- rElem $ NE.fromList ctxs
+                      k <- getRandomR (0, length rtxs - 1 )
+                      return $ take ql $ take k rtxs ++ gtxs
+        -- Prepend a sequence from the corpus with random ones
+        (2, _ ) -> do rtxs <- rElem $ NE.fromList ctxs
+                      k <- getRandomR (0, length rtxs - 1 )
+                      return $ take ql $ take k gtxs ++ rtxs
+        _       -> error "invalid pattern in randseq"
```
incertia: we should properly define functions and then take a weighted random selection instead of hardcoding some integers.

lib/Echidna/Campaign.hs
```diff
@@ -235,14 +236,30 @@ addToCorpus res = unless (null rtxs) $ hasLens . corpus %= (rtxs:) where
...
-  let txs = ca ^. corpus  
-      p   = ca ^. ncallseqs 
-  if length txs > p then -- Replay the transactions in the corpus, if we are executing the first iterations
-    return $ txs !! p
-  else                   -- Randomly generate new transactions 
-    replicateM ql (evalStateT (genTxM o) (w, ca ^. genDict))
+  let ctxs = ca ^. corpus
+      p    = ca ^. ncallseqs
+  if length ctxs > p then -- Replay the transactions in the corpus, if we are executing the first iterations
+    return $ ctxs !! p
```
incertia: i think in the future we want to rethink this. we should preload the `GenDict` with values obtained from executing corpus transactions and populate the coverage dictionary accordingly.

ggrieco-tob: Yes, we could do that. But this simple approach guarantees that the corpus transactions are tested before anything else. In that case:
* The tests can fail before starting with the actual fuzzing campaign.
* The coverage will be as high as possible. This is useful to keep a "stable" corpus that will grow over time when new coverage is detected.

incertia: yeah i'm not saying change it now. but a future PR could improve this

lib/Echidna/Campaign.hs
incertia: also instead of `(0, _ ) ->` use `(0, _)  ->`

ggrieco-tob: Ok done. Can you re-check the code?

lib/Echidna/Campaign.hs
```diff
@@ -231,18 +232,40 @@ addToCorpus :: (MonadState s m, Has Campaign s) => [(Tx, (VMResult, Int))] -> m
 addToCorpus res = unless (null rtxs) $ hasLens . corpus %= (rtxs:) where
   rtxs = map fst res
 
+seqMutators :: (MonadRandom m) => m (Int -> [[Tx]] -> [Tx] -> m [Tx])
+seqMutators = fromList [(cnm, 1), (apm, 1), (prm, 1)]
+  where -- Use the generated random transactions
+        cnm _  _         = return
```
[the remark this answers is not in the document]
ggrieco-tob: Every function (`cnm`, `apm` and `prm`) have the same arguments, so the positions of the `_` mirrors them vertically. Anyany, I fixed this.

lib/Echidna/Campaign.hs
```diff
@@ -231,18 +232,40 @@ addToCorpus :: (MonadState s m, Has Campaign s) => [(Tx, (VMResult, Int))] -> m
...
+        -- Append a sequence from the corpus with random ones
+        apm ql ctxs gtxs = do rtxs <- rElem $ NE.fromList ctxs
+                              k <- getRandomR (0, length rtxs - 1 )
```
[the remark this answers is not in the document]
ggrieco-tob: Fixed.

incertia:
#371

PR #373, opened 2020-03-14
Allow to control corpus mutation constans using a configuration file

ggrieco-tob:
This small PR allows to control the constants used to select the most likely corpus mutators. The tuning of this parameter is for experimentation only (normal users should not need to modify it)

### DIFF
--- a/examples/solidity/basic/default.yaml
+++ b/examples/solidity/basic/default.yaml
@@ -68,4 +68,5 @@ filterFunctions: []
 filterBlacklist: true
 #directory to save the corpus; by default is disabled  
 corpusDir: null
-
+# constants for corpus mutations (for experimentation only)
+mutConsts: [1, 1, 1]
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -51,6 +51,8 @@ instance MonadThrow m => MonadThrow (RandT g m) where
 instance MonadCatch m => MonadCatch (RandT g m) where
   catch = liftCatch catch
 
+type MutationConsts = (Integer, Integer, Integer)
+
 -- | Configuration for running an Echidna 'Campaign'.
 data CampaignConf = CampaignConf { testLimit     :: Int
                                    -- ^ Maximum number of function calls to execute while fuzzing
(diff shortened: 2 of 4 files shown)


examples/solidity/basic/default.yaml
```diff
@@ -68,4 +68,5 @@ filterFunctions: []
 filterBlacklist: true
 #directory to save the corpus; by default is disabled  
 corpusDir: null
-
+# constants for corpus mutations (for experimentation only)
+mutConsts: [1, 1, 1]
```
incertia: let's make these constants more descriptive, e.g.
```yaml
weightMutateXXX: 1
weightMutateYYY: 1
weightMutateZZZ: 2
```

lib/Echidna/Campaign.hs
```diff
@@ -51,6 +51,8 @@ instance MonadThrow m => MonadThrow (RandT g m) where
 instance MonadCatch m => MonadCatch (RandT g m) where
   catch = liftCatch catch
 
+type MutationConsts = (Integer, Integer, Integer)
```
incertia: let's turn this into an actual lensible product type:

```haskell
data MutatorWeights = MutatorWeights { _mutatorXXX :: Integer, ... }
makeLenses ''MutatorWeights
```

src/test/Spec.hs
```diff
@@ -114,7 +114,7 @@ seedTests =
     , testCase "same seeds" $ assertBool "results differ" =<< same 0 0
     ]
     where cfg s = defaultConfig & sConf . quiet .~ True
-                                & cConf .~ CampaignConf 600 False False 20 0 Nothing (Just s) 0.15 Nothing
+                                & cConf .~ CampaignConf 600 False False 20 0 Nothing (Just s) 0.15 Nothing (1,1,1)
```
incertia: note to self: i'm definitely lensing `CampaignConf` so we don't have to deal with this bullshit

examples/solidity/basic/default.yaml
ggrieco-tob: This will grow in the near feature (check the other PRs with more mutations). I don't think that will be a good idea.

lib/Echidna/Campaign.hs
ggrieco-tob: Let's keep this a list, while we experiment which mutator is actually useful

PR #376, opened 2020-03-18
Prioritization to mutate each list of transactions from the corpus

ggrieco-tob:
This small PR implements a simple idea to select certain list of transactions from the collected corpus: the ones that are discovered later during the fuzzing campaign are assumed to be more valuable.


lib/Echidna/Campaign.hs
```diff
@@ -97,6 +98,8 @@ instance ToJSON TestState where
                                Solved  l -> (False, Just ("callseq", toJSON l))
                                Failed  e -> (False, Just ("exception", toJSON $ show e))
 
+type Corpus = DS.Set([Tx], Integer)
```
incertia: space after `Set`

PR #378, opened 2020-03-19
Added new research tests and reorganized examples folder

ggrieco-tob:


src/test/Spec.hs
```diff
@@ -190,9 +190,9 @@ integrationTests = testGroup "Solidity Integration Testing"
       [ ("echidna_balance failed",                 passed      "echidna_balance") ]
   , testContract "basic/library.sol"      (Just "basic/library.yaml")
       [ ("echidna_library_call failed",            solved      "echidna_library_call") ]
-  , testContract "harvey/foo.sol"         Nothing
+  , testContract "research/harvey_foo.sol" Nothing
       [ ("echidna_assert failed",                  solved      "echidna_assert") ]
-  , testContract "harvey/baz.sol"         Nothing
+  , testContract "research/harvey_baz.sol" Nothing
```
incertia: it might be worth it to put these in their own `testGroup`

ggrieco-tob: Done!

PR #383, opened 2020-03-25
Support for solc 0.6.x

ggrieco-tob:


incertia:
why must we disable the fallback test?

ggrieco-tob:
Syntax for fallback functions is not compatible between 0.5.x and 0.6.x

.travis.yml
```diff
@@ -50,8 +50,9 @@ script:
      stack --no-terminal test --ghc-options="$GHC_OPTIONS" $STACK_OPTS
    }
 
-   run_tests_using "solc-0.4.25"
+   #run_tests_using "solc-0.4.25"
```
ggrieco-tob: This PR looks good to me. The only concern is the time spent testing with 0.4.x, 0.5.x and 0.5.x. Should we keep support for the oldest version?

ggrieco-tob: Let's try to support it, if possible.

incertia:
ill investigate what happened with multicontract

ggrieco-tob:
@incertia I already merged with master, if you want to start digging into this.

ggrieco-tob:
Superseeded by #571

Issue #354, opened 2020-02-07
lensify everything


incertia:
i'm really tired of half our types not being lensed while the other half are. do this after #340 maybe?

ggrieco-tob:
This is no longer relevant

PR #384, opened 2020-03-25
refactor `CampaignConf` and `Campaign`

incertia:
this is the spiritual successor to #340 but it's happening over the course of multiple smaller commits. this also partly addresses #354. no new functionality was added so this should be pretty quick


ggrieco-tob:
This PR looks good to me, but let's try to merge some of the PRs we had in review first to minimize conflicts.

PR #385, opened 2020-03-26
Enable addresses shrinking

ggrieco-tob:
This small PR enables addresses shrinking with the following rules:
* The `0x0` address cannot be reduced. 
* The `0xdeadbeef` address can be reduced to `0x0`. 
* Any other address can be reduced to `0xdeadbeef` or `0x0`.


lib/Echidna/ABI.hs
```diff
@@ -232,7 +233,9 @@ shrinkInt x = fromIntegral <$> getRandomR (0, toInteger x)
 shrinkAbiValue :: MonadRandom m => AbiValue -> m AbiValue
 shrinkAbiValue (AbiUInt n m)         = AbiUInt n <$> shrinkInt m
 shrinkAbiValue (AbiInt n m)          = AbiInt n  <$> shrinkInt m
-shrinkAbiValue x@AbiAddress{}        = pure x
+shrinkAbiValue x@AbiAddress{}        = case x of
```
incertia, line 236: let's put the pattern match in `shrinkAbiValue` and not a separate case

ggrieco-tob: Done.

PR #393, opened 2020-04-01
Use safe mutation parameters to favor purely random generation

ggrieco-tob:


arcz:
what about same config in `examples/solidity/basic/default.yaml`? it is still `[1,1,1]`

ggrieco-tob:
Good catch, the tests are not failing if the default values do not match.

PR #394, opened 2020-04-01
fix dockerfile maybe

incertia:


reaperhulk:
You can shrink this docker image a bit more by adding `&& apt-get clean` to the end of that apt command to nuke the apt index and cache.

reaperhulk:
Also why is line 3 there? (That's not part of this diff, I'm just askin' questions)

PR #403, opened 2020-04-05
Corrected corpus prioritization usage and added comments

ggrieco-tob:


lib/Echidna/Campaign.hs
```diff
@@ -249,8 +248,10 @@ seqMutators (c1, c2, c3) = fromList
   where -- Use the generated random transactions
         cnm _ _          = return
         mut flp ql ctxs gtxs = do
-          let five_percent = 1 + (DS.size ctxs `div` 20)
-          rtxs <- fromList $ map (second fromInteger) $ take five_percent $ DS.toDescList ctxs
+          let some_percent = if (fst . DS.findMax) ctxs > 1   -- if the corpus already contains new elements
```
arcz, line 251: is there a reason why `some_percent` is snake case?

ggrieco-tob: Not really, let me fix that.

ggrieco-tob: Done, please check and re-approve if possible.

PR #410, opened 2020-04-07
Fixed coverage counting when a contract is deployed multiple times

ggrieco-tob:
This PR will fix how coverage is counted when multiple contracts are deployed more than once, to avoid counting several times.

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -31,6 +31,7 @@ import Data.Traversable (traverse)
 import EVM
 import EVM.ABI (getAbi, AbiType(AbiAddressType), AbiValue(AbiAddress))
 import EVM.Types (Addr)
+import EVM.Keccak (keccak)
 import Numeric (showHex)
 import System.Random (mkStdGen)
 
@@ -121,7 +122,7 @@ data Campaign = Campaign { _tests       :: [(SolTest, TestState)]
 instance ToJSON Campaign where
   toJSON (Campaign ts co gi _ _ _ _) = object $ ("tests", toJSON $ mapMaybe format ts)
     : ((if co == mempty then [] else [
-    ("coverage",) . toJSON . mapKeys (`showHex` "") $ DF.toList <$> co]) ++
+    ("coverage",) . toJSON . mapKeys (("0x" ++) . (`showHex` "") . keccak) $ DF.toList <$> co]) ++
        [(("maxgas",) . toJSON . toList) gi | gi /= mempty]) where
         format (Right _,      Open _) = Nothing
         format (Right (n, _), s)      = Just ("assertion in " <> n, toJSON s)
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -24,6 +24,7 @@ import qualified Data.Map as M
 import qualified Data.Set as S
 
 import Echidna.Transaction
+import Echidna.ABI (stripBytecodeMetadata)
 
 -- | Broad categories of execution failures: reversions, illegal operations, and ???.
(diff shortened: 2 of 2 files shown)


lib/Echidna/Exec.hs
```diff
@@ -106,8 +106,8 @@ scoveragePoints = sum . fmap (S.size . S.map fst)
 -- | Capture the current PC and codehash. This should identify instructions uniquely (maybe? EVM is weird).
```
arcz, line 106: Is this comment still relevant?

lib/Echidna/Exec.hs
```diff
@@ -106,8 +106,8 @@ scoveragePoints = sum . fmap (S.size . S.map fst)
...
 pointCoverage :: (MonadState x m, Has VM x) => Lens' x CoverageMap -> m ()
 pointCoverage l = use hasLens >>= \v ->
-  l %= M.insertWith (const . S.insert $ (v ^. state . pc, Success)) (fromMaybe (W256 maxBound) $ h v) mempty where
-    h v = v ^? env . contracts . at (v ^. state . contract) . _Just . codehash
+  l %= M.insertWith (const . S.insert $ (v ^. state . pc, Success)) (fromMaybe BS.empty $ h v) mempty where
```
arcz, line 109: I know it's been like that before, but I'm curious what is the scenario when `fromMaybe` defaults to `BS.empty`? Is this a valid state?

lib/Echidna/Campaign.hs
```diff
@@ -121,7 +122,7 @@ data Campaign = Campaign { _tests       :: [(SolTest, TestState)]
 instance ToJSON Campaign where
   toJSON (Campaign ts co gi _ _ _ _) = object $ ("tests", toJSON $ mapMaybe format ts)
     : ((if co == mempty then [] else [
-    ("coverage",) . toJSON . mapKeys (`showHex` "") $ DF.toList <$> co]) ++
+    ("coverage",) . toJSON . mapKeys (("0x" ++) . (`showHex` "") . keccak) $ DF.toList <$> co]) ++
```
arcz, line 125: Is this JSON deserialized somewhere?

lib/Echidna/Exec.hs
ggrieco-tob: I believe `fromMaybe` will never use it.

lib/Echidna/Campaign.hs
ggrieco-tob: What do you mean exactly? Where this JSON is used after it is saved?

lib/Echidna/Exec.hs
ggrieco-tob: Good catch, let me update it.

lib/Echidna/Exec.hs
arcz: It might be good to crash the program instead if that's the "impossible" case

lib/Echidna/Campaign.hs
arcz: Yes

lib/Echidna/Exec.hs
ggrieco-tob: Done.

lib/Echidna/Exec.hs
ggrieco-tob: Done.

lib/Echidna/Campaign.hs
ggrieco-tob: I don't know, perhaps @agroce is using it to capture Echidna usage in his experiments.

arcz: got it, I guess we can proceed since this is undocumented interface, whoever is using that needs to adjust though

PR #419, opened 2020-04-10
Develop and build with Nix

arcz:
This PR introduces Nix build for Echidna along with reproducible one command development environment setup - `nix-shell`. I've been using and testing this setup for months. Build has been successfully tested on macOS, Ubuntu and NixOS.

This doesn't override the current Docker/stack based workflow, just provides a non-conflicting alternative.

Future work will include a `solc-select` equivalent and binary cache to cut the setup time to the minimum.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,11 +87,6 @@ for example
 $ docker run -it -v `pwd`:/src echidna echidna-test /src/examples/solidity/basic/flags.sol
 ```
 
-Nix users can install the lastest Echidna with:
-```
-nix-env -i -f https://github.com/crytic/echidna/tarball/master
-```
-
 If you'd prefer to build from source, use [Stack](https://docs.haskellstack.org/en/stable/README/).
 `stack install` should build and compile `echidna-test` in `~/.local/bin`.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -87,6 +87,11 @@ for example
 $ docker run -it -v `pwd`:/src echidna echidna-test /src/examples/solidity/basic/flags.sol
 ```
 
+Nix users can install the lastest Echidna with:
```
ggrieco-tob, line 90: I think we should list Nix as the last option (maybe in its own subsection), to avoid confusing new users (they should use docker or the precompiled version)

arcz: sure, I can move it below stack

default.nix
```diff
@@ -0,0 +1,71 @@
+{ nixpkgs ? import ./nix/dapptools.nix, compiler ? "default", doBenchmark ? false }:
+
+let
+  inherit (nixpkgs) pkgs;
+
+  v = "1.4.0.1";
```
ggrieco-tob, line 6: Should we use the latest release here?

Issue #322, opened 2019-11-13
FairWin memory leak


ggrieco-tob:
A contract based on the [FairWin](https://etherscan.io/address/0x5Ec8515d15C758472f3E1A7B9eCa3e996E8Ba902) contract causes echidna to consume a large amount of memory (at least 10GB in less than a minute and keeps growing).  To reproduce, just run `echidna-test` with this contract:

```solidity
/**
 *Submitted for verification at Etherscan.io on 2019-07-29
*/

//pragma solidity ^0.4.24;

contract UtilFairWin  {
    uint ethWei = 1 ether;
    //getlevel
    function getlevel(uint value) public view returns(uint){
        if(value>=1*ethWei && value<=5*ethWei){
            return 1;
        }if(value>=6*ethWei && value<=10*ethWei){
            return 2;
        }if(value>=11*ethWei && value<=15*ethWei){
            return 3;
        }
            return 0;
    }
    //getLinelevel
    function getLineLevel(uint value) public view returns(uint){
        if(value>=1*ethWei && value<=5*ethWei){
            return 1;
        }if(value>=6*ethWei && value<=10*ethWei){
            return 2;
        }if(value>=11*ethWei){
            return 3;
        }
    }
    
    //level-commend  /1000
    function getScBylevel(uint level) public view returns(uint){
        if(level == 1){ 
            return 5;
        }if(level == 2){
            return 7;
        }if(level == 3) {
            return 10;
        }
        return 0;
    }
    
    //level fire scale   /10
    function getFireScBylevel(uint level) public view returns(uint){
        if(level == 1){
            return 3;
        }if(level == 2){
            return 6;
        }if(level == 3) {
            return 10;
        }return 0;
    }
    
    //level and times => invite scale /100
    function getRecommendScaleBylevelandTim(uint level,uint times) public view returns(uint){
        if(level == 1 && times == 1){ 
            return 50;
        }if(level == 2 && times == 1){
            return 70;
        }if(level == 2 && times == 2){
            return 50;
        }if(level == 3) {
            if(times == 1){
                return 100;
            }if(times == 2){
                return 70;
            }if(times == 3){
                return 50;
            }if(times >= 4 && times <= 10){
                return 10;
            }if(times >= 11 && times <= 20){
                return 5;
            }if(times >= 21){
                return 1;
            }
        } return 0;
    }
    
    function compareStr (string _str,string str) public view returns(bool) {
         bool checkResult = false;
[...]

PR #423, opened 2020-04-13
liftSH hack to solve memory issue

erivas:
Commit related to issue #322 changing liftSH.

### DIFF
--- a/lib/Echidna/Transaction.hs
+++ b/lib/Echidna/Transaction.hs
@@ -254,11 +254,11 @@ shrinkTx tx'@(Tx c _ _ _ gp (C _ v) (C _ t, C _ b)) = let
 liftSH :: (MonadState a m, Has b a) => State b x -> m x
 liftSH = stateST . runState . zoom hasLens
   -- This is the default state function written in terms of get and set:
-  where stateST :: MonadState s m => (s -> (a, s)) -> m a
-        stateST f = do s <- get
-                       let ~(a, s') = f s
-                       put s'
-                       return a
+  where stateST f = do
+          s <- get
+          let ~(a, s') = f s
+          put s'
+          return a
 
 -- | Given a 'Transaction', set up some 'VM' so it can be executed. Effectively, this just brings
 -- 'Transaction's \"on-chain\".
--- a/lib/Echidna/Transaction.hs
+++ b/lib/Echidna/Transaction.hs
@@ -16,7 +16,7 @@ import Control.Monad (join, liftM2, liftM3, liftM5, unless)
 import Control.Monad.Catch (MonadThrow, bracket)
 import Control.Monad.Random.Strict (MonadRandom, getRandomR, uniform)
 import Control.Monad.Reader.Class (MonadReader)
-import Control.Monad.State.Strict (MonadState, State, evalStateT, runState)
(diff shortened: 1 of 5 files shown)


lib/Echidna/Transaction.hs
```diff
@@ -253,7 +252,13 @@ shrinkTx tx'@(Tx c _ _ _ gp (C _ v) (C _ t, C _ b)) = let
 -- | Lift an action in the context of a component of some 'MonadState' to an action in the
 -- 'MonadState' itself.
 liftSH :: (MonadState a m, Has b a) => State b x -> m x
-liftSH = S.state . runState . zoom hasLens
+liftSH = stateST . runState . zoom hasLens
+  -- This is the default state function written in terms of get and set:
+  where stateST :: MonadState s m => (s -> (a, s)) -> m a
+        stateST f = do s <- get
```
incertia, line 258: put this on a newline and indent 2 spaces

incertia:
is there any reason why copying the mtl `state` function fixes this? is it due to not sufficient inlining or something?

lib/Echidna/Transaction.hs
erivas: Sorry, I don't get it. Should the `where` be in the same line as the `liftSH` and the `stateST` start in a newline with indent 2?
```
liftSH = stateST . runState . zoom hasLens where
  -- This is the default state function written in terms of get and set:
  stateST :: MonadState s m => (s -> (a, s)) -> m a
  stateST f = do s <- get
```

erivas:
> is there any reason why copying the mtl `state` function fixes this? is it due to not sufficient inlining or something?

My guess was that the particular instance of `m` used is overriding the default implementation of `state` (which is part of the class `MonadState`)?

ggrieco-tob:
According to [your research](https://github.com/crytic/echidna/issues/322#issuecomment-612932417), this PR also needs to limit the amount of gas per transaction to be effective. Please limit the amount of gas to the block limit (`8000030`) in the following lines:
* The gas used during the contract deployment ([here](https://github.com/crytic/echidna/blob/master/lib/Echidna/Solidity.hs#L159) and [here](https://github.com/crytic/echidna/blob/master/lib/Echidna/Solidity.hs#L231)).
* The gas used during the transaction generation ([here](https://github.com/crytic/echidna/blob/master/lib/Echidna/Config.hs#L110)).

ggrieco-tob:
It seems that one of the @agroce gas tests is failing, perhaps he can join the discussion here..

ggrieco-tob:
I can confirm this is a good workaround to keep control of the memory used. Thanks a lot @erivas !

agroce:
Not sure I follow what's going on with my code here?

ggrieco-tob:
@erivas had to add [this](https://github.com/crytic/echidna/pull/423/files#diff-09634d206772efe68f447bd8802bdd10R2) in order to fix [the failing gas usage test](https://github.com/crytic/echidna/runs/585706667#step:13:106).

lib/Echidna/Transaction.hs
incertia:
```haskell
where stateST f = do
  s <- get
  -- ...
```

PR #437, opened 2020-04-17
refactor `Tx`

incertia:
also introducing some more micromodules

* `Echidna.Types.Signature` for all sig-related types
* `Echidna.Orphans.XXX` for any of our orphan instances
* `Echidna.Pretty` for `ppXXX`


lib/Echidna/Transaction.hs
```diff
@@ -1,9 +1,7 @@
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE ViewPatterns #-}
```
ggrieco-tob, line 7: Perhaps this module should be called Echidna.Transaction.Gen ?

incertia: we can move it to gen in another PR i think. i want to mostly keep any internal modules in-place until we start finalizing the refactor

incertia: it would probably just be `Echidna.Gen` rather than `Echidna.Transaction.Gen` but yeah

lib/Echidna/Campaign.hs
```diff
@@ -42,6 +42,8 @@ import Echidna.Solidity
 import Echidna.Test
 import Echidna.Transaction
 import Echidna.Types.Campaign
+import Echidna.Types.Tx (TxCall(SolCall), Tx(Tx), TxConf, getResult, src, call, _SolCall)
```
ggrieco-tob, line 45: I think it is better to import all the constructors here..

incertia: anything other than this? i can totally push fixes for this soon tm

ggrieco-tob: The rest looks fine to me.

PR #443, opened 2020-04-22
Create new accounts to ensure they return zero balance

### LINKED ISSUE #442: inspecting the balance of fresh addresses fails
It appears that echidna has trouble checking the balance of a fresh address. With format: "text" and the following code, I receive "echidna_freshbalance: failed with no transactions made ⁉️".

```solidity
pragma solidity ^0.5.0;

contract FreshBalance {
  address payable public constant owner = address(0xcafe);

  function echidna_freshbalance() public returns (bool) {
    uint a = owner.balance;
    return true;
  }
}
```

If I give some balance to the address beforehand (eg., with sender: ["0xcafe"] in the config, even if balanceAddr == 0), fuzzing works as expected.

ggrieco-tob:
This PR fixes #442

### DIFF
--- a/examples/solidity/basic/balance.sol
+++ b/examples/solidity/basic/balance.sol
@@ -9,4 +9,8 @@ contract C {
       return address(msg.sender) == address(0x42) && msg.sender.balance == (42+123) && address(this).balance == 0;
   }
 
+  function echidna_balance_new() public returns (bool) { 
+      return address(0xcafe).balance == 0;
+  }
+
 }
--- a/examples/solidity/basic/balance.sol
+++ b/examples/solidity/basic/balance.sol
@@ -1,4 +1,6 @@
 contract C {
+  bool value_found = false;
+
   constructor() payable public {
       msg.sender.transfer(0);
       msg.sender.transfer(123);
@@ -13,4 +15,12 @@ contract C {
       return address(0xcafe).balance == 0;
   }
 
+  function echidna_no_magic() public returns (bool) {
+    if (value_found)
+      return false;
+
+    value_found = true;
+    uint x = address(0x123).balance;
+    return true;
+  }
 }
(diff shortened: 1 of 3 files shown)


incertia:
i'm not entirely sure this does what we want it to do

i am almost certain that hard failing for any contract we don't know about is a good thing, and would cause errors elsewhere. say we wanted to fuzz and see if we can set an address of an internal contract to something that doesn't exist. this would create that contract and i'm pretty sure running an empty contract doesn't do what we want in this case.

ggrieco-tob:
> i'm not entirely sure this does what we want it to do
> 
> i am almost certain that hard failing for any contract we don't know about is a good thing

Currently, we are not "hard failing" for any contract we don't know. Echidna just reverts.

> , and would cause errors elsewhere. say we wanted to fuzz and see if we can set an address of an internal contract to something that doesn't exist. this would create that contract and i'm pretty sure running an empty contract doesn't do what we want in this case.

This is not a crazy assumption about unknown contracts, this is to avoid reverting when executing the `balance` opcode, something it is certainly not what the yellow paper (or other Ethereum clients) are doing.

incertia:
btw, since it returns a `PleaseFetchContract` is there any way we can turn this into an RPC for an ethereum node and fetch the contract from there? that would be pretty cool

ggrieco-tob:
> btw, since it returns a `PleaseFetchContract` is there any way we can turn this into an RPC for an ethereum node and fetch the contract from there? that would be pretty cool

We could, but we need to refactor a few things to be able to add IO inside. However, we need a proper fix first. This is not correct, since this property will magically fail:

```solidity
contract C {
  bool value_found = false;

  function f() public { }

  function echidna_magic() public returns (bool) {
    if (value_found)
      return false;

    value_found = true;
    uint x = address(0xcafe).balance;
    return true;
  }
}
```

hevm provides a monadic continuation in the  `PleaseFetchContract`, but I'm not sure if we can use it..

ggrieco-tob:
While I'm not completely satisfied with the code, this looks like it is ready for review. Let's discuss some  alternatives for the formating.

ggrieco-tob:
Replaced by #657. As @montyly notes, this should be implemented by default to ensure low-level calls work as expected and cause an unexpected revert in the caller.

Issue #444, opened 2020-04-23
Upload our prebuilt docker to dockerhub 


ggrieco-tob:

ggrieco-tob:
Done.

PR #445, opened 2020-04-23
Uploads the Docker image to both GitHub packages as well as DockerHub

ESultanik:
Resolves Issue #444.

I am not sure how to test this other than just yeeting it to `master` and seeing if it works.


incertia:
name the branch [github-docker](https://github.com/crytic/echidna/blob/b1368743d8f04eff6efa1503455a5d65bd943c99/.github/workflows/docker.yml#L7) and repush

it tags as testing:testing though

PR #448, opened 2020-04-27
Simplify code without semantic changes

arcz:
This PR simplifies some of the most syntactically complex places in our code.

### DIFF
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -66,23 +66,26 @@ vmExcept e = throwM $ case VMFailure e of {Illegal -> IllegalExec e; _ -> Unknow
 -- | Given an error handler, an execution function, and a transaction, execute that transaction
 -- using the given execution strategy, handling errors with the given handler.
 execTxWith :: (MonadState x m, Has VM x) => (Error -> m ()) -> m VMResult -> Tx -> m (VMResult, Int)
-execTxWith h m t = do (og :: VM) <- use hasLens
-                      setupTx t
-                      gasIn <- use $ hasLens . state . gas
-                      res <- m
-                      gasOut <- use $ hasLens . state . gas
-                      cd  <- use $ hasLens . state . calldata
-                      case (res, t ^. call) of
-                        (f@Reversion, _)            -> do hasLens .= og
-                                                          hasLens . state . calldata .= cd
-                                                          hasLens . result ?= f
-                        (VMFailure x, _)            -> h x
-                        (VMSuccess bc, SolCreate _) -> (hasLens %=) . execState $ do
-                          env . contracts . at (t ^. dst) . _Just . contractcode .= InitCode ""
-                          replaceCodeOfSelf (RuntimeCode bc)
-                          loadContract (t ^. dst)
-                        _                        -> pure ()
-                      return (res, fromIntegral (gasIn - gasOut))
+execTxWith h m t = do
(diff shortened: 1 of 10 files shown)


lib/Echidna/Exec.hs
```diff
@@ -66,23 +66,26 @@ vmExcept e = throwM $ case VMFailure e of {Illegal -> IllegalExec e; _ -> Unknow
...
+  gasOut <- use $ hasLens . state . gas
+  cd  <- use $ hasLens . state . calldata
+  case (res, t ^. call) of
+    (f@Reversion, _) -> do
+      hasLens .= og
+      hasLens . state . calldata .= cd
+      hasLens . result ?= f
+    (VMFailure x, _) -> h x
+    (VMSuccess bc, SolCreate _) ->
+      (hasLens %=) . execState $ do
```
ggrieco-tob, line 83: Should we remove the use of `%=` here?

lib/Echidna/Solidity.hs
```diff
@@ -258,8 +261,10 @@ loadWithCryticCompile fp name = contracts fp >>= loadSpecified name
 -- for running a 'Campaign' against the tests found.
 prepareForTest :: (MonadReader x m, Has SolConf x)
                => (VM, NE.NonEmpty SolSignature, [Text], M.HashMap BS.ByteString (NE.NonEmpty SolSignature)) -> m (VM, World, [SolTest])
-prepareForTest (v, a, ts, m) = view hasLens <&> \SolConf { _sender = s, _checkAsserts = ch } ->
-  (v, World s m, fmap Left (zip ts $ repeat r) ++ if ch then Right <$> drop 1 a' else []) where
+prepareForTest (v, a, ts, m) = do
+  SolConf{_sender = s, _checkAsserts = ch} <- view hasLens
```
ggrieco-tob, line 265: spaces here before `_sender` and after `ch`

lib/Echidna/Exec.hs
```diff
@@ -66,23 +66,26 @@ vmExcept e = throwM $ case VMFailure e of {Illegal -> IllegalExec e; _ -> Unknow
...
-                                                          hasLens . result ?= f
-                        (VMFailure x, _)            -> h x
-                        (VMSuccess bc, SolCreate _) -> (hasLens %=) . execState $ do
-                          env . contracts . at (t ^. dst) . _Just . contractcode .= InitCode ""
-                          replaceCodeOfSelf (RuntimeCode bc)
-                          loadContract (t ^. dst)
-                        _                        -> pure ()
-                      return (res, fromIntegral (gasIn - gasOut))
+execTxWith h m t = do
+  (og :: VM) <- use hasLens
```
ggrieco-tob, line 70: I always found confusing the `og` name, where it was easier to use `vm` or something like that.

ggrieco-tob:
This PR looks good!, my only request is include more comments when possible, in particular in the `do` statements that you modified (since this way to code makes easier to add them).

lib/Echidna/Exec.hs
arcz: good point, I will change this to `vm`

lib/Echidna/Exec.hs
arcz: I think we can leave it for now as is. I will be looking into more simplifications after we merge this PR.

ggrieco-tob: Fair enough.

Issue #453, opened 2020-04-30
extcodesize returns wrong value on contracts being constructed


elopez:
extcodesize on a contract being constructed should return 0, but on echidna it returns non-0. Here's a testcase:

```solidity
pragma solidity ^0.5.0;

contract C {

  uint sz;

  constructor() public {
    uint x;
    address t = address(this);
    assembly { x := extcodesize(t) }
    sz = x;
  }

  function echidna_test_zero() public returns (bool) {
    return sz == 0;
  }
}
```

incertia:
i believe this is more of an hevm regression than an echidna issue

in particular, we have to run constructors as a `RuntimeCode "bytecode here"` rather than an `InitCode "bytecode here"` and hevm likely implements `extcodesize` for `RuntimeCode` as a non-constructor.

incertia:
jk

i *think* we can get around this by manually setting `state . code`. in particular,

```haskell
vm & env . contracts . at addr ?= initialContract (InitCode bc)
loadContract addr
vm & state . code .= bc
```

incertia:
check out the PR, i am almost certain it is the right thing to do

[extcodesize](https://github.com/dapphub/dapptools/blob/master/src/hevm/src/EVM.hs#L720)

ggrieco-tob:
Fixed in #454 and #465

PR #454, opened 2020-04-30
create contracts correctly

incertia:
addresses #453


ggrieco-tob:
If this works as expected, we should add something like @elopez test into the Spec.hs

incertia:
i am debating whether or not i want to PR `loadContract` upstream instead to work better.

ggrieco-tob:
Think we can have this patch here until your PR in hevm is accepted (and a new version is released)

incertia:
see dapphub/dapptools#364 for the upstream fix

PR #455, opened 2020-04-30
Set reasonable initial block time and number

ggrieco-tob:
This small PR sets initial timestamp and block number to reasonable values following Manticore code:

https://github.com/trailofbits/manticore/blob/8a77ad468082ed905449459e769b95839db3b54f/manticore/platforms/evm.py#L2283-L2291

### DIFF
--- a/lib/Echidna/Types/Tx.hs
+++ b/lib/Echidna/Types/Tx.hs
@@ -24,9 +24,10 @@ makePrisms ''TxCall
 $(deriveJSON defaultOptions ''TxCall)
 
 initialTimestamp :: Word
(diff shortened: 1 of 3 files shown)


lib/Echidna/Types/Tx.hs
```diff
@@ -23,6 +23,12 @@ data TxCall = SolCreate   ByteString
 makePrisms ''TxCall
 $(deriveJSON defaultOptions ''TxCall)
 
+initialTimestamp :: Word
```
incertia, line 26: i think it is better to eventually make these configurable but we'll leave it as a constant for now

PR #457, opened 2020-05-01
split each test group into separate files

incertia:
review should be quick and easy, the only other real change i did was add prisms to `SolException`. it shouldn't affect any other PRs other than #452.


incertia:
this is the first step to the testing refactor, the next step will be to fix `testContract` and `testContract'` to be more understandable by using a `TestOptions` parameter. `testContract = testContract' defaultOptions` etc.

incertia:
semantically the only testing code i actually changed was using `pmatch` instead of `-XLambdaCase` [here](https://github.com/crytic/echidna/pull/457/files#diff-011a42b5a00f70705ebff942934712c1R20).

PR #465, opened 2020-06-04
Add test for code size during contract creation (#453, #454)

elopez:


incertia:
we don't really need to run the entire fuzzing campaign for this contract. all we need is to construct the contract with `loadSolTests` and call the resulting test with `checkETest`. feel free to write an auxiliary function that checks all tests and call it `checkConstructorConditions` or something, or bake the functionality into `loadSolidity`.

src/test/Common.hs
```diff
@@ -62,6 +65,13 @@ testContract' fp n cfg s as = testCase fp $ do
   res <- runContract fp n c'
   mapM_ (\(t,f) -> assertBool t $ f res) as
 
+checkConstructorConditions :: FilePath -> String -> TestTree
+checkConstructorConditions fp as = testCase fp $ do
+  r <- flip runReaderT testConfig $ do
+    (v, _, t) <- loadSolTests (fromList [fp]) Nothing
```
incertia, line 71: `fp :| []`

src/test/Common.hs
```diff
@@ -23,16 +24,18 @@ import Test.Tasty.HUnit (testCase, assertBool)
 import Control.Lens (view, set, (.~))
 import Control.Monad.Reader (runReaderT)
 import Control.Monad.Random (getRandom)
+import Control.Monad.State.Strict (evalStateT)
 import Data.Function ((&))
 import Data.List (find)
-import Data.List.NonEmpty (NonEmpty(..))
+import Data.List.NonEmpty (NonEmpty(..), fromList)
```
incertia, line 30: just use the constructor (detailed below)

elopez:
The CI on macOS appears to be broken but it is unrelated to this change, see PR #468 for a possible fix.

PR #468, opened 2020-06-24
GitHub Actions: macOS: Only install needed packages

elopez:
This PR fixes the CI on macOS, which was failing with the following error:

    Warning: autoconf 2.69 is already installed and up-to-date
    To reinstall 2.69, run `brew reinstall autoconf`
    Error: libtool 2.4.6_1 is already installed
    To upgrade to 2.4.6_2, run `brew upgrade libtool`.


incertia:
[this file](https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md) appears to be out of date

Issue #474, opened 2020-08-12
Shift + PageUp/PageDown unexpectedly closes Echidna


ggrieco-tob:
If the curses dashboard is used, pressing Shift + PageUp/PageDown unexpectedly closes Echidna (same as ESC).

erivas:
The `defaultConfig` for vty doesn't have an input map sequence for "\ESC[5;2~" (Shift + PageUp) and "\ESC[5;2~" (Shift + PageDown), so the first thing parsed is "\ESC", and it leads echidna to halt. I'll submit a patch adding those combinations.

ggrieco-tob:
Fixed. Thanks @erivas !

PR #503, opened 2020-08-13
Fix inputMap (issue #474)

erivas:


erivas:
There seems to be an issue with the CI test for macos. Related to this https://github.com/mstksg/setup-stack/issues/7?

ggrieco-tob:
It seems that the CI issue was fixed. Can you update your PR with master so we can make sure it works as expected?

PR #506, opened 2020-08-17
Added missing line return

ggrieco-tob:
This will avoid this `Unique codehashes: XSeed: Y` in the console output


incertia:
feel free to merge even if the tests fail

PR #510, opened 2020-08-24
fix issue where libraries would get loaded at extreme timestamps and block numbers, also switch to the new hevm stripBytecodeMetadata

incertia:
`loadLibraries` would use `(initialTimestamp, initialBlockNumber)` as the delay property, causing `block . number` and `block . timestamp` to be incremented way more times than desired. we introduce a new `initialVM` that takes care of all this business for us so we can just use `(0, 0)`, as well as extend the timestamp and block number business to the `RPC` (multicontract, we should probably also rename this) module.

hevm-0.37 contains the new `stripBytecodeMetadata` that actually works and is a copy of the version i introduced a while back, so we remove that bit of duplicated code as well.

### DIFF
--- a/examples/solidity/basic/library.sol
+++ b/examples/solidity/basic/library.sol
@@ -1,23 +1,28 @@
-library Test{
+library Test {
     struct Storage{
         bool flag;
     }
 
-     function set(Storage storage st) public{
+    function set(Storage storage st) public{
         st.flag = true;
     }
 
  }
 
-contract Contract{
+contract Contract {
     using Test for Test.Storage;
     Test.Storage st;
 
-     function set() public{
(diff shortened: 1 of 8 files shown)


incertia:
b8bc129 adds some intermediate hevm-0.41 version that we should aim to change before we release echidna, or do a beta release

incertia:
newer hevm has been moved to #511 and this should just aim to fix some things

ggrieco-tob (commented):
This needs a test. I'm working on that right now.

ggrieco-tob:
@incertia please take a look to the new test

ggrieco-tob (approved):

PR #511, opened 2020-09-09
hevm 0.42

incertia:
**Edited by @ggrieco-tob**

This PR was updated to use hevm 0.42. There is one important point to dicuss: recent hevm releases are simulating the gas spent/refunded by the called once a transaction ends. This will break the balance unit test when the `gasprice` positive. Even using hevm-0.42, [this new feature needs some non-trivial modification in Echidna to subtract the balance from the caller once each call is complete](https://github.com/dapphub/dapptools/pull/512/files#diff-ed387fc665ccab94976051f0fb6da689b748541a8e7a748367a339afe512b559R339-R370).

I believe this detail is not important when you are testing your contracts (unless you depend on `gasprice`, which is unusual). Moreover, if you try to create/test a property using the balance of the sender (which is very usual, but it can happen), then you won't be able to calculate your exact balance after any transaction, if gasprice is positive. That's why I modified this PR to have the `gasprice` as 0 by default. If the user really need that, they can change it.


stack.yaml
```diff
@@ -1,30 +1,30 @@
...
-- HSH-2.1.3
-- ipprint-0.6
-- sr-extra-1.46.3.2
-- Unixutils-1.54.1
 - witherable-0.3.5
 - witherable-class-0
 - github: dmjio/semver-range
   commit: d8d9db892ddb6ae267c9bcbc4f6602668433f12a
+- github: leventerkok/sbv # 8.7.5
+  # TODO: sbv 8.8
```
arcz, line 26: what happened here?

lib/Echidna/Types/Buffer.hs
```diff
@@ -0,0 +1,7 @@
+module Echidna.Types.Buffer where
+
+import Data.ByteString (ByteString)
+import EVM.Symbolic (Buffer(..))
+
+viewBuffer :: Buffer -> Maybe ByteString
+viewBuffer = undefined
```
arcz: why this is undefined?

lib/Echidna/UI.hs
```diff
@@ -111,7 +111,9 @@ ui v w ts d txs = do
           Just _ -> liftIO $ updateUI CampaignUpdated
         )
         (const $ liftIO $ killThread ticker)
-      app <- customMain (mkVty vtyConfig) (Just bc) <$> monitor
+      let vty = mkVty vtyConfig
+      initialVty <- liftIO vty
+      app <- customMain initialVty vty (Just bc) <$> monitor
```
arcz, line 116: is this a code to support new brick?

incertia: customMain now takes vty and IO to make new vty

stack.yaml
incertia: hevm currently pins 8.7.5 (unofficial release) so we should do the same until they pin 8.8 and release, whereby we can also update and release

incertia:
looks like balance fails somehow i'll look into it some time

ggrieco-tob:
> looks like balance fails somehow i'll look into it some time

I believe this is blocked by https://github.com/dapphub/dapptools/issues/511

arcz:
There is hevm 0.42.0 released 2 days ago https://github.com/dapphub/dapptools/releases/tag/hevm%2F0.42.0. It should fix the gas price issue as this was closed https://github.com/dapphub/dapptools/issues/511.

stack.yaml
arcz: @ggrieco-tob we can use 8.8 as hevm relaxed this dependency since then

lib/Echidna/Transaction.hs
```diff
@@ -84,12 +87,12 @@ genTxM :: (MonadRandom m, MonadReader x m, Has TxConf x, MonadState y m, Has Gen
   => Map Addr Contract
   -> m Tx
 genTxM m = do
-  TxConf _ g maxGp t b mv <- view hasLens
+  TxConf _ g gp t b mv <- view hasLens
   genTxWith
     m
     rElem rElem                                                                -- src and dst
     (const $ genInteractionsM . snd)                                           -- call itself
-    (pure g) (inRange maxGp) mv                                                -- gas, gasprice, value
```
arcz, line 92: is this still an issue?

lib/Echidna/Types/Tx.hs
```diff
@@ -109,3 +112,6 @@ getResult (VMFailure InvalidMemoryAccess)       = ErrorInvalidMemoryAccess
 getResult (VMFailure CallDepthLimitReached)     = ErrorCallDepthLimitReached
 getResult (VMFailure (MaxCodeSizeExceeded _ _)) = ErrorMaxCodeSizeExceeded
 getResult (VMFailure PrecompileFailure)         = ErrorPrecompileFailure
+getResult (VMFailure UnexpectedSymbolicArg)     = ErrorUnexpectedSymbolic
+getResult (VMFailure DeadPath)                  = ErrorDeadPath
+getResult (VMFailure (Choose _))                = ErrorChoose -- not entirely sure what this is
```
arcz, line 117: https://github.com/dapphub/dapptools/blob/786f8cac6450de29ddc92ab59a5056fad44abee0/src/hevm/src/EVM.hs#L1537-L1538

ggrieco-tob: That's if you use symbolic execution. I will remove that comment.

lib/Echidna/Transaction.hs
ggrieco-tob: What exactly?

arcz: gasprice, but I've read the updated description and see what's the problem

stack.yaml
ggrieco-tob: Done! Keep in mind that in practice won't change anything in Echidna, since symbolic execution is not used.

arcz (approved):

ggrieco-tob (approved):

PR #516, opened 2020-09-28
Source coverage printing after fuzzing campaign

ggrieco-tob:
This PR allows Echidna to use source information from hevm to print which lines were covered during a fuzzing campaign. There is no need to configure anything, the source information is included in the crytic-compile output. The user should only enable coverage and `corpusDir`. However, the user must update to [crytic-compile 0.1.10](https://github.com/crytic/crytic-compile/releases/tag/0.1.10) (or newer) to make sure the source files are properly sorted when compiling.

The directly structure for the corpus directly was slight changed. For instance, after running `flags.sol`, the corpus directory will looks like this:

```
$ ls corpus/ -R
corpus/:
coverage  covered.txt

corpus/coverage:
-4293026413532156583.txt  4589328475588012129.txt
```

where the `covered.txt` file looks like this:

```
examples/solidity/basic/flags.sol
*r  |contract Test {
  bool private flag0 = true;
  bool private flag1 = true;

*   |  function set0(int val) public returns (bool){
*   |    if (val % 100 == 0)
*   |      flag0 = false;
  }

*   |  function set1(int val) public returns (bool){
*   |    if (val % 10 == 0 && !flag0)
*   |      flag1 = false;
  }

  function echidna_alwaystrue() public returns (bool){
    return(true);
  }

  function echidna_revert_always() public returns (bool){
    revert();
  }

  function echidna_sometimesfalse() public returns (bool){
    return(flag1);
  }

}
```

Constructors (#539)  and echidna properties are not counting coverage. This should be fixed in another PR.

Each reachable line will have zero or more 'markers', signaling if how the sequence of transactions ended when covering that line. So far, we have support for the following markers:

- `*` for sequences ended with a return.
- `r` for sequences ended with a revert.
- `o` for sequences ended with an out-of-gas error.
- `e` for every other type of error (e.g. self-destruction, insufficient balance, etc).

It also introduced a small refactoring to move the corpus functions into a separated module inside `Output`.


lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,123 @@
...
+saveCoveredCode :: Maybe FilePath -> SourceCache -> [SolcContract] -> CoverageMap -> IO ()
+saveCoveredCode (Just d) sc cs s = let fn = d ++ "/covered.txt"
+                                       cc = ppCoveredCode sc cs s 
+                                   in writeFile fn cc
+saveCoveredCode Nothing  _  _  _ = pure ()
+
+
+-- | Pretty-print the covered code
+ppCoveredCode :: SourceCache -> [SolcContract] -> CoverageMap -> Text
+ppCoveredCode sc cs s | s == mempty = "Coverage map is empty"
```
arcz, line 41: shouldn't this fail if CoverageMap is empty?

lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,123 @@
...
+                      | otherwise   =
+  let allLines = M.toList $ view sourceLines sc
+  -- ^ Collect all the possible lines from all the files
+      findFile k = fst $ M.findWithDefault ("<no source code>", mempty) k (view sourceFiles sc)
```
arcz, line 45: `(view sourceFiles sc)` -> `(sc ^. sourceFiles)`

also, is it possible that there is no source code for a file?

lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,123 @@
...
+markerHeader :: String
+markerHeader = "    |"
+
+-- | Add space for markers if necessary
+checkMarkers  :: String -> Text
+checkMarkers l@(_:_:_:_:'|':_) = pack l
+checkMarkers l                 = pack $ markerHeader ++ l
+
+-- | Add a proper marker to a line, and convert it to Text
+markStringLine  :: TxResult -> String -> Text
```
arcz, line 70: would it be simpler to gather all markers into a set for each line and later render the line in one pass based on markers in the set?

lib/Echidna.hs
```diff
@@ -39,14 +40,14 @@ import qualified Data.List.NonEmpty as NE
 -- * A list of transaction sequences to initialize the corpus
 prepareContract :: (MonadCatch m, MonadRandom m, MonadReader x m, MonadIO m, MonadFail m,
                     Has TxConf x, Has SolConf x)
-                => EConfig -> NE.NonEmpty FilePath -> Maybe String -> Seed -> m (VM, World, [SolTest], Maybe GenDict, [[Tx]])
+                => EConfig -> NE.NonEmpty FilePath -> Maybe String -> Seed -> m (VM, SourceCache, [SolcContract], World, [SolTest], Maybe GenDict, [[Tx]])
```
arcz, line 43: maybe `prepareContract` should take `[SolcContract]` provided by prior calling of `Echidna.Solidity.contracts` as an argument rather than `FilePath`s?

lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,123 @@
...
+
+-- | Select the proper marker, according to the result of the transaction
+getMarker :: TxResult -> Char
+getMarker Success       = '*'
+getMarker ErrorRevert   = 'r' 
+getMarker ErrorOutOfGas = 'o'
+getMarker _             = 'e'
+
+-- | Filter the lines per file, marking each line
+filterLines :: [Maybe (FilePathText, Int, TxResult)] -> [(FilePathText, Text)] -> [(FilePathText, Text)]
```
arcz, line 99: can we move this function before `markLine` to be consistent and read more general functions first?

lib/Echidna/Solidity.hs
```diff
@@ -126,22 +127,27 @@ contracts fp = let usual = ["--solc-disable-warnings", "--export-format", "solc"
...
               ExitFailure _ -> throwM $ CompileFailure out err
-          maybe (throwM SolcReadFailure) (pure . toList . fst) mSolc
-    concat <$> sequence (compileOne <$> fps)
+            
+          maybe (throwM SolcReadFailure) (pure . first toList) mSolc
+    cps <- mapM compileOne fps
+    let (cs, ss) = unzip cps
+    let lss = length ss
+    when (lss > 1) $ liftIO $ putStrLn "WARNING: more than one SourceCache was found after compile. Only the first one will be used."
+    return (concat cs, head ss)
```
arcz, line 144: `return` -> `pure`, there is partial `head` maybe we could `throwM` instead?

lib/Echidna/Exec.hs
```diff
@@ -11,13 +11,14 @@ import Control.Lens
...
+import Data.Maybe (fromMaybe, fromJust)
 import Data.Map.Strict (Map)
-import Data.Maybe (fromMaybe)
 import Data.Set (Set)
+import Data.Tuple.Extra (fst3)
 import EVM
 import EVM.Op (Op(..))
 import EVM.Exec (exec, vmForEthrunCreation)
-import EVM.Solidity (stripBytecodeMetadata)
+import EVM.Solidity --(SourceCache, SrcMap, SolcContract, contractName, sourceLines, sourceFiles, runtimeCode, runtimeSrcmap, creationSrcmap)
```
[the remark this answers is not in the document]
ggrieco-tob: Fixed

lib/Echidna/Exec.hs
```diff
@@ -98,7 +99,7 @@ execTxWith h m t = do
 execTx :: (MonadState x m, Has VM x, MonadThrow m) => Tx -> m (VMResult, Int)
 execTx = execTxWith vmExcept $ liftSH exec
 
-type CoverageMap = Map BS.ByteString (Set (Int, TxResult))
+type CoverageMap = Map BS.ByteString (Set (Int, Int, TxResult))
```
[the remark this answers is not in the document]
ggrieco-tob: Using a tuple is faster so I named the types.

lib/Echidna/Output/Source.hs
ggrieco-tob: It is better to show a message in the file, than to have a hard error here.

lib/Echidna/Solidity.hs
```diff
@@ -183,8 +189,8 @@ abiOf pref cc = fallback NE.:| filter (not . isPrefixOf pref . fst) (elems (cc ^
 -- usable for Echidna. NOTE: Contract names passed to this function should be prefixed by the
 -- filename their code is in, plus a colon.
 loadSpecified :: (MonadIO m, MonadThrow m, MonadReader x m, Has SolConf x, Has TxConf x, MonadFail m)
-              => Maybe Text -> [SolcContract] -> m (VM, EventMap, NE.NonEmpty SolSignature, [Text], SignatureMap)
-loadSpecified name cs = do
+              => Maybe Text -> ([SolcContract], SourceCache) -> m (VM, EventMap, NE.NonEmpty SolSignature, [Text], SignatureMap)
+loadSpecified name (cs,_) = do
```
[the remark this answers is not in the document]
ggrieco-tob: Good catch. I removed that argument.

lib/Echidna/Output/Source.hs
ggrieco-tob: > (view sourceFiles sc) -> (sc ^. sourceFiles)

Fixed.

> also, is it possible that there is no source code for a file?

That should be a bug, but it depends from the compilation process, so could be.

lib/Echidna/Solidity.hs
ggrieco-tob: > return -> pure,

Fixed.

>  there is partial head maybe we could throwM instead?


I believe `ss` cannot be empty, otherwise the result of the compilation will be "invalid".

lib/Echidna/Output/Source.hs
ggrieco-tob: Done.

lib/Echidna/Output/Source.hs
ggrieco-tob: Yes, perhaps that will make `markStringLine` simpler, but I believe it will also make the other functions more complex. This approach keeps the types as we have them directly from the hevm functions that are used to collect the coverage information, but feel free to provide some code on how to improve it. :smile:

lib/Echidna.hs
ggrieco-tob: What do you mean exactly?

lib/Echidna/Campaign.hs
```diff
@@ -161,7 +161,7 @@ execTxOptC t = do
   res <- execTxWith vmExcept (usingCoverage $ pointCoverage cov) t
   let vmr = getResult $ fst res
   -- Update the coverage map with the proper binary according to the vm result
-  cov %= mapWithKey (\_ s -> DS.map (set _2 vmr) s)
+  cov %= mapWithKey (\_ s -> DS.map (\(i, j, _) -> (i, j, vmr)) s)
```
incertia, line 164: set _3

lib/Echidna/Exec.hs
```diff
@@ -113,13 +119,13 @@ coveragePoints = sum . fmap S.size
...
 scoveragePoints :: CoverageMap -> Int
-scoveragePoints = sum . fmap (S.size . S.map fst)
+scoveragePoints = sum . fmap (S.size . S.map fst3)
 
 -- | Capture the current PC and bytecode (without metadata). This should identify instructions uniquely.
 pointCoverage :: (MonadState x m, Has VM x) => Lens' x CoverageMap -> m ()
 pointCoverage l = do
   v <- use hasLens
-  l %= M.insertWith (const . S.insert $ (v ^. state . pc, Success))
+  l %= M.insertWith (const . S.insert $ ( v ^. state . pc, fromJust $ vmOpIx v, Success))
```
incertia, line 128: rogue space

lib/Echidna/Output/JSON.hs
```diff
@@ -22,7 +22,7 @@ data Campaign = Campaign
   , _error :: Maybe String
   , _tests :: [Test]
   , seed :: Int
-  , coverage :: Map String [(Int, TxResult)]
+  , coverage :: Map String [(Int, Int, TxResult)]
```
incertia, line 25: use your new types?

lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,122 @@
+{-# LANGUAGE DataKinds #-}
```
incertia, line 1: DataKinds?

lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,122 @@
...
+
+-- | Filter the lines per file, marking each line
+filterLines :: [Maybe (FilePathText, Int, TxResult)] -> [(FilePathText, Text)] -> [(FilePathText, Text)]
+filterLines []                  ls  = ls
+filterLines (Nothing      : ns) ls  = filterLines ns ls
+filterLines (Just (f,n,r) : ns) ls  = filterLines ns (markLine n r f ls)
+
+-- | Mark one particular line, from a list of lines, keeping the order of them 
+markLine :: Int -> TxResult -> FilePathText -> [(FilePathText, Text)] -> [(FilePathText, Text)] 
+markLine n r cf ls = case splitAt (n-1) ls of
```
incertia, line 62: let's make this multiline and indent a bit less

lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,122 @@
...
+
+-- | Add space for markers if necessary
+checkMarkers  :: String -> Text
+checkMarkers l@(_:_:_:_:'|':_) = pack l
+checkMarkers l                 = pack $ markerHeader ++ l
+
+-- | Add a proper marker to a line, and convert it to Text
+markStringLine  :: TxResult -> String -> Text
+markStringLine r (' ': ' ': ' ': ' ': '|':l) = pack $ getMarker r : ' ' : ' ' : ' ': '|' : l
+markStringLine r (m1: ' ': ' ': ' ':  '|':l)  = pack $ case getMarker r of
```
incertia, line 78: consistent spacing (please check this entire file)

lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,122 @@
...
+-- | Select the proper marker, according to the result of the transaction
+getMarker :: TxResult -> Char
+getMarker Success       = '*'
+getMarker ErrorRevert   = 'r' 
+getMarker ErrorOutOfGas = 'o'
+getMarker _             = 'e'
+
+-- | Given a source cache, a coverage map, a contract returns a list of covered lines
+srcMapCov :: SourceCache -> CoverageMap -> SolcContract -> [Maybe (FilePathText, Int, TxResult)]
+srcMapCov sc s c = nub $  -- Deduplicate results
```
incertia, line 106: `. . $` and align `.` to `=` or indent extra 2 spaces

lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,122 @@
...
+srcMapCov :: SourceCache -> CoverageMap -> SolcContract -> [Maybe (FilePathText, Int, TxResult)]
+srcMapCov sc s c = nub $  -- Deduplicate results 
+                     map (srcMapCodePosResult sc) $ -- Get the filename, number of line and tx result
+                       mapMaybe (srcMapForOpLocation c) $ -- Get the mapped line and tx result
+                         S.toList $ fromMaybe S.empty $    -- Convert from Set to list
+                             M.lookup (getBytecodeMetadata $ c ^. runtimeCode) s -- Get the coverage information of the current contract
+
+-- | Given a source cache, a mapped line, return a tuple with the filename, number of line and tx result
+srcMapCodePosResult :: SourceCache -> (SrcMap, TxResult) -> Maybe (Text, Int, TxResult)
+srcMapCodePosResult sc (n, r) = case srcMapCodePos sc n of
```
incertia, line 114: indentation

lib/Echidna/Solidity.hs
```diff
@@ -126,22 +127,26 @@ contracts fp = let usual = ["--solc-disable-warnings", "--export-format", "solc"
...
             case ec of
               ExitSuccess -> readSolc "crytic-export/combined_solc.json"
               ExitFailure _ -> throwM $ CompileFailure out err
-          maybe (throwM SolcReadFailure) (pure . toList . fst) mSolc
-    concat <$> sequence (compileOne <$> fps)
+            
+          maybe (throwM SolcReadFailure) (pure . first toList) mSolc
+    cps <- mapM compileOne fps
+    let (cs, ss) = unzip cps
+    when (length ss > 1) $ liftIO $ putStrLn "WARNING: more than one SourceCache was found after compile. Only the first one will be used."
```
incertia, line 142: do we maybe want `stderr` here

lib/Echidna/Output/Corpus.hs
```diff
@@ -0,0 +1,47 @@
...
+import Data.Hashable (hash)
+import Data.Maybe (catMaybes)
+import System.Directory (createDirectoryIfMissing)
+
+import qualified System.Directory as SD
+import qualified Data.ByteString as BS
+
+import Echidna.Types.Tx
+
+listDirectory :: FilePath -> IO [FilePath]
```
incertia, line 17: maybe move these directory functions to a util module

lib/Echidna/Output/Source.hs
```diff
@@ -0,0 +1,122 @@
...
+import qualified Data.Map as M
+import qualified Data.Set as S
+
+import Echidna.Exec 
+import Echidna.Types.Tx
+import Echidna.Types.Signature (getBytecodeMetadata)
+
+type FilePathText = Text
+
+saveCoveredCode :: Maybe FilePath -> SourceCache -> [SolcContract] -> CoverageMap -> IO ()
```
incertia, line 32: maybe rename this to `saveCoverage`

src/test/Common.hs
```diff
@@ -73,7 +73,7 @@ runContract :: FilePath -> Maybe ContractName -> EConfig -> IO Campaign
 runContract f c cfg =
   flip runReaderT cfg $ do
     g <- getRandom
-    (v, w, ts, d, txs) <- prepareContract cfg (f :| []) c g
+    (v, _, _, w, ts, d, txs) <- prepareContract cfg (f :| []) c g
     -- start ui and run tests
     campaign (pure ()) v w ts d txs
```
incertia, line 78: now that we have a bunch of things from `prepareContract` can we make a custom type for `prepareContract` and `campaign` and `ui` to ingest the custom type instead of adding a new parameter every time? feel free to make an issue and refactor in a separate PR.

lib/Echidna/Types/Signature.hs
```diff
@@ -2,8 +2,8 @@ module Echidna.Types.Signature where
 
 import Data.Foldable (find)
 import Data.HashMap.Strict (HashMap)
-import Data.ByteString (ByteString)
 import Data.List.NonEmpty (NonEmpty)
+import Data.ByteString (ByteString)
```
incertia, line 6: alphabetize these imports

lib/Echidna/Transaction.hs
```diff
@@ -11,24 +11,20 @@ module Echidna.Transaction where
 import Prelude hiding (Word)
 
 import Control.Lens
-import Control.Monad (join, liftM2, unless)
-import Control.Monad.Catch (bracket)
 import Control.Monad.Random.Strict (MonadRandom, getRandomR, uniform)
 import Control.Monad.Reader.Class (MonadReader)
 import Control.Monad.State.Strict (MonadState, State, runState, get, put)
-import Data.Aeson (ToJSON(..), decodeStrict, encodeFile)
+import Control.Monad (join, liftM2)
```
incertia, line 17: alphabetize imports

incertia (changes requested):

lib/Echidna/Campaign.hs
ggrieco-tob: Fixed.

lib/Echidna/Exec.hs
ggrieco-tob: Fixed.

lib/Echidna/Output/Source.hs
ggrieco-tob: Removed.

lib/Echidna/Output/Source.hs
ggrieco-tob: I think now it is better

lib/Echidna/Output/Source.hs
ggrieco-tob: Fixed.

lib/Echidna/Output/Source.hs
ggrieco-tob: Fixed

lib/Echidna/Solidity.hs
ggrieco-tob: We could, but I would prefer that the user is annoyed with this (and perhaps report it as an issue), because I don't know if it can happen.

lib/Echidna/Output/Source.hs
ggrieco-tob: Fixed.

lib/Echidna/Types/Signature.hs
ggrieco-tob: Done.

lib/Echidna/Output/Source.hs
ggrieco-tob: I think that's fixed.

lib/Echidna/Output/Corpus.hs
ggrieco-tob: Fixed.

src/test/Common.hs
ggrieco-tob: Create a new issue to track down this.

PR #518, opened 2020-10-03
Added explicit delay in shrinking and improved pp functions

ggrieco-tob:
This small PR allows Echidna to use a special transaction to discover during shrinking when a function called is actually useless and only the time/block delay is triggering an failure. It will also improve a little how calls are printed to make them easier to understand.

### DIFF
--- a/lib/Echidna/Output/JSON.hs
+++ b/lib/Echidna/Output/JSON.hs
@@ -126,4 +126,5 @@ mapTest (solTest, testState) =
 
   mapCall (SolCreate _) = ("<CREATE>", Nothing)
   mapCall (SolCall (name, args)) = (name, Just $ ppAbiValue <$> args)
+  mapCall NoCall                 = ("*delay*", Nothing)
   mapCall (SolCalldata x) = (decodeUtf8 $ "0x" <> BS16.encode x, Nothing)
--- a/lib/Echidna/Output/JSON.hs
+++ b/lib/Echidna/Output/JSON.hs
@@ -126,5 +126,5 @@ mapTest (solTest, testState) =
 
   mapCall (SolCreate _) = ("<CREATE>", Nothing)
   mapCall (SolCall (name, args)) = (name, Just $ ppAbiValue <$> args)
-  mapCall NoCall                 = ("*delay*", Nothing)
+  mapCall NoCall                 = ("*wait*", Nothing)
   mapCall (SolCalldata x) = (decodeUtf8 $ "0x" <> BS16.encode x, Nothing)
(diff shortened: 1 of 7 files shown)


lib/Echidna/Types/Tx.hs
```diff
@@ -19,6 +19,7 @@ import Echidna.Types.Signature (SolCall)
 data TxCall = SolCreate   ByteString
             | SolCall     SolCall
             | SolCalldata ByteString
+            | NoCall
```
arcz, line 22: I get the semantics from the context of this PR but I think this data constructor can be confusing in future. Can we document the semantics here? 

Looking further, as this is not really a call but rather other event altering the VM state maybe we could refactor that into something like:
```haskell
data TemporalEvent = AddDelay
data EventType = TxCall TxCall | TemporalEvent TemporalEvent
```
This would also eliminate the need of `_              -> error "NoCall"` in `setupTx`. This could involve a bigger refactor to make space for future expansions as `setupTx` and `Tx` type don't handle just Tx calls anymore.

lib/Echidna/Transaction.hs
```diff
@@ -108,8 +108,12 @@ canShrinkTx :: Tx -> Bool
 canShrinkTx (Tx (SolCreate _) _ _ _ 0 0 (0, 0))   = False
 canShrinkTx (Tx (SolCall (_,l)) _ _ _ 0 0 (0, 0)) = any canShrinkAbiValue l
 canShrinkTx (Tx (SolCalldata _) _ _ _ 0 0 (0, 0)) = False
+canShrinkTx (Tx NoCall _ _ _ _ _ (0, 0))          = False
```
arcz, line 111: does shrinker remove NoCalls with (0,0) delay?

lib/Echidna/Output/JSON.hs
```diff
@@ -126,4 +126,5 @@ mapTest (solTest, testState) =
 
   mapCall (SolCreate _) = ("<CREATE>", Nothing)
   mapCall (SolCall (name, args)) = (name, Just $ ppAbiValue <$> args)
+  mapCall NoCall                 = ("*delay*", Nothing)
```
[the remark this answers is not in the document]
ggrieco-tob: Good catch!

lib/Echidna/Types/Tx.hs
ggrieco-tob: Yes, that's a reasonable idea, but yes, it will require a larger refactor, for such a small feature. This should be the only "special" event

lib/Echidna/Transaction.hs
ggrieco-tob: `Tx NoCall _ _ _ _ _ (0, 0)` should be removed, but it cannot be reduced more. The shrinker takes care of removing useless transactions very quickly, so this is not a problem.

arcz (approved):

PR #523, opened 2020-10-21
Fixed the use of large constants in the dictionary

ggrieco-tob:
This small PR will enable the use of max values per type (`uint256` -> `2 ** 256 - 1`) in the dictionary by default, since these are interesting values for test.

### DIFF
--- a/lib/Echidna/Solidity.hs
+++ b/lib/Echidna/Solidity.hs
@@ -293,25 +293,28 @@ loadSolTests :: (MonadIO m, MonadThrow m, MonadReader x m, Has SolConf x, Has Tx
              => NE.NonEmpty FilePath -> Maybe Text -> m (VM, World, [SolTest])
 loadSolTests fp name = loadWithCryticCompile fp name >>= (\t -> prepareForTest t Nothing [])
 
+commonTypeSizes :: [Int]
+commonTypeSizes = [8,16..256]
+
 mkValidAbiInt :: Int -> Int256 -> Maybe AbiValue
 mkValidAbiInt i x = if abs x <= 2 ^ (i - 1) - 1 then Just $ AbiInt i x else Nothing
 
 mkLargeAbiInt :: Int -> AbiValue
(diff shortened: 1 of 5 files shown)


lib/Echidna/Solidity.hs
```diff
@@ -304,6 +310,9 @@ timeConstants = concatMap dec [initialTimestamp, initialBlockNumber]
   where dec i = let l f = f <$> [8,16..256] <*> fmap fromIntegral [i-1..i+1] in
                 catMaybes (l mkValidAbiInt ++ l mkValidAbiUInt)
 
+largeConstants :: [AbiValue]
+largeConstants = concatMap (\i -> [mkLargeAbiInt i, mkLargeAbiUInt i]) [8,16..256]
```
incertia, line 314: let's pull out `[8, 16, ...]` into a variable

lib/Echidna/Solidity.hs
```diff
@@ -296,6 +296,12 @@ loadSolTests fp name = loadWithCryticCompile fp name >>= (\t -> prepareForTest t
 mkValidAbiInt :: Int -> Int256 -> Maybe AbiValue
 mkValidAbiInt i x = if abs x <= 2 ^ (i - 1) - 1 then Just $ AbiInt i x else Nothing
 
+mkLargeAbiInt :: Int -> AbiValue
+mkLargeAbiInt i = AbiInt i $  2 ^ (i - 1) - 1
+
+mkLargeAbiUInt :: Int -> AbiValue
+mkLargeAbiUInt i = AbiUInt i $ (2 ^ i) - 1
```
incertia, line 303: we don't need parens here

incertia (changes requested):

lib/Echidna/Solidity.hs
ggrieco-tob: It's odd that hlint won't warn about it

lib/Echidna/Solidity.hs

lib/Echidna/Solidity.hs

incertia (approved):

Issue #525, opened 2020-10-21
MacOS CI is failing


ggrieco-tob:
https://github.com/crytic/echidna/pull/523/checks?check_run_id=1288048270

ggrieco-tob:
MacOS is no longer failing (because it is disabled)

PR #526, opened 2020-10-21
GitHub Actions: fix macOS build

elopez:
Currently the CI system fails on macOS with several errors related
to Homebrew and cache corruption. This implements temporal
workarounds for both issues. For more details, see:

 * https://github.com/actions/virtual-environments/issues/1811
 * https://github.com/actions/cache/issues/403

Closes: #525

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,12 @@ jobs:
           brew untap local/homebrew-openssl
           brew untap local/homebrew-python2
 
+      - name: Workaround for actions/cache#403
+        if: runner.os == 'macOS'
+        run: |
+          brew install gnu-tar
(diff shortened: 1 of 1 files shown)


elopez:
Looks like there is a second issue here; if the macOS pipeline is run twice, it fails the second time with some strange error when trying to run Cabal. I thought it could be some older binaries on the cache, but it does not look like that's the case. I'll continue investigating soon.

elopez:
I think this should be it now. Give it a pipeline rerun, just in case :smile:

incertia:
on a personal level i just want to ignore anything macos related until github gets their shit together but this looks good

ggrieco-tob:
macOS tests are temporarily disabled now, but we can use this PR for the fix, when we re-enabled them

incertia:
seems like macOS builds are running fine now, is it ok if we close this?

elopez:
yeah, this got merged through #527, closing then :smile:

Issue #522, opened 2020-10-21
dyld: Library not loaded: libff.dylib


PaulRBerg:
I ran these steps:

1. Download the v1.5.1 release from the [Releases][1] page
2. Unzip the tar.gz archive
3. Open the terminal and run `./echidna-test --version`

And got this error:

> dyld: Library not loaded: libff.dylib
> Referenced from: /Users/paulrberg/Binaries/./echidna-test
> Reason: image not found
> Abort trap: 6

Using macOS@10.15.7.

[1]: https://github.com/crytic/echidna/releases/tag/v1.5.1

ggrieco-tob:
Hi!, thanks for testing our binary for OSX. You should have two additional libraries installed:

* libff (https://github.com/crytic/echidna/blob/master/.github/scripts/install-libff.sh) 
* libsecp256k1 (https://github.com/crytic/echidna/blob/master/.github/scripts/install-libsecp256k1.sh)

Can you run these scripts to check if it solves this. Otherwise, you can use our docker container to run Echidna in OSX.

PaulRBerg:
Thanks @ggrieco-tob, I ended up using Docker, it was much easier.

PaulRBerg:
For what it's worth, I just ran both of those scripts:

```sh
HOST_OS=macOS ./.github/scripts/install-libff.sh
HOST_OS=macOS ./.github/scripts/install-libsecp256k1.sh
```

And I reloaded my terminal, but I'm still getting the same error as in my original post.

ggrieco-tob:
Uhm, interesting. We will investigate this issue. Let's keep this issue open. Thanks for the report. :slightly_smiling_face:

PaulRBerg:
Happy to help. Sorry to bug you - but do you happen to know what's going on in this other issue?

https://github.com/crytic/crytic-compile/issues/115

ggrieco-tob:
Sure, let me check that.

fabiohild:
Ran into the same problem today. Installing libff with the bash script or from other sources does not seem to work. I'm on MacOS Mohave. Let me know if I can provide any useful info, will use the Docker container for now. Thanks!

elopez:
@fabiohild @PaulRBerg could you try out the binaries from the draft PR above and see if they work any better? You should not need to build/install libff nor libsecp256k1, as they are bundled in the zip file.

https://github.com/crytic/echidna/actions/runs/323365632

fabiohild:
[...]

PR #527, opened 2020-10-22
GitHub Actions: improve macOS releases

elopez:
Currently, releases for macOS are not very portable. The
binaries have hardcoded paths and might not run on other
systems. This commit improves the macOS binary releases
made via GitHub actions by bundling binary copies of libff,
libgmp and libsecp256k1, and by adjusting the rpath and
library names to make echidna run on other systems.

### DIFF
--- a/.github/scripts/build-macos-release.sh
+++ b/.github/scripts/build-macos-release.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+add_rpath()
+{
+    BINARY="$1"
+    install_name_tool -add_rpath "@executable_path/." "$BINARY"
+}
+
+fix_path()
+{
+    BINARY="$1"
+    MATCH="$2"
+    NEW="$3"
+    OLD=$(otool -L "$BINARY" | grep "$MATCH" | awk '{print $1}')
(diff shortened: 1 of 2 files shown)


elopez:
I've marked this as draft, because it builds on top of PR #526, but I'd love to hear your thoughts on implementing something like the top commit here ("GitHub Actions: improve macOS releases"). This is meant to help resolve situations like the one in #522.

incertia:
where exactly does it put the rpath? does it bundle it as a 3-file group in `.` or will it check for libraries in `/usr/local`? it's probably easier to do the former.

elopez:
It will add `@executable_path/.` as rpath on echidna-test, and then do a 3-file bundle. You can have a look at how the files look on the CI run here: https://github.com/crytic/echidna/actions/runs/323365632

ggrieco-tob:
Why this is still a draft? Are we waiting for the #525 to be merged?

elopez:
I had built this branch on top of #526 and that's why I had marked it as a draft. I'll open this one in case you want to handle both in one review 🙂

incertia (approved):

Issue #531, opened 2020-10-23
Feature request: allow testing of very large contracts


moodysalem:
Per https://github.com/crytic/echidna/issues/401

If you have a test case that depends on a whole bunch of other contracts, you will run into the block gas limit trying to test it with echidna

The recommendation is to use etheno + truffle to create an initialization file https://github.com/crytic/echidna/wiki/Deployment-of-a-contract-using-Truffle,-Ganache-and-Etheno-to-test-with-Echidna

However if you do not use truffle, this is a hassle to set up. Prefer to just be able to configure the block gas limit used in the EVM, so can deploy arbitrarily large contracts.

ggrieco-tob:
Please test #532 that will use a very very large gas limit (over the gas block limit) for deployment.

moodysalem:
the artifact that is built from that pr works for me

ggrieco-tob:
Fixed.

PR #532, opened 2020-10-23
Refactor Tx constants to use in deployment and config

ggrieco-tob:
* Refactor constants to have clear names.
* Increase max gas block from 8000030 to 12500000.
* Use virtually unlimited gas during deployment (as a workaround when users need to deploy large contracts).

### DIFF
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -33,7 +33,7 @@ import qualified Data.Yaml as Y
 import Echidna.Solidity
 import Echidna.Test
 import Echidna.Types.Campaign (CampaignConf(CampaignConf))
-import Echidna.Types.Tx
+import Echidna.Types.Tx  (TxConf(TxConf), maxGasPerBlock, defaultTimeDelay, defaultBlockDelay)
 import Echidna.UI
 import Echidna.UI.Report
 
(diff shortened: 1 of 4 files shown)


moodysalem:
closes https://github.com/crytic/echidna/issues/531

lib/Echidna/Config.hs
```diff
@@ -33,7 +33,7 @@ import qualified Data.Yaml as Y
 import Echidna.Solidity
 import Echidna.Test
 import Echidna.Types.Campaign (CampaignConf(CampaignConf))
-import Echidna.Types.Tx (TxConf(TxConf))
+import Echidna.Types.Tx
```
incertia, line 36: let's continue using qualified imports

we can replace things with `(..)` if necessary

lib/Echidna/Types/Tx.hs
```diff
@@ -23,6 +23,18 @@ data TxCall = SolCreate   ByteString
 makePrisms ''TxCall
 $(deriveJSON defaultOptions ''TxCall)
 
+maxGasPerBlock :: Integer
+maxGasPerBlock = 12500000
```
incertia, line 27: any reason for 8000030 ->12500000 besides a less random number?

incertia (changes requested):

lib/Echidna/Types/Tx.hs
ggrieco-tob: This was recently [increased by miners](https://cointelegraph.com/news/ethereum-miners-vote-to-increase-gas-limit-causing-community-debate).

lib/Echidna/Config.hs
ggrieco-tob: Fixed.

incertia (approved):

PR #535, opened 2020-10-30
Warn the user if slither is not installed

ggrieco-tob:
This small PR adds a explicit message to warn the user if slither is not installed.

### DIFF
--- a/lib/Echidna.hs
+++ b/lib/Echidna.hs
@@ -5,7 +5,7 @@ module Echidna where
 import Control.Lens (view, (^.), to)
 import Data.Has (Has(..))
 import Control.Monad.Catch (MonadCatch(..))
-import Control.Monad.Reader (MonadReader, MonadIO, liftIO)
+import Control.Monad.Reader (MonadReader, MonadIO, liftIO, when)
 import Control.Monad.Random (MonadRandom)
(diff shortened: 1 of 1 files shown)


lib/Echidna.hs
```diff
@@ -52,6 +52,7 @@ prepareContract cfg fs c g = do
   -- run processors
   ca <- view (hasLens . cryticArgs)
   si <- runSlither (NE.head fs) ca
+  when (null si) $ liftIO $ putStrLn "WARNING: slither failed to run or extracted no information at all"
```
arcz, line 55: Wouldn't it be better to put this message inside `runSlither` function? There is a separate case when slither is not found. I find also the message lacking a bit in terms of information. What about something like: "slither was not found on PATH, it is optional but installing it will give you better results <link to slither>"

ggrieco-tob: We don't want to hard-fail if slither is not found (so slither is not technically required to run Echidna). Also, it could be the case that Slither is too old (or it has a bug) and it doesn't return anything at all. This should cover everything.

arcz: ok, we can always tweak that

arcz (approved):

arcz (approved):

PR #538, opened 2020-11-05
Use more the values from the dictionary 

ggrieco-tob:
This PR allows echidna to use numeric values from the dictionary to generate:

* The amount to pay to payable function.
* The number to increase the blocks number.
* The number of seconds to increase the block timestamp.

### DIFF
--- a/lib/Echidna/Transaction.hs
+++ b/lib/Echidna/Transaction.hs
@@ -63,7 +63,7 @@ genTxM m = do
   mm <- getSignatures hmm lmm
   let ns = dictValues genDict
   s' <- rElem ss
-  r' <- rElem $ NE.fromList . catMaybes $ (toContractA mm) <$> toList m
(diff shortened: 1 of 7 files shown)


lib/Echidna/Transaction.hs
```diff
@@ -57,16 +57,16 @@ genTxWith :: (MonadRandom m, MonadState x m, Has World x, MonadThrow m)
           -> (Addr -> ContractA -> m SolCall)         -- ^ Call generator
           -> m Word                                   -- ^ Gas generator
           -> m Word                                   -- ^ Gas price generator
-          -> Word                                     -- ^ Max value generator
+          -> ([FunctionHash] -> Addr -> ContractA -> SolCall -> m Word)   -- ^ Value generator
```
arcz, line 60: did you try to simplify `genTxWith`?

arcz (approved):

PR #543, opened 2020-11-09
Simplify slither info parsing

arcz:
Improved JSON parsing with shorter code and more info stored in SlitherInfo.

### DIFF
--- a/lib/Echidna.hs
+++ b/lib/Echidna.hs
@@ -5,10 +5,9 @@ module Echidna where
 import Control.Lens (view, (^.), to)
 import Data.Has (Has(..))
 import Control.Monad.Catch (MonadCatch(..))
-import Control.Monad.Reader (MonadReader, MonadIO, liftIO, when)
+import Control.Monad.Reader (MonadReader, MonadIO, liftIO)
 import Control.Monad.Random (MonadRandom)
 import Data.Map.Strict (keys)
-import Data.Text (pack)
 
 import EVM (env, contracts, VM)
 import EVM.ABI (AbiValue(AbiAddress))
@@ -18,6 +17,7 @@ import Echidna.Config
 import Echidna.Solidity
 import Echidna.Types.Campaign
 import Echidna.Types.Random
+import Echidna.Types.Signature
 import Echidna.Types.Tx
 import Echidna.Types.World
 import Echidna.Transaction
@@ -39,27 +39,25 @@ import qualified Data.List.NonEmpty as NE
 -- * A list of transaction sequences to initialize the corpus
 prepareContract :: (MonadCatch m, MonadRandom m, MonadReader x m, MonadIO m, MonadFail m,
                     Has TxConf x, Has SolConf x)
-                => EConfig -> NE.NonEmpty FilePath -> Maybe String -> Seed -> m (VM, World, [SolTest], Maybe GenDict, [[Tx]])
+                => EConfig -> NE.NonEmpty FilePath -> Maybe ContractName -> Seed -> m (VM, World, [SolTest], Maybe GenDict, [[Tx]])
 prepareContract cfg fs c g = do
   txs <- liftIO $ loadTxs cd
 
   -- compile and load contracts
   cs <- Echidna.Solidity.contracts fs
   ads <- addresses
-  p <- loadSpecified (pack <$> c) cs
+  p <- loadSpecified c cs
 
   -- run processors
   ca <- view (hasLens . cryticArgs)
   si <- runSlither (NE.head fs) ca
-  when (null si) $ liftIO $ putStrLn "WARNING: slither failed to run or extracted no information at all"
-
-  -- filter extracted constants
-  let extractedConstants = filterConstantValue si
 
   -- load tests
   (v, w, ts) <- prepareForTest p c si
   let ads' = AbiAddress <$> v ^. env . EVM.contracts . to keys
+  let constants' = enhanceConstants si ++ timeConstants ++ largeConstants ++ NE.toList ads ++ ads'
+
   -- start ui and run tests
-  return (v, w, ts, Just $ mkGenDict df (extractedConstants ++ timeConstants ++ largeConstants ++ NE.toList ads ++ ads') [] g (returnTypes cs), txs)
+  return (v, w, ts, Just $ mkGenDict df constants' [] g (returnTypes cs), txs)
   where cd = cfg ^. cConf . corpusDir
         df = cfg ^. cConf . dictFreq
--- a/lib/Echidna/Processor.hs
+++ b/lib/Echidna/Processor.hs
@@ -1,25 +1,25 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module Echidna.Processor where
 
-import Control.Arrow              (second)
-import Control.Lens
-import Control.Monad.IO.Class     (MonadIO(..))
-import Control.Exception          (Exception)
-import Control.Monad.Catch        (MonadThrow(..))
-import Data.Aeson                 (decode, Value(..))
-import Data.Text                  (Text, pack, unpack)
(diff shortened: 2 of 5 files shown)


lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
+enhanceConstants si =
+  nub . concatMap enh . concat . concat . M.elems $ M.elems <$> constantValues si
+  where
+  enh (AbiUInt _ n) = makeNumAbiValues (fromIntegral n)
+  enh (AbiInt _ n) = makeNumAbiValues (fromIntegral n)
+  enh (AbiString s) = makeArrayAbiValues s
+  enh v = [v]
+
+-- we loose info on what constants are in which functions
+data SlitherInfo = SlitherInfo
```
incertia, line 62: if it makes any sense we should also lens this

lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
-
-filterConstantValue :: [SlitherInfo] -> [AbiValue]
-filterConstantValue = slitherFilter _ConstantValue
+filterResults Nothing rs = hashSig <$> (concat . M.elems) rs
+
+enhanceConstants :: SlitherInfo -> [AbiValue]
+enhanceConstants si =
+  nub . concatMap enh . concat . concat . M.elems $ M.elems <$> constantValues si
+  where
+  enh (AbiUInt _ n) = makeNumAbiValues (fromIntegral n)
```
incertia, line 56: keep the first def inline with `where` or indent another 2 spaces

lib/Echidna.hs
```diff
@@ -39,27 +39,25 @@ import qualified Data.List.NonEmpty as NE
...
   si <- runSlither (NE.head fs) ca
-  when (null si) $ liftIO $ putStrLn "WARNING: slither failed to run or extracted no information at all"
-
-  -- filter extracted constants
-  let extractedConstants = filterConstantValue si
 
   -- load tests
   (v, w, ts) <- prepareForTest p c si
   let ads' = AbiAddress <$> v ^. env . EVM.contracts . to keys
+  let allConstants = enhanceConstants si ++ timeConstants ++ largeConstants ++ NE.toList ads ++ ads'
```
incertia, line 58: the `all` seems superfluous here

lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
+        where failure v t = fail $ "failed to parse constant: "++ t ++ " " ++ v
 
 -- Slither processing
-runSlither :: (MonadIO m, MonadThrow m) => FilePath -> [String] -> m [SlitherInfo]
-runSlither fp args = let args' = ["--ignore-compile", "--print", "echidna", "--json", "-"] ++ args in do
+runSlither :: (MonadIO m, MonadThrow m) => FilePath -> [String] -> m SlitherInfo
+runSlither fp extraArgs = do
   mp <- liftIO $ findExecutable "slither"
   case mp of
     Nothing -> throwM $ ProcessorNotFound "slither" "You should install it using 'pip3 install slither-analyzer --user'"
```
incertia, line 126: we should probably just warn here rather than throw

if we want to require slither in the future, that's up for debae

lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
     Just path -> liftIO $ do
-      (ec, out, err) <- readCreateProcessWithExitCode (proc path $ args' |> fp) {std_err = Inherit} ""
+      let args = ["--ignore-compile", "--print", "echidna", "--json", "-"] ++ extraArgs ++ [fp]
+      (ec, out, err) <- readCreateProcessWithExitCode (proc path args) {std_err = Inherit} ""
       case ec of
-        ExitSuccess -> return $ procSlither out
+        ExitSuccess ->
+          case decode (BSL.pack out) of
+            Just si -> pure si
+            Nothing -> throwM $ ProcessorFailure "slither" "decoding slither output failed"
```
incertia, line 134: again let's warn

lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
         ExitFailure _ -> throwM $ ProcessorFailure "slither" err
```
incertia, line 135: warn

lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
+        payableFunctions <- o .: "payable"
+        constantFunctions <- o .: "constant_functions"
+        asserts <- o .: "assert"
+        constantValues'
+          -- the type annotation is needed
+          :: M.HashMap ContractName (M.HashMap FunctionName [[Maybe AbiValue]])
+          <- o .: "constants_used" >>= (traverse . traverse . traverse . traverse) parseConstant
+        -- flatten [[AbiValue]], the array probably shouldn't be nested, fix it in Slither
+        let constantValues = (fmap . fmap) (catMaybes . concat) constantValues'
+        generationGraph <-
```
incertia, line 87: make this more clear with `do`

lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
+          o .: "functions_relations" >>= (traverse . traverse) (withObject "relations" (.: "impacts"))
+        pure SlitherInfo {..}
+
+      parseConstant :: Value -> Parser (Maybe AbiValue)
```
incertia, line 91: we can probably omit this type signature but it's fine to keep

lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
+              Nothing -> failure v t
+              i -> pure i
+
+          'i':'n':'t':x ->
+            case AbiInt <$> readMaybe x <*> readMaybe v of
+              Nothing -> failure v t
+              i -> pure i
+
+          "string" ->
+            pure $ Just $ AbiString $ BSU.fromString v
```
incertia, line 107: `..$`

lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
+            case readMaybe v :: Maybe Int of
+              Just a ->
+                case readMaybe ("0x" ++ showHex a "") of
+                  Nothing -> failure v t
+                  a' -> pure $ AbiAddress <$> a'
+              Nothing -> failure v t
+
+          -- we don't need all the types for now
+          _ -> pure Nothing
+        where failure v t = fail $ "failed to parse constant: "++ t ++ " " ++ v
```
incertia, line 119: space before `++`, and maybe put `t` in parens in the output?

```
failed to parse constant: <value> (<type>)
failed to parse <type>: <value>
```

lib/Echidna/Processor.hs
```diff
@@ -41,155 +42,94 @@ instance Exception ProcException
...
+  , constantFunctions :: M.HashMap ContractName [FunctionName]
+  , asserts :: M.HashMap ContractName [FunctionName]
+  , constantValues  :: M.HashMap ContractName (M.HashMap FunctionName [AbiValue])
+  , generationGraph :: M.HashMap ContractName (M.HashMap FunctionName [FunctionName])
+  } deriving (Show)
+
+instance FromJSON SlitherInfo where
+  parseJSON = withObject "slitherOutput" $ \o -> do
+    -- take the value under 'description' through the path - $['results]['printers'][0]['description']
+    description <- ((o .: "results" >>= (.: "printers")) <&> head) >>= (.: "description")
```
incertia, line 73: let's make this not as confusing

incertia (changes requested):

lib/Echidna/Processor.hs
ggrieco-tob: We discussed this a few weeks ago. We decide to make Slither mandatory for Echidna, since it is used to get the constants from the source code now. If you don't have Slither installed, Echidna won't be able to explore code like this one:

```
function f(uint x) {
  require(x == 424242);
  ...
}
``` 

Forcing the user to use Slither is not so bad, since it is easy to install.

PR #544, opened 2020-11-09
Allow to bypass EIP-170 and set up a custom max code size

### LINKED ISSUE #542: Feature request: Allow to disable EIP-170
It would be useful to be able to get rid of max contract size limit (EIP-170). As Echidna uses hevm, and there’s already an option there to change that limit (`--maxcodesize`), it seems it could be doable without too much effort.

bingen:
Fixes #542


lib/Echidna/Solidity.hs
```diff
@@ -210,7 +211,7 @@ loadSpecified name cs = do
 
   -- Set up initial VM, either with chosen contract or Etheno initialization file
   -- need to use snd to add to ABI dict
-  blank' <- maybe (pure initialVM)
+  blank' <- maybe (pure (initialVM & block . maxCodeSize .~ w256 (fromInteger mcs)))
```
incertia, line 214: maybe we should make `_codeSize` a `Word` instead

ggrieco-tob:
@bingen good work! I'm wondering if it is possible to add a unit test for this feature?

lib/Echidna/Solidity.hs
bingen: What’s the reason? I mimicked the way balance works.

ggrieco-tob: I think we should merge it and then re-organize the types of the maxCodeSize, balance, etc in another PR

ggrieco-tob (approved):

Issue #487, opened 2020-08-12
simplify Haskell code to lower entry barrier


MatthewARinehart:
simplify Haskell code to lower entry barrier

PR #545, opened 2020-11-10
Simplify Echidna.ABI

arcz:
No semantic changes. Replacing MonadReader with an explicit argument and some lens with bare accessors. This effort is towards https://github.com/crytic/echidna/issues/487

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -8,15 +8,12 @@
 
 module Echidna.ABI where
 
-import Control.Lens
+import Control.Lens (makeLenses, (<>~))
 import Control.Monad (join, liftM2, liftM3, foldM, replicateM)
-import Control.Monad.Reader (runReaderT)
-import Control.Monad.Reader.Class (MonadReader)
 import Control.Monad.Random.Strict (MonadRandom, getRandom, getRandoms, getRandomR, uniformMay)
 import Data.Bool (bool)
 import Data.ByteString (ByteString)
 import Data.Foldable (toList)
-import Data.Has (Has(..))
 import Data.Hashable (Hashable(..))
 import Data.HashMap.Strict (HashMap)
 import Data.HashSet (HashSet, fromList, union)
@@ -38,11 +35,11 @@ import qualified Data.ByteString as BS
 import qualified Data.HashMap.Strict as M
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Text as T
-import qualified Data.Text.Encoding  as TE 
+import qualified Data.Text.Encoding as TE
 import qualified Data.Vector as V
 import qualified Data.HashSet as H
 
-import Echidna.Mutator (mutateLL, replaceAt) 
+import Echidna.Mutator (mutateLL, replaceAt)
 import Echidna.Types.Random
 import Echidna.Types.Signature
 
@@ -61,7 +58,7 @@ mkValidAbiUInt i x = if x <= 2 ^ i - 1 then Just $ AbiUInt i x else Nothing
 
 makeNumAbiValues :: Integer -> [AbiValue]
 makeNumAbiValues i = let l f = f <$> commonTypeSizes <*> fmap fromIntegral [i-1..i+1] in
-    catMaybes (l mkValidAbiInt ++ l mkValidAbiUInt) 
+    catMaybes (l mkValidAbiInt ++ l mkValidAbiUInt)
 
 makeArrayAbiValues :: BS.ByteString -> [AbiValue]
 makeArrayAbiValues b = let size = BS.length b in [AbiString b, AbiBytesDynamic b] ++
@@ -96,7 +93,7 @@ encodeSig :: SolSignature -> Text
 encodeSig (n, ts) = n <> "(" <> T.intercalate "," (abiTypeSolidity <$> ts) <> ")"
 
 -- | Get the signature of a solidity method
-hashSig :: Text -> FunctionHash 
+hashSig :: Text -> FunctionHash
 hashSig = abiKeccak . TE.encodeUtf8
 
 -- | Configuration necessary for generating new 'SolCalls'. Don't construct this by hand! Use 'mkConf'.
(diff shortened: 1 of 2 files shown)


lojikil (approved):
lgtm 👍

lib/Echidna/ABI.hs
```diff
@@ -317,15 +313,20 @@ genAbiValueM = genWithDict (fmap toList . view constants) $ \case
...
+                             >>= flip V.replicateM (genAbiValueM genDict t)
+  (AbiArrayType n t)      -> AbiArray n t <$> V.replicateM n (genAbiValueM genDict t)
+  (AbiTupleType v)        -> AbiTuple <$> traverse (genAbiValueM genDict) v
 
 -- | Given a 'SolSignature', generate a random 'SolCalls' with that signature, possibly with a dictionary.
-genAbiCallM :: (MonadReader x m, Has GenDict x, MonadRandom m) => SolSignature -> m SolCall
-genAbiCallM abi = genWithDict (fmap toList . view wholeCalls) (traverse $ traverse genAbiValueM) abi >>= mutateAbiCall
+genAbiCallM :: MonadRandom m => GenDict -> SolSignature -> m SolCall
+genAbiCallM genDict abi =
+  genWithDict
```
incertia, line 323: if we are to make this multiline can we use do

lib/Echidna/ABI.hs
```diff
@@ -159,7 +156,7 @@ getRandomUint n = join $ R.fromList [(getRandomR (0, 1023), 1), (getRandomR (0,
 -- Note that we define the dictionary case ('genAbiValueM') first (below), so recursive types can be
 -- be generated using the same dictionary easily
 genAbiValue :: MonadRandom m => AbiType -> m AbiValue
-genAbiValue = flip runReaderT defaultDict . genAbiValueM
+genAbiValue = genAbiValueM defaultDict
```
incertia, line 159: we could always just give explicit parameters here to avoid `runReaderT`

```haskell
genAbiValue t = genAbiValueM t defaultDict
```

lib/Echidna/ABI.hs
```diff
@@ -126,8 +123,8 @@ gaddCalls c = wholeCalls <>~ hashMapBy (fmap $ fmap abiValueType) c
 defaultDict :: GenDict
 defaultDict = mkGenDict 0 [] [] 0 (const Nothing)
 
-dictValues :: GenDict -> [Integer] 
-dictValues g = catMaybes $ concatMap (\(_,h) -> map fromValue $ H.toList h) $ M.toList $ g ^. constants
+dictValues :: GenDict -> [Integer]
+dictValues g = catMaybes $ concatMap (\(_,h) -> map fromValue $ H.toList h) $ M.toList $ _constants g
```
incertia, line 127: i would try to use lens wherever possible to promote usage and keep things consistent

lib/Echidna/ABI.hs
```diff
@@ -297,16 +294,15 @@ mutateAbiCall = traverse f
...
 -- provided generator or uses the 'GenDict' dictionary (when available).
-genWithDict :: (Eq a, Hashable a, MonadReader x m, Has GenDict x, MonadRandom m)
-            => (GenDict -> HashMap a [b]) -> (a -> m b) -> a -> m b
-genWithDict f g t = do
-  let fromDict = uniformMay . M.lookupDefault [] t . f
-  c <- view hasLens
-  fromMaybe <$> g t <*> (bool (pure Nothing) (fromDict c) . (c ^. pSynthA >=) =<< getRandom)
+genWithDict :: (Eq a, Hashable a, MonadRandom m)
+            => GenDict -> HashMap a [b] -> (a -> m b) -> a -> m b
+genWithDict genDict f g t = do
```
incertia, line 299: we should rename the f parameter here because its meaning has changed and is no longer a function

lib/Echidna/ABI.hs
```diff
@@ -297,16 +294,15 @@ mutateAbiCall = traverse f
...
+  let fromDict = uniformMay (M.lookupDefault [] t f)
+  fromMaybe <$> g t <*> (bool (pure Nothing) fromDict . (_pSynthA genDict >=) =<< getRandom)
```
incertia, line 301: if we want to target #545 this is a good line to target

```haskell
do
  r <- getRandom
  fromMaybe <$> g t <*> bool (pure Nothing) fromDict (genDict ^. pSynthA >= r)
```

lib/Echidna/ABI.hs
```diff
@@ -297,16 +294,15 @@ mutateAbiCall = traverse f
...
 
 -- | Synthesize a random 'AbiValue' given its 'AbiType'. Requires a dictionary.
-genAbiValueM :: (MonadReader x m, Has GenDict x, MonadRandom m) => AbiType -> m AbiValue
-genAbiValueM = genWithDict (fmap toList . view constants) $ \case
+genAbiValueM :: MonadRandom m => GenDict -> AbiType -> m AbiValue
+genAbiValueM genDict = genWithDict genDict (toList <$> _constants genDict) $ \case
```
incertia, line 305: this is why i want to avoid removing MonadReader from things. we now have to pass this around everywhere

lib/Echidna/ABI.hs
```diff
@@ -317,15 +313,20 @@ genAbiValueM = genWithDict (fmap toList . view constants) $ \case
   AbiStringType           -> liftM2 (\n -> AbiString       . BS.pack . take n)
                                     (getRandomR (1, 32)) getRandoms
   (AbiArrayDynamicType t) -> fmap (AbiArrayDynamic t) $ getRandomR (1, 32)
-                             >>= flip V.replicateM (genAbiValueM t)
-  (AbiArrayType n t)      -> AbiArray n t <$> V.replicateM n (genAbiValueM t)
-  (AbiTupleType v)        -> AbiTuple <$> traverse genAbiValueM v
+                             >>= flip V.replicateM (genAbiValueM genDict t)
```
incertia, line 316: like here

PR #548, opened 2020-11-12
provide nicer API for normal transactions

incertia:
quick and easy

### DIFF
--- a/lib/Echidna/Types/Tx.hs
+++ b/lib/Echidna/Types/Tx.hs
@@ -7,7 +7,9 @@ import Prelude hiding (Word)
 import Control.Lens.TH (makePrisms, makeLenses)
 import Data.Aeson.TH (deriveJSON, defaultOptions)
 import Data.ByteString (ByteString)
+import Data.Text (Text)
 import EVM (VMResult(..), Error(..))
+import EVM.ABI (AbiValue)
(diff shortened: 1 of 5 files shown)


lib/Echidna/Types/Tx.hs
```diff
@@ -55,6 +57,18 @@ data Tx = Tx { _call  :: TxCall       -- | Call
 makeLenses ''Tx
 $(deriveJSON defaultOptions ''Tx)
 
+basicTx :: Text -> [AbiValue] -> Addr -> Addr -> Word -> Tx
+basicTx f a s d g = basicTxWithValue f a s d g 0
```
arcz, line 61: could we provide more self-explanatory names? or some docs, or type aliases?

incertia: i guess we could change it to `contractCall` and `create` or something, do you have any suggestions?

arcz: I think `basicTx` is fine. It's the `f`, `a`, `s`, `d`, `g` vars and their types which is hard to decipher and to use the function I need to look at the Tx definition. A doc for `basicTx` describing the args should do the job.

arcz (approved):

PR #550, opened 2020-11-16
Allow echidna to load calls from truffle tests

ggrieco-tob:
This PR allows to use a new config option `testSamples` to specify a .json file with serialized transactions collected during the unit test execution. In order to collect these, [etheno](https://github.com/crytic/etheno) should be used:

```
$ etheno --truffle --ganache --truffle-args test -x test.json
```

Then, echidna will read each transaction, discarding any CREATE, and determinate which calls ones can be replayed with the current target (using the ABI). This will help speed up the testing, since any corner case already included in the unit tests will be replayed as part of the initial corpus.


lib/Echidna/Types/Tx.hs
```diff
@@ -116,3 +116,7 @@ getResult (VMFailure PrecompileFailure)         = ErrorPrecompileFailure
 getResult (VMFailure UnexpectedSymbolicArg)     = ErrorUnexpectedSymbolic
 getResult (VMFailure DeadPath)                  = ErrorDeadPath
 getResult (VMFailure (Choose _))                = ErrorChoose -- not entirely sure what this is
+
+makeSingleTx :: Addr -> Addr -> W256 -> TxCall -> [Tx]
```
arcz, line 120: you can use `Echidna.Types.Signature.SolCall` instead of `TxCall` and there won't be need for `error`

ggrieco-tob:
No longer need it.

PR #552, opened 2020-11-17
Fix crash when reading negative addresses

ggrieco-tob:
This is causing a crash in one of the contract I'm testing right now.


incertia:
`Addr` is just `newtype Addr = Addr Word160` and the `Data.DoubleWord` package has nice `Read` instances.

```
> read (show (2^96 - 1)) :: Word96
79228162514264337593543950335
> read (show (2^96 - 1)) :: Int96
-1
> read (show (2^96 - 2)) :: Word96
79228162514264337593543950334
> read "-1" :: Word96
79228162514264337593543950335
> read "-2" :: Word96
79228162514264337593543950334
```

thus instead of using `readMaybe` on `showHex` to generate a `Maybe Addr`, we can just directly use `Addr <$> readMaybe v`.

incertia (approved):

Issue #555, opened 2020-11-18
Migrating to GitHub Container Registry for Docker images


ggrieco-tob:
https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images

PR #558, opened 2020-11-19
GitHub Actions: switch to GitHub Container Registry

elopez:
This PR switches from GitHub Packages Docker to GitHub Container
Registry. See the following document for more details:

    https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images

This change requires enabling the feature on GitHub by completing the
following instructions:

    https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/enabling-improved-container-support

The feature must be also enabled on the organization level.

Then you must also get a Personal Access Token (PAT) with only
`write:packages` scope (`repo` or any additional ones are not needed),
and store it in the repository settings as a secret with
the name `CR_PAT`, together with a `CR_USERNAME` secret. You can read
the document linked above for more details, and then create the token
by going to:

    https://github.com/settings/tokens

Additionally, this change disables pushes to the trailofbits
Docker Hub repository for non-`crytic` owned repositories,
so pushes are not attempted from echidna forks without proper
credentials.

As a final thing, the Dockerfile and actions were slightly reworked to
enable layer caching and speed up the docker image builds.

Closes: #555

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -15,12 +15,27 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: GitHub Login
-        uses: azure/docker-login@v1
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
         with:
-          login-server: ghcr.io
+          version: latest
+          install: true
+
+      - name: GitHub Login
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
           username: ${{ secrets.CR_USERNAME }}
           password: ${{ secrets.CR_PAT }}
+
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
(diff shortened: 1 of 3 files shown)


elopez:
There's probably room for improvement in the Dockerfile still. Most of the build time is spent building Haskell deps, so maybe we can do something like:
 1. copy just `stack.yaml` & `package.yaml`
 1. `stack build ... --only-dependencies`
 1. copy all of the code
 1.  `stack install ...`

This way layers 1 and 2 get invalidated very infrequently, and most of the time is spent there.

incertia:
you are the docker/github actions machine... thanks for all the work. i'll take a look into all this tomorrow.

incertia:
@ESultanik would you happen to know what's going on with migrating to GHCR internally and are able to turn these settings on for us?

ESultanik:
@incertia I'm not sure. Switching to GHCR would require changes to other dependencies like [`eth-security-toolbox`](https://github.com/crytic/eth-security-toolbox), so we need to tread carefully if we are going to abandon the current deployment method.

ggrieco-tob:
@ESultanik Where we can discuss the `eth-security-toolbox` changes to merge this? Perhaps @elopez can help with that.

PR #562, opened 2020-11-22
Fix issue 505 by updating abiMapping with result from constructor

erivas:
Note: adding a test for this would need Solidity 0.6.x (see #383)


ggrieco-tob:
Is this patch complete? It seems it will be necessary to correct the bytecode in case we load raw transactions here:

https://github.com/crytic/echidna/blob/673aece0382e4da1a5ad1ac880b8a3bc9f231b92/lib/Echidna/RPC.hs#L96-L111

ggrieco-tob:
Superseeded by #593 (but thanks for the PR, this was an interesting idea to consider)

PR #564, opened 2020-11-22
Show the user how the properties failed

ggrieco-tob:
This PR will allow the user to see in the UI the reason why a property failed: either the property returned true/false, reverted or there was another error from the EVM. When running `flags.sol` example, it will show this on the terminal:

```
echidna_sometimesfalse: failed!💥 when property returned false
  Call sequence, shrinking (879/5000):
    set0(0)
    set1(0)
```

This includes a small refactoring of the types and functions from Test module.


ggrieco-tob:
@incertia @arcz what is your opinion on this? I did a little bit of refactoring as well, but perhaps there is a better way to do it.

Issue #464, opened 2020-06-04
gasuse.sol test is flaky


elopez:
When running `stack test` several times, `gasuse.sol` fails randomly.

```
...
    basic/darray-mutation.sol:             OK (3.69s)
    basic/gasuse.sol:                      FAIL (27.83s)
      src/test/Common.hs:63:
      g gas estimate wrong
    coverage/boolean.sol:                  OK (5.88s)
...
```

incertia:
we can always relax the gas estimate a bit, but there is currently no way to specifically target for gas usage and we simply report the max gas that was seen during the campaign.

arcz:
It's been quite stable for some time already.

PR #571, opened 2020-11-30
Run echidna tests in parallel

### LINKED ISSUE #556: MacOS CI workaround should be removed

### LINKED ISSUE #563: Test Echidna with newer version of solc and report results
We should check how echidna works with more recent versions of solc: 0.6.x, 0.7.x and ideally, 0.8.x (which is a preview). This can be easily done running the CI tests using the different versions of solc. We should identify a list of issues to fix (if any).

elopez:
This PR implements parallel execution of the echidna tests on Actions,
which improves testing speed, especially once several solc versions
are added to the test matrix.

The main drawback is that test execution now waits for both Mac and
Linux builds to finish, because they are both part of the same job. As
far as I could see, there is no way to depend on a specific matrix
combination of a job. A possible solution would be to duplicate the
job, and have different ones for Mac and Linux, instead of using
a build matrix.

The testing system could also use an extra feature to specify a range
of solc versions that a test is good for. There are currently some tests
in the suite with hardcoded versions (i.e. with something like
`cryticArgs: ["--solc", "solc-x.y.z"]` in the corresponding .yml).
I saw an attempt at something like that in #383, but I think it should be
something more flexible using e.g. `Data.SemVer`, to allow for fewer
version number hardcoding.

Resolves #563

### DIFF
--- a/.github/scripts/build-macos-release.sh
+++ b/.github/scripts/build-macos-release.sh
@@ -1,4 +1,5 @@
-#!/bin/sh
+#!/bin/bash
+set -eux
 
 add_rpath()
 {
@@ -13,7 +14,7 @@ fix_path()
     NEW="$3"
     OLD=$(otool -L "$BINARY" | grep "$MATCH" | awk '{print $1}')
     install_name_tool -change "$OLD" "$NEW" "$BINARY"
-    cp -n "$OLD" "$(dirname "$BINARY")/$(basename "$NEW")"
+    cp -n "$OLD" "$(dirname "$BINARY")/$(basename "$NEW")" || true
 }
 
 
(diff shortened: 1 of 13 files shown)


ggrieco-tob:
Uhm, it seems we need to fix the fallback test before merging this. @incertia @arcz any thoughts ?

ggrieco-tob:
@elopez can you try merging with #594 to check if the tests are passing?

elopez:
@ggrieco-tob I've rebased on that PR and it looks like it works fine :+1: I noticed on a personal run that `gasuse.sol` failed once; [that test can be a bit flaky](https://github.com/crytic/echidna/issues/464) and it may become more annoying with this patchset now that there's two extra test suite runs.

PR #572, opened 2020-12-01
Improved signature maps

ggrieco-tob:
This could overcome #562, but it is still unclear if it is robust enough.  @incertia @erivas Any thoughts?


ggrieco-tob:
We have a better approach for this.

PR #576, opened 2020-12-04
Optimized testing

### LINKED ISSUE #575: saving and restoring EVM state is not necessary when checking for assertions
https://github.com/crytic/echidna/blob/4af9a3dfe9ac3a0569fed93515688b798492ea51/lib/Echidna/Test.hs#L55-L85

ggrieco-tob:
Fixes #575


incertia:
suggestion:

keep `vm <- get` outside of the `case`. get rid of functor/monad inside of assertion. e.g.
```haskell
      let ret = matchR $ vm ^. hasLens . result
          correctFn = matchC sig $ vm ^. hasLens . state . calldata . _1
```
although `vm` here is a misnomer

incertia (approved):

incertia (approved):

Issue #582, opened 2020-12-16
Shrinking and *wait* bug


agroce:
This contract:

```
pragma solidity 0.4.25;

contract TEST {

  address [] a;

  bool lengthChecking = false;

  function push_1() public {
    a.push(1);
  }

  function turn_on_length_checking() public {
    lengthChecking = true;
  }

  function turn_off_length_checking() public {
    lengthChecking = false;
  }

  function test_long_8() public {
    if (a.length >= 8) {
      if (lengthChecking) {
	assert(false);
      }
    }
  }
}
```

when checked with `checkAsserts: true`

produces weird output like:

```
Analyzing contract: /Users/adg326/echidna-hammer/justlen.sol:TEST
assertion in test_long_8: failed!💥  
  Call sequence:
    push_1() from: 0x0000000000000000000000000000000000010000
    *wait*
    *wait*
    *wait*
    turn_on_length_checking() from: 0x0000000000000000000000000000000000010000
    push_1() from: 0x0000000000000000000000000000000000010000
    *wait*
    *wait*
    *wait*
    test_long_8() from: 0x0000000000000000000000000000000000010000
```

if each `*wait*` were a call to `push_1()` it would make total sense.  Looks like `*wait*` is a synonym for "repeat last step"!

ggrieco-tob:
Please test #584 with a possible fix

agroce:
works for me!

ggrieco-tob:
Fixed.

PR #584, opened 2020-12-24
Fixed wait bug when shrinking

ggrieco-tob:
#582

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,12 +15,12 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          - macos-latest
+            #- macos-latest
         include:
           - os: ubuntu-latest
             apt-get: autoconf automake libtool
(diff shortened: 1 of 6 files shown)


agroce:
(As noted, this fixes the problem for me, with proper shrinking output)

.github/workflows/ci.yml
```diff
@@ -15,12 +15,12 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          - macos-latest
+            #- macos-latest
```
arcz, line 18: is this intended in this PR?

arcz (dismissed):
This looks fine, just the ci.yml to check

.github/workflows/ci.yml
ggrieco-tob: yes, macos building was failing. I think we should disabled it until we can make it work again (or remove it)

incertia (approved):

PR #594, opened 2021-01-26
Semver integration for improve testing with different solc versions

ggrieco-tob:
This PR allows to specify which versions of solc can be used for each test (by default, all the versions are used). This is needed for #593 and #571.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -36,8 +36,10 @@ dependencies:
   - process
   - random
   - rosezipper
+  - semver
   - sbv
   - stm
+  - split
   - temporary
   - text
   - transformers
--- a/src/test/Common.hs
+++ b/src/test/Common.hs
@@ -21,16 +21,15 @@ module Common
 
 import Prelude hiding (lookup)
 
-import Test.Tasty (TestTree, testGroup)
+import Test.Tasty (TestTree)
 import Test.Tasty.HUnit (testCase, assertBool)
 
 import Control.Lens (view, set, (.~), (^.))
-import Control.Monad (when)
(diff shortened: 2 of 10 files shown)


src/test/Common.hs
```diff
@@ -46,6 +52,24 @@ testConfig = defaultConfig & sConf . quiet .~ True
                            & cConf . testLimit .~ 10000
                            & cConf . shrinkLimit .~ 4000
 
+type SolcVersion = Version
+type SolcVersionComp = Version -> Bool
+
+solcV :: Int -> Int -> Int -> SolcVersion
```
arcz, line 58: wouldn't simply tuple with three integers work for our use case?

package.yaml
```diff
@@ -36,8 +36,10 @@ dependencies:
   - process
   - random
   - rosezipper
+  - semver
   - sbv
   - stm
+  - split
```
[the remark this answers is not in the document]
ggrieco-tob: Well, we need splitOn with String, or we could encode string as text.

src/test/Common.hs
ggrieco-tob: I used like that to take the smallest amount of space when writting new tests (`solcV a b c` instead of `solcV (a,b,c)`)

arcz: I mean instead of semver, not a big deal though

incertia: i think tuple makes more sense here

we shouldn't be optimizing for code space

ggrieco-tob: Fair enough. Fixed.

incertia (approved):

PR #604, opened 2021-02-03
Update nix to run test suite without any additional setup

arcz:


incertia (approved):
i'm just going to assume this works

PR #628, opened 2021-03-13
Reworked mutators and tests to be more efficient

ggrieco-tob:
By default, Echidna spent most of the time without re-using transactions from the corpus. This will fix that.


incertia:
i don't know why now is failing but do you think it's possible because of the new changes?

ggrieco-tob:
I tweaked a little the tests.

Issue #634, opened 2021-04-04
UI: improve output for function with a lot of parameters


montyly:
Running Echidna on

```solidity
contract C{
    bool flag = true;

    function test(uint[100] memory v) public{
        flag = false;
    }

    function echidna_test() public returns(bool){
        return flag;
    }
}
```

Leads to

![image](https://user-images.githubusercontent.com/13798342/113517947-538dfa00-9583-11eb-9233-265a07824b72.png)

The parameters are not all shown. 

We should add a carriage return, or change the UI in a way that all the parameters are displayed.

ggrieco-tob:
This is indeed an issue (however, you should still see the complete information when you exit echidna in your console).

montyly:
Yes, you can see the information once it's complete. This is more an issue during a long run if you want to check the results without killing Echidna

PR #635, opened 2021-04-05
Wrap long function calls in UI

arcz:
Fixes https://github.com/crytic/echidna/issues/634.

Screenshot 2021-04-05 at 17 31 11


incertia (approved):
this is a stopgap measure. do we want to do a new pretty-printer that will replace `[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]` lines with `[0, 0, ..., 0]` etc?

PR #643, opened 2021-04-18
Fix numerical range of (u)ints after mutation

ggrieco-tob:


lib/Echidna/ABI.hs
```diff
@@ -176,6 +176,14 @@ genInteractions l = genAbiCall =<< rElem l
 mutateNum :: (Integral a, MonadRandom m) => a -> m a
 mutateNum x = bool (x +) (x -) <$> getRandom <*> (fromIntegral <$> getRandomR (0, toInteger x))
 
+fixAbiUInt :: AbiValue -> AbiValue
+fixAbiUInt (AbiUInt n x) = AbiUInt n (x `mod` ((2 ^ n) - 1))
```
arcz, line 180: This would be simpler and won't require the error case:
```haskell
fixAbiUInt n x = AbiUInt n (x `mod` ((2 ^ n) - 1))
...
0 -> (fixAbiUInt n <$> mutateNum x)
```
The same for `fixAbiInt`.

ggrieco-tob:
Already fixed in #650

PR #659, opened 2021-05-04
Support for loading multiple files with compiled contracts from hardhat/brownie

ggrieco-tob:
This require latest `dev` revision from crytic-compile and slither to work.

### DIFF
--- a/lib/Echidna/Solidity.hs
+++ b/lib/Echidna/Solidity.hs
@@ -114,7 +114,7 @@ readSolcs d = do
   fs <- listDirectory d
   mxs <- mapM (\ f -> readSolc (d ++ "/" ++ f)) fs
   case catMaybes mxs of
-    [] -> error "everythin failed"
+    [] -> return Nothing
     xs -> return $ Just ( unions $ map fst xs, snd $ head xs )
 
 -- | Given a list of files, use its extenstion to check if it is a precompiled
--- a/lib/Echidna/Solidity.hs
+++ b/lib/Echidna/Solidity.hs
@@ -18,14 +18,14 @@ import Control.Monad.State.Strict (execStateT)
 import Data.Foldable              (toList)
 import Data.Has                   (Has(..))
 import Data.List                  (find, partition)
-import Data.Map                   (elems)
+import Data.Map                   (Map, elems, unions)
 import Data.Maybe                 (isJust, isNothing, catMaybes)
 import Data.Text                  (Text, isPrefixOf, isSuffixOf, append)
 import Data.Text.Lens             (unpacked)
 import System.Process             (StdStream(..), readCreateProcessWithExitCode, proc, std_err)
 import System.IO                  (openFile, IOMode(..))
 import System.Exit                (ExitCode(..))
-import System.Directory           (findExecutable)
+import System.Directory           (findExecutable, listDirectory)
 
 import Echidna.ABI                (encodeSig, encodeSigWithName, hashSig, fallback, commonTypeSizes, mkValidAbiInt, mkValidAbiUInt)
 import Echidna.Exec               (execTx, initialVM)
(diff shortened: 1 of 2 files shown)


lib/Echidna/Solidity.hs
```diff
@@ -108,6 +108,15 @@ data SolConf = SolConf { _contractAddr    :: Addr             -- ^ Contract addr
                        }
 makeLenses ''SolConf
 
+readSolcs :: FilePath -> IO (Maybe (Map Text SolcContract, SourceCache))
+
+readSolcs d = do
+  fs <- listDirectory d
+  mxs <- mapM (\ f -> readSolc (d ++ "/" ++ f)) fs
```
incertia, line 115: `\ f` -> `\f`

lib/Echidna/Solidity.hs
```diff
@@ -108,6 +108,15 @@ data SolConf = SolConf { _contractAddr    :: Addr             -- ^ Contract addr
...
+  case catMaybes mxs of
+    [] -> return Nothing
+    xs -> return $ Just ( unions $ map fst xs, snd $ head xs )
```
incertia, line 118: remove space after `(` and before `)`

lib/Echidna/Solidity.hs
```diff
@@ -108,6 +108,15 @@ data SolConf = SolConf { _contractAddr    :: Addr             -- ^ Contract addr
                        }
 makeLenses ''SolConf
 
+readSolcs :: FilePath -> IO (Maybe (Map Text SolcContract, SourceCache))
+
```
incertia, line 112: delete this newline

lib/Echidna/Solidity.hs
```diff
@@ -108,6 +108,15 @@ data SolConf = SolConf { _contractAddr    :: Addr             -- ^ Contract addr
                        }
 makeLenses ''SolConf
 
+readSolcs :: FilePath -> IO (Maybe (Map Text SolcContract, SourceCache))
```
incertia, line 111: probably want to rename this, like `readSolcBatch` or something

incertia (changes requested):

lib/Echidna/Solidity.hs
ggrieco-tob: Done.

lib/Echidna/Solidity.hs
ggrieco-tob: Done.

lib/Echidna/Solidity.hs
ggrieco-tob: Done.

lib/Echidna/Solidity.hs
ggrieco-tob: Done.

TransmissionsDev:
Tested on my CI, works perfectly!

https://github.com/Rari-Capital/nova/runs/2593112827

ZeframLou:
When do y'all plan to release this? I really need this and trying to compile echidna locally has been killing me :(
(on an M1 Mac so probably the issue is on my end)

lib/Echidna/Solidity.hs
```diff
@@ -108,10 +108,36 @@ data SolConf = SolConf { _contractAddr    :: Addr             -- ^ Contract addr
                        }
 makeLenses ''SolConf
 
+-- | List of contract names from every source cache
+type SourceCaches = [([ContractName], SourceCache)]
```
arcz, line 112: wouldn't a Map be better here? or at least non-empty list?

ggrieco-tob: I tried that, but it is only used once, and we need `find` because of the signature that takes a boolean function (believe it or not, there is no equivalent in `Data.Map`. Same with NE lists). This will probably be refactored in later commit.

PR #660, opened 2021-05-04
Update hevm to 0.46.0

arcz:


lib/Echidna/Exec.hs
```diff
@@ -159,14 +158,16 @@ pointCoverage l = do
...
+      buffer <- v ^? env . contracts . at (v ^. state . contract) . _Just . bytecode
+      bc <- viewBuffer buffer
+      pure $ getBytecodeMetadata bc
 
 traceCoverage :: (MonadState x m, Has VM x, Has [Op] x) => m ()
 traceCoverage = do
   v <- use hasLens
-  let c = v ^. state . code
-  hasLens <>= [readOp (BS.index c $ v ^. state . pc) c]
+  let ConcreteBuffer c = v ^. state . code
```
arcz, line 169: @ggrieco-tob @incertia this is an irrefutable pattern, I didn't know what to do with the symbolic buffer here. Let me know if you have a better idea

arcz: this function is not used anywhere actually

ggrieco-tob: Let's remove it.

lib/Echidna/ABI.hs
```diff
@@ -344,3 +346,8 @@ genAbiCallM genDict abi = do
 -- | Given a list of 'SolSignature's, generate a random 'SolCall' for one, possibly with a dictionary.
 genInteractionsM :: MonadRandom m => GenDict -> NE.NonEmpty SolSignature -> m SolCall
 genInteractionsM genDict l = genAbiCallM genDict =<< rElem l
+
+abiCalldata :: Text -> Vector AbiValue -> BS.ByteString
+abiCalldata s xs = BSLazy.toStrict . runPut $ do
```
ggrieco-tob, line 351: Is there any way to avoid calling `toStrict` ?

arcz: this function is copied from `hevm` as it was missing in the newer version, why it is bad? what would you suggest instead?

Issue #595, opened 2021-01-26
Reject ancient solc versions


ggrieco-tob:
We can safely disallow any use of solc below 0.4.25.

snd:
I've been reading the code to see where and how best to add this check.
Before implementing it, I wanted to sketch out what I'm about to do, to verify I'm on the right path.
Let me know if the following sounds good.

Comparing the version and throwing an error for versions before 0.4.25 looks trivial.
It seems that the code generally uses `throwM` to error.
I'd add a new constructor `OldSolcVersion String` to `SolException` and implement `show` for it.
Then I'd `throwM $ OldSolcVersion version` if the version is before 0.4.25.

The remaining piece of the puzzle is getting the version.
It seems to me that information about the `solc` version is currently not required or present within echidna.
The simplest way would be to shell out to `solc --version`.
A more sophisticated way would be to be aware of the `--solc`, `--solc-solcs-select` and `--solc-solcs-bin` which are potentially passed to `crytic-compile` and handle old versions for those arguments.

It would make sense to add the version query and check to `Echidna.Solidity.contracts` since this is where the only interaction with `solc` via `crytic-compile` seems to happen.

ggrieco-tob:
> I've been reading the code to see where and how best to add this check.
Before implementing it, I wanted to sketch out what I'm about to do, to verify I'm on the right path.
Let me know if the following sounds good.

> Comparing the version and throwing an error for versions before 0.4.25 looks trivial.
It seems that the code generally uses throwM to error.
I'd add a new constructor OldSolcVersion String to SolException and implement show for it.
Then I'd throwM $ OldSolcVersion version if the version is before 0.4.25.

This sounds good to me.

> The remaining piece of the puzzle is getting the version.
It seems to me that information about the solc version is currently not required or present within echidna.
The simplest way would be to shell out to solc --version.

There is some code in the `src/test`  for getting the solc version. Please refactor and use that. :smile: 

> A more sophisticated way would be to be aware of the --solc, --solc-solcs-select and --solc-solcs-bin which are potentially passed to crytic-compile and handle old versions for those arguments.

We have integration with crytic-compile, but not with solc-select (yet). Perhaps we should discuss that in a different issue.

> It would make sense to add the version query and check to Echidna.Solidity.contracts since this is where the only interaction with solc via crytic-compile seems to happen.

Sounds good.

PR #661, opened 2021-05-05
Error on solc versions before 0.4.25

snd:
resolves https://github.com/crytic/echidna/issues/595


ggrieco-tob:
If the user is using truffle or any other framework to compile, it is not necessary to have `solc` in the PATH. Is this PR handling that case correctly?

snd:
Good point.
It's not handling that yet.
I'll look into handling it correctly.

snd:
Seems that, when frameworks like truffle are used, then details like the solc version are abstracted away behind crytic-compile.
I've investigated this but found no clean way to get the solc version.
If `crytic-export/combined_solc.json` would contain the solc version this would be easy.
In my experiments it didn't.
@ggrieco-tob Any ideas?

snd:
@montyly just had an idea for a solution: output the solc version used in the slither echidna printer. then check for it. going to implement this.

snd:
superseded by https://github.com/crytic/echidna/pull/675

Issue #625, opened 2021-03-03
Add the config options to the cli 


montyly:
Configuring Echidna is fastidious as the [config doc](https://github.com/crytic/echidna/wiki/Config) is never up to date

I think it would improve usability to have most of the config options accessible through command-line flags

arcz:
I'm working on this slowly

arcz:
@montyly do you think we can close it? there are almost all config options available as cli switches

montyly:
The new helper in echidna 2 helps a lot, however, I think we still have room for improvements. 

For example, we should put the default value and the different options for relevant flags:
- `--test-mode ARG          Test mode to use.`  -> what are the different modes available?
- What are the default value for the test/shrink? What values are we recommending for what usages? 
```
  --test-limit INTEGER     Number of sequences of transactions to generate
                           during testing.
  --shrink-limit INTEGER   Number of tries to attempt to shrink a failing
                           sequence of transactions.
  --seq-len INTEGER        Number of transactions to generate during testing.
```

If we don't want to put all the info within the cli flags (ex: for the recommended value of test/shrink limit), we should link to a place where we have this info

ggrieco-tob:
This should mitigate this issue: https://github.com/crytic/echidna/pull/741. Help will look like this:

```
Available options:
  -h,--help                Show this help text
  --version                Show version
  FILES                    Solidity files to analyze
  --contract CONTRACT      Contract to analyze
  --config CONFIG          Config file (command-line arguments override config
                           options)
  --format FORMAT          Output format. Either 'json', 'text', 'none'. All
                           these disable interactive UI
  --corpus-dir PATH        Directory to save and load corpus and coverage data.
  --test-mode ARG          Test mode to use. Either 'property', 'assertion',
                           'dapptest', 'optimization', 'overflow' or
                           'exploration'
  --multi-abi              Use multi-abi mode of testing.
  --test-limit INTEGER     Number of sequences of transactions to generate
                           during testing. Default is 50000
  --shrink-limit INTEGER   Number of tries to attempt to shrink a failing
                           sequence of transactions. Default is 5000
  --seq-len INTEGER        Number of transactions to generate during testing.
                           Default is 100
  --contract-addr ADDRESS  Address to deploy the contract to test. Default is
                           0x00a329c0648769A73afAc7F9381E08FB43dBEA72
  --deployer ADDRESS       Address of the deployer of the contract to test.
                           Default is 0x0000000000000000000000000000000000030000
[...]

PR #664, opened 2021-05-08
Additional CLI options

arcz:
New CLI options (https://github.com/crytic/echidna/issues/625):

```
  --multi-abi              Use multi-abi mode of testing.
  --test-limit INTEGER     Number of sequences of transactions to generate
                           during testing.
  --shrink-limit INTEGER   Number of tries to attempt to shrink a failing
                           sequence of transactions.
  --seq-len INTEGER        Number of transactions to generate during testing.
  --contract-addr ADDRESS  Address to deploy the contract to test.
  --deployer ADDRESS       Address of the deployer of the contract to test.
  --sender ADDRESS         Addresses to use for the transactions sent during
                           testing. Can be passed multiple times.
  --seed SEED              Run with a specific seed.
  --crytic-args ARGS       Additional arguments to use in crytic-compile for the
                           compilation of the contract to test.
  --solc-args ARGS         Additional arguments to use in solc for the
                           compilation of the contract to test.
```

### DIFF
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -9,6 +9,7 @@ import Control.Monad.Random (getRandom)
 import Data.Maybe (fromMaybe)
 import Data.Text (Text, unpack)
 import Data.Version (showVersion)
+import EVM.Types (Addr)
 import Options.Applicative
 import Paths_echidna (version)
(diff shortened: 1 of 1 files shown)


incertia (approved):
LGTM

PR #674, opened 2021-06-23
RFC: Echidna 2.0

ggrieco-tob:
The next major release of Echidna should allows us to **go beyond property testing for smart contracts and see through their  code when it fails**. In particular, we want a tool that is able to run:

1. Ordinary property tests (e.g. boolean).
2. Assertion tests, either with failed assertions or using events (e.g. `AssertionFailure(...)`) in any contract and even if the execution reverts.
3. If solc 0.8.x or greater is used, automatic detection of [integer overflows, zero division, invalid casts, and others](https://docs.soliditylang.org/en/v0.8.0/control-structures.html?highlight=assert#panic-via-assert-and-error-via-require). 
3. Value optimization tests, where there is a function that computes a value (`int256`) and Echidna tries to find a maximum.  
4. Other useful tests like the detection of contract destruction (which usually should trigger failures in other tests).

If no test is selected, Echidna should run in exploration mode, where coverage feedback allows our tool to reach deep lines of the contracts. 

On top of that, we want Echidna to show us valuable information regarding why the test is failing and what is the state of the contract before and after it failed. That's is why Echidna will show:

1. The cause that triggers a failure in a property or assertion (e.g. a revert in a property test). This allows auditors to quickly detect unexpected failures in properties.
2. The list of events collected during the sequence of transactions that leads to a test to fail (or that provides the maximum value of a certain function if optimization is used). This allows auditors to print any state variable or computation just using events.

This new release also features a simplified interface, using a "test mode" to specify what type of tests you want. For instance, we offer two examples:

1. To test assertions and enables the automatic detection of integer overflows (if solc 0.8.x is used):

```
echidna-test contract.sol --test-mode assertion 
```

2. To run an value optimization and find the max in some polynomial:

```
echidna-test poly.sol --test-mode optimization 
```

This change enables the removal of the usage of `checkAssertion`, `benchmarkMode` or any boolean setting that indicates the testing mode.

The code in this PR is very unstable and it is NOT ready for testing yet.


rmi7:
Here is an idea i had, maybe it makes sense, maybe it doesn't 🤷‍♂️

Instead of having to manually create valid values using a modulo, or returning early if a value is too low or high, it would be nice if you could specify valid ranges for values.

Pretty much all of the important args to functions in smart contracts are (arrays of) integers. So the below only applies to (u)int and arrays of (u)int). 

If you use a modulo to generate valid values then you (1) have to write all the modulo code and (2) echidna will report the original function arg inputs when it finds a violation (instead of the value after the modulo). However, with the proposed changes in this PR you could emit an event to tell you the actual value used, still a bit cumbersome.

If you don't use a modulo but do something like this on the first line of the function `if (arg1 > MY_MAX_ALLOWED_UINT) return;` , then echidna will waste time on returning here (so don't do it). But if it finds a violation the input params it reports are the actual values.

So how about this
```solidity
//@echidna amount: <1-10000000000>, fee: <1,10,100>, distribution: <[33,33,34], [25, 50, 25], [10-20, 10-20, 10-20]>
function deposit(uint amount, uint fee, uint[3] memory distribution, bytes data) public {
  // ...
}
```

- The values of each arg should always be specified inside `<` and `>` , lets call  the value within these the **selector**
- Each item in a **selector** is either an **rvalue** (=range value) or an **array**
- An **array** can only contain **rvalues**
- An **rvalue** is either (1) an integer, or a (2) range B-C where B and C can only be integers (and it's inclusive). 

I think having such syntax would eliminate a lot of code using modulo to get valid values, be more explicit, and would generally simplify using echidna.

Let me know what you think :slightly_smiling_face:

wuestholz:
@ggrieco-tob asked me to expand on my feedback in https://github.com/crytic/echidna/issues/682 here.

Apparently, Echidna currently does not report multiple assertion violations if they are in the same function. This behavior was a bit unexpected for me since most fuzzers will simply accumulate all the crashes until they hit the time limit. I understand that it is important to report violations as soon as they are found. However, that does not necessarily mean that the fuzzer needs to stop. The current design seems to be motivated by the expectation that developers would fix the violation and restart the fuzzer afterwards. However, for audits and long-running campaigns (say 24h or more) it's much more difficult to "fix" the code. It might therefore be useful to support this use case through a separate option.

I had originally tried to use `stopOnFail: false` to disable this behavior, but this option seems to concern the first violation of **the entire campaign**. Maybe one could consider replacing `stopOnFail` with two options `maxFails : <int>` and `maxFailsPerFunction: <int>` to make the behavior more explicit.

naszam:
Hi @ggrieco-tob,
I've put together some feedback for the Echidna 2.0 release:
https://hackmd.io/@naszam/HkRb6Cp0d
Feel free to add comments or ping me for discussion :slightly_smiling_face:

ggrieco-tob:
Great suggestions @naszam !, let's go over the easy ones:

> Debug Mode
> a mode to easily debug failures without the need to check the corpus or in addition to it

Echidna 2.0 already prints the events found when a property failed, so I think that will give a minimal support for that.

> Printers
> in addition to the debug mode, a printer system like slither could help showing the invariants tested and coverage as well as for > gas estimation (if estimateGas opt is set on true)

I think this can be done more easily with a slither printer that can take an Echidna json output. (@montyly)

> CREATE
> add support for create opcode to be used to deploy a contract via bytecode and reduce contract dependencies.

We already [have a workaround of that using Solidity](https://github.com/crytic/building-secure-contracts/blob/master/program-analysis/echidna/testing-bytecode.md).

> hevm.warp()
> add support for a custom function to arbitrarly set block.timestamp in the tests as per dapptools.

This should already work, since we are using hevm, but you will need to know exactly which address/function to call.

> Pip
> allow to install Echidna via pip3 as per slither.

[echidna-parade](https://github.com/crytic/echidna-parade) is already on pip could be updated to include an echidna binary, however, we when need to consider all the security considerations of pushing a precompiled binary from pip.

> Nix
> improve Nix support by adding all dependencies (like slither, solc-select, …).

@arcz is the official nix guru.

hacker-DOM:
**Feature request**

I get a result from an Echidna campaign, e.g. a test failure with a set of event outputs. I hit Esc. Then I want to see it again

It would be great if the results would be copied to the console

ggrieco-tob:
> I get a result from an Echidna campaign, e.g. a test failure with a set of event outputs. I hit Esc. Then I want to see it again
> It would be great if the results would be copied to the console

That's a known issue, basically a feature from Echidna 1.x that I need to re-implement. :smile:

ggrieco-tob:
> **Feature request**
> 
> I get a result from an Echidna campaign, e.g. a test failure with a set of event outputs. I hit Esc. Then I want to see it again
> 
> It would be great if the results would be copied to the console

I already started to work in this reimplementation, so it already shows basic information. Please take a look.

snd:
The event logging feature seems to ignore indexed address types.

For example ERC20 `Transfer` and `Approval` events are output as follows if the `to` and `from` fields are `indexed`:
```
Approval(3104)                                      
Approval(1001059)                                       
Transfer(1001059)                                        
Approval(0)                                     
Approval(1001059)                                        
Transfer(1001059)                               
Approval(0)                             
```
Only the `amount` parameter is output.

It works as expected if the fields are not indexed:

```
Transfer(@0x00a329c0648769A73afAc7F9381E08FB43dBEA72, @0x13d5Bf6416c98E81f667752Ee9591fAF8E98e029, 1000)
```

It would be useful for debugging to get this output for indexed fields as well.

Another related idea:
Could echidna look up the contract name for addresses and annotate addresses with names if they are found?

Like this:
```
Transfer(EchidnaTest@0x00a329c0648769A73afAc7F9381E08FB43dBEA72, MyReceiver@0x13d5Bf6416c98E81f667752Ee9591fAF8E98e029, 1000)
```

This would make the output even more useful.
For ERC20 tokens it would be visible at a glance which contracts are on the sender and receiver side.

ggrieco-tob:
Uhm, interesting. hevm seems to collect that information, but it doesn't implement printing `indexed` parameters! :disappointed: 

https://github.com/dapphub/dapptools/blob/6aad1a62ff2e7cc207683ccb3a4985b64f155806/src/hevm/src/EVM/Format.hs#L193-L197

Perhaps @MrChico knows?

> Could echidna look up the contract name for addresses and annotate addresses with names if they are found?

This sounds useful. I will add it as a TODO card.

MrChico:
> Uhm, interesting. hevm seems to collect that information, but it doesn't implement printing `indexed` parameters! 😞
> 
> https://github.com/dapphub/dapptools/blob/6aad1a62ff2e7cc207683ccb3a4985b64f155806/src/hevm/src/EVM/Format.hs#L193-L197
> 
> Perhaps @MrChico knows?

Yeah, we should implement this. Looks like an oversight on our part.

> 
> > Could echidna look up the contract name for addresses and annotate addresses with names if they are found?
> 
> This sounds useful. I will add it as a TODO card.

Heres our code for that
https://github.com/dapphub/dapptools/blob/master/src/hevm/src/EVM/Format.hs#L82

smsunarto:
For the filter configuration, I think it would be more intuitive if 

instead of
```
filterBlacklist: [true/false]
filterFunctions: [...]
```

we change it to
```
filterType: [WHITELIST/BLACKLIST]
filterFunctions: [...] 
```

snd:
An alternative that might be even more succinct and intuitive:

```
functionsWhitelist: [...]`
```
or
```
functionsBlacklist: [...]
```

specifying both at the same time results in a helpful error message.

hacker-DOM:
It would be fabulous if Echidna also printed event emissions when deployment fails. : )

snd:
Echidna displays a `Panic(1)` before any events even if those events are emitted before the cause of the panic.
See the example output below.
This is inconsistent with the display of `AssertionFailed(...)` which happens correctly after the events.
Caused me some confusion today.

## Example

```solidity
contract Test {
  event AssertionFailed(string message);
  event Debug(string message);

  function test() external {
    emit Debug("one");
    emit Debug("two");
    assert(false);
  }

  function test2() external {
    emit Debug("one");
    emit Debug("two");
    emit AssertionFailed("three");
  }
}
```

when run as

```
echidna-test-2.0 --test-mode assertion --contract Test Example.sol
```

outputs

```
                                          │                                                                                                                      │
                                          │ Call sequence:                                                                                                       │
                                          │ 1.test2()                                                                                                            │
                                          │ Event sequence:                                                                                                      │
                                          │ Debug("one") from: Test@0x00a329c0648769A73afAc7F9381E08FB43dBEA72                                                   │
                                          │ Debug("two") from: Test@0x00a329c0648769A73afAc7F9381E08FB43dBEA72                                                   │
                                          │ AssertionFailed("three") from: Test@0x00a329c0648769A73afAc7F9381E08FB43dBEA72                                       │
                                          │──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────│
                                          │ Assertion failure detector: FAILED! with ErrorRevert                                                                 │
                                          │                                                                                                                      │
                                          │ Call sequence:                                                                                                       │
                                          │ 1.test()                                                                                                             │
                                          │ Event sequence:                                                                                                      │
                                          │ Panic(1)                                                                                                             │
                                          │ Debug("one") from: Test@0x00a329c0648769A73afAc7F9381E08FB43dBEA72                                                   │
                                          │ Debug("two") from: Test@0x00a329c0648769A73afAc7F9381E08FB43dBEA72                                                   │
```

Note that the `Panic(1)` appears before the two events but the events are emitted before it.

snd:
Echidna correctly decodes and displays `0x4e487b710000000000000000000000000000000000000000000000000000000000000001` as `Panic(1)` if it occurs in the main contract. If it occurs in another contract it is not decoded and hard to read.

It would be super helpful if Echidna decoded all panics throughout and also added a little note explaining the error code.

## Example

```solidity
contract Inner {
  function alwaysFails() external {
    assert(false);
  }
}

contract Outer {
  event AssertionFailed(string message);
  event Debug(string message);

  function test() external {
    Inner inner = new Inner();
    bool success = true;
    try inner.alwaysFails() { } catch {
      success = false;
    }
    assert(success);
  }
}
```

when run as

```
echidna-test-2.0 --test-mode assertion --contract Outer Example.sol
```

outputs

```
                                          │ Call sequence:                                                                                                       │
                                          │ 1.test()                                                                                                             │
                                          │ Event sequence:                                                                                                      │
                                          │ Panic(1)                                                                                                             │
                                          │ merror Revert 0x4e487b710000000000000000000000000000000000000000000000000000000000000001 from: Inner  
```

ggrieco-tob:
> Echidna displays a `Panic(1)` before any events even if those events are emitted before the cause of the panic.
> See the example output below.
> This is inconsistent with the display of `AssertionFailed(...)` which happens correctly after the events.
> Caused me some confusion today.

This issue is easy to solve, you can do it in your "improved events" branch and then we can merge it. Just modify this line:

https://github.com/crytic/echidna/blob/dev-test-refactoring/lib/Echidna/Events.hs#L46

ggrieco-tob:
> Echidna correctly decodes and displays `0x4e487b710000000000000000000000000000000000000000000000000000000000000001` as `Panic(1)` if it occurs in the main contract. If it occurs in another contract it is not decoded and hard to read.

I'm not 100% sure how to solve it, but it seems to be produced here:

https://github.com/crytic/echidna/blob/dev-test-refactoring/lib/Echidna/Events.hs#L41-L42

I need to check hevm to see how exactly we should decode reverts from `ErrorTrace`.

hacker-DOM:
> It would be fabulous if Echidna also printed event emissions when deployment fails (constructor execution). : )

While we're on this one we could also add the possibility of having constructors have parameters.

Example use case: I want to do a sequence with a possibility of three different values (for some parameters like oracle price).

montyly:
Regarding `The list of events collected during the transaction that trigger a test failure (or that provides the maximum value of a certain function if optimization is used). This allows users to print any state variable or computation just using events`, Echidna only prints the event in the property, but we could extend this to print all the events of the transactions

naszam:
Hi @ggrieco-tob,
I think would also be nice to have an execution counter along with the `*` when the fuzzing is interrupted, in the coverage text file to have a better understanding of the covered code and overall reachability.

ggrieco-tob:
Superseded by #716

PR #675, opened 2021-06-26
Error on solc versions before 0.4.25

snd:
resolves https://github.com/crytic/echidna/issues/595.

supersedes https://github.com/crytic/echidna/pull/661 (see discussion there for context).

depends on slither release containing https://github.com/crytic/slither/pull/878.
if `solc_versions` is not present in the slither echidna printer output then it is ignored and the version check always succeeds.
this makes this change backwards compatible with previous slither versions.
otherwise the current tests fail because they use a slither version not containing the change in https://github.com/crytic/slither/pull/878.

tested manually. works as expected.

I'm not super happy with the `case` statement with the `Nothing -> return ()` case I added to `lib/Echidna.hs`.
Let me know if there's a neater way to achieve this.
I'm still refreshing my haskell knowledge :wink:


ggrieco-tob:
Can you please add a unit test that checks the solc version that use in tests?

snd:
Happy to add tests.
Could you please elaborate a bit more on which test you'd like me to add?
Do you mean 1) a test that calls `prepareContract` using solc `0.4.24` and asserts that it fails
or 2) a test that calls `prepareContract` using solc `0.4.25` and asserts that it doesn't fail or something else?

PR #678, opened 2021-07-12
Fix source printing and update hevm to 0.47.0

arcz:
Fix to the source printing. the code is also shortened and simplified to handle the coming HTML coverage output.

Before:
```
/Users/artur/projects/echidna/examples/solidity/basic/constants.sol
*r  |contract Constants {
  bool found = false;
  bool found_large = false;

*   |  function find(int i) public {
*   |    if (i == 1447) { found = true; }
*   |    if (i == 133700000000) { found = false; }
*   |    if (i == 11234567890123456789012345678901234560) { found_large = true; }
  }

  function echidna_found() public view returns (bool) {
    return(!found);
  }

  function echidna_found_large() public view returns (bool) {
    return(!found_large);
  }

}

```

After:
```
/Users/artur/projects/echidna/examples/solidity/basic/constants.sol
*r  | contract Constants {
    |   bool found = false;
    |   bool found_large = false;
    |
*   |   function find(int i) public {
*   |     if (i == 1447) { found = true; }
*   |     if (i == 133700000000) { found = false; }
*   |     if (i == 11234567890123456789012345678901234560) { found_large = true; }
    |   }
    |
    |   function echidna_found() public view returns (bool) {
    |     return(!found);
    |   }
    |
    |   function echidna_found_large() public view returns (bool) {
    |     return(!found_large);
    |   }
    |
    | }
    |
```

Additionally, hevm is updated to the latest 0.47.0 along with stack to lts-17.14. The file `nix/crytic-compile.nix` will be removed once slither is fixed in nixpkgs.

### DIFF
--- a/lib/Echidna/Output/Source.hs
+++ b/lib/Echidna/Output/Source.hs
@@ -1,26 +1,27 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TupleSections #-}
 
 module Echidna.Output.Source where
 
 import Control.Lens
 import Data.Maybe (fromMaybe, mapMaybe)
-import Data.Text (Text, unlines, pack, unpack)
+import Data.Text (Text, pack, unpack)
 import Data.Text.Encoding (decodeUtf8)
 import Data.Text.IO (writeFile)
 import Data.Time.Clock.System (getSystemTime, systemSeconds)
 import Data.List (nub)
+import Text.Printf (printf)
 
 import EVM.Solidity (SourceCache, SrcMap, SolcContract, sourceLines, sourceFiles, runtimeCode, runtimeSrcmap, creationSrcmap)
 import EVM.Debug (srcMapCodePos)
-import Prelude hiding (unlines, writeFile)
+import Prelude hiding (writeFile)
 
 import qualified Data.Vector as V
 
 import qualified Data.Map as M
 import qualified Data.Set as S
+import qualified Data.Text as T
 
 import Echidna.Types.Coverage (CoverageMap, CoverageInfo)
 import Echidna.Types.Tx (TxResult(..))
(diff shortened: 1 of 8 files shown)


incertia:
we should fix [this](https://github.com/crytic/echidna/pull/678/checks?check_run_id=3050569882#step:11:30) and then merge

lib/Echidna/Output/Source.hs
```diff
@@ -39,59 +40,26 @@ saveCoverage Nothing  _  _  _ = pure ()
 ppCoveredCode :: SourceCache -> [SolcContract] -> CoverageMap -> Text
 ppCoveredCode sc cs s | s == mempty = "Coverage map is empty"
                       | otherwise   =
-  let allLines = M.toList $ sc ^. sourceLines
+  let allFiles = zipWith (\(srcPath, _) srcLines -> (srcPath, V.map decodeUtf8 srcLines))
```
incertia, line 43: we can prefilter with `fst` so we can avoid the messy lambda. e.g. `over _2 (V.map decodeUtf8) $ zip (fst <$> sc ^. sourceFiles) (sc ^. sourceLines)`.

incertia: maybe even `sc ^.. sourceFiles . traverse . _1`

lib/Echidna/Output/Source.hs
```diff
@@ -101,12 +69,19 @@ getMarker ErrorOutOfGas = 'o'
...
 
 -- | Given a source cache, a coverage map, a contract returns a list of covered lines
-srcMapCov :: SourceCache -> CoverageMap -> SolcContract -> [Maybe (FilePathText, Int, TxResult)]
-srcMapCov sc s c = nub $                                               -- Deduplicate results
-                   map (srcMapCodePosResult sc) $                      -- Get the filename, number of line and tx result
-                   mapMaybe (srcMapForOpLocation c) $                  -- Get the mapped line and tx result
-                   S.toList $ fromMaybe S.empty $                      -- Convert from Set to list
-                   M.lookup (getBytecodeMetadata $ c ^. runtimeCode) s -- Get the coverage information of the current contract
+srcMapCov :: SourceCache -> CoverageMap -> [SolcContract] -> M.Map FilePathText (M.Map Int [TxResult])
+srcMapCov sc s contracts =
```
incertia, line 73: this section is probably clearer using `. . $` instead of `$ $ $`

lib/Echidna/Output/Source.hs
```diff
@@ -101,12 +69,19 @@ getMarker ErrorOutOfGas = 'o'
...
+  M.map (M.fromListWith (++)) $
+  M.fromListWith (++) $
+  map (\(srcPath, line, txResult) -> (srcPath, [(line, [txResult])])) $
+  nub $                                               -- Deduplicate results
+  mapMaybe (srcMapCodePosResult sc) $                 -- Get the filename, number of line and tx result
+  concatMap mapContract contracts
+  where
+    mapContract c =
```
incertia, line 81: `. . $`

src/test/Tests/Seed.hs
```diff
@@ -15,7 +15,7 @@ import Echidna.Mutator.Corpus (defaultMutationConsts)
 seedTests :: TestTree
 seedTests =
   testGroup "Seed reproducibility testing"
-    [ testCase "different seeds" $ assertBool "results are the same" . not =<< same 0 1
+    [ testCase "different seeds" $ assertBool "results are the same" . not =<< same 0 2
```
incertia, line 18: did we do something to the RNG that makes this fail?

stack.yaml
```diff
@@ -1,28 +1,24 @@
...
+- base16-bytestring-1.0.1.0@sha256:33b9d57afa334d06485033e930c6b13fc760baf88fd8f715ae2f9a4b46e19a54,2641
+- restless-git-0.7@sha256:346a5775a586f07ecb291036a8d3016c3484ccdc188b574bcdec0a82c12db293,968
+- s-cargot-0.1.4.0@sha256:61ea1833fbb4c80d93577144870e449d2007d311c34d74252850bb48aa8c31fb,3525
+- sbv-8.15@sha256:bc0ea4e3564626030ac2cb21905b97cd9bd25fddcd6d6010834e8cc2ebf79067,29016
+- semver-range-0.2.8@sha256:44918080c220cf67b6e7c8ad16f01f3cfe1ac69d4f72e528e84d566348bb23c3,1941
+- tree-view-0.5.1@sha256:a32b16fdbe24ad21cb6679c2a925bcc7715ae5c5db0ae6b633e35e28e2fefd98,1184
+- witherable-0.4.1@sha256:5b4840efe20d16cab6b13fa2f39ed383703cccbc246f5b710591069fcc6884c0,2249
+- HSH-2.1.3@sha256:71ded11b224f5066373ce985ec63b10c87129850b33916736dd64fa2bea9ea0a,1705
+- indexed-traversable-instances-0.1@sha256:3aaf97040001bbe583e29c2b9c7d41660df265e6565a0d2ac09a3ed5b8bc21be,2874
+- libBF-0.6.2@sha256:7bffc0e4dbc9bd9e851ba5c29255b62fd2c288dbc9a401cd3761b740f013775e,1770
```
incertia, line 21: do we need the sha hashes?

lib/Echidna/RPC.hs
```diff
@@ -72,8 +72,8 @@ instance FromJSON Etheno where
 
          _ -> M.fail "event should be one of \"AccountCreated\", \"ContractCreated\", or \"FunctionCall\""
     where decode x = case BS16.decode . encodeUtf8 . T.drop 2 $ x of
-                          (a, "") -> pure a
-                          _       -> M.fail $ "could not decode hexstring: " <> T.unpack x
+                          Right a -> pure a
+                          Left e -> M.fail $ "could not decode hexstring: " <> e
```
incertia, line 76: align the arrow here

lib/Echidna/Output/Source.hs
```diff
@@ -39,59 +40,26 @@ saveCoverage Nothing  _  _  _ = pure ()
...
-  m | m3 == m -> m1 :  m2  : m3 : ' ': '|' : l
-  m           -> m1 :  m2  : m3 : m  : '|' : l
-
-markStringLine _ (_: _ : _ : _ : '|':_) = error "impossible to add another marker"
-markStringLine r l = getMarker r : ' ' : ' ' : ' ': '|' : l
+markLines :: V.Vector Text -> M.Map Int [TxResult] -> V.Vector Text
+markLines codeLines resultMap =
+  V.map markLine (V.indexed codeLines)
+  where
+    markLine :: (Int, Text) -> Text
```
incertia, line 59: we likely don't need this sig

lib/Echidna/Output/Source.hs
```diff
@@ -39,59 +40,26 @@ saveCoverage Nothing  _  _  _ = pure ()
...
-  m | m1 == m -> m1 :  m2  : m3 : ' ': '|' : l
-  m | m2 == m -> m1 :  m2  : m3 : ' ': '|' : l
-  m | m3 == m -> m1 :  m2  : m3 : ' ': '|' : l
-  m           -> m1 :  m2  : m3 : m  : '|' : l
-
-markStringLine _ (_: _ : _ : _ : '|':_) = error "impossible to add another marker"
-markStringLine r l = getMarker r : ' ' : ' ' : ' ': '|' : l
+markLines :: V.Vector Text -> M.Map Int [TxResult] -> V.Vector Text
+markLines codeLines resultMap =
+  V.map markLine (V.indexed codeLines)
```
incertia, line 57: move this to previous line

incertia (changes requested):

incertia:
feel free to merge when green once you address the comments

stack.yaml
arcz: this is what stack suggests, I guess it's the a feature

src/test/Tests/Seed.hs
arcz: not really, but for some reason, the results were the same for 0 and 1 in the nix build, I think it is possible to happen?

lib/Echidna/Output/Source.hs
arcz: I decided to leave it as is, the lambda is just one line and I feel it's pretty clear what's going on here.

PR #686, opened 2021-08-15
Print contract names and addresses for events and errors

snd:
I implemented this rapidly because it was needed for an assessment. Still needs to be cleaned up.
Any feedback is greatly appreciated.

## TODO 

- [ ] only print addresses if there is more than one contract with the same name
  - to make output less cluttered
- [ ] replace the `Env` type, that I added quickly to pass both `EConfig` and `DappInfo` via the `MonadReader` to where it was needed, by something more sensible
- [ ] address `TODO`s in code
- [ ] find a way around passing `DappInfo` into the `checkAssertionTest` and `checkSelfDestructedTarget`  in `lib/Echidna/Test.hs`

## Example of the output

```
Call sequence:                                                                                                       │
1.test()                                                                                                             │
Event sequence:                                                                                                      │
Debug("hello world") from: Alice@0xb4c79daB8f259C7Aee6E5b2Aa729821864227e84                                          │
Debug("now it's your turn Bob") from: Alice@0xb4c79daB8f259C7Aee6E5b2Aa729821864227e84                               │
Debug("hello world") from: Bob@0xCe71065D4017F316EC606Fe4422e11eB2c47c246                                            │
AssertionFailed("this was a mistake") from: Bob@0xCe71065D4017F316EC606Fe4422e11eB2c47c246                           
```

contract that produces output above:

```solidity
pragma solidity 0.8.4;

contract Bob {
  event AssertionFailed(string message);
  event Debug(string message);
  function helloWorld() external {
    emit Debug("hello world");
    emit AssertionFailed("this was a mistake");
  }
}

contract Alice {
  event Debug(string message);
  Bob private bob = new Bob();
  function helloWorld() external {
    emit Debug("hello world");
    emit Debug("now it's your turn Bob");
    bob.helloWorld();
  }
}

contract Main {
  Alice private alice = new Alice();
  function test() external {
    alice.helloWorld();
  }
}
```

### DIFF
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -23,6 +23,7 @@ import Data.Maybe (fromMaybe)
 import Data.Text (Text, isPrefixOf)
 import EVM (result)
 import EVM.Types (w256)
+import EVM.Dapp (DappInfo)
 
 import qualified Control.Monad.Fail as M (MonadFail(..))
 import qualified Data.ByteString as BS
@@ -180,3 +181,34 @@ withDefaultConfig = (`runReaderT` defaultConfig)
 -- | 'withDefaultConfig' but not for transformers
 withDefaultConfig' :: Reader EConfig a -> a
 withDefaultConfig' = (`runReader` defaultConfig)
+
+data Env = Env {
+  _cfg :: EConfig,
+  _dapp :: DappInfo
+}
+
+makeLenses ''Env
+
+instance Has EConfig Env where
+  hasLens = cfg
+
+instance Has CampaignConf Env where
+  hasLens = cfg . cConf
+
+instance Has Names Env where
+  hasLens = cfg . nConf
+
+instance Has SolConf Env where
+  hasLens = cfg . sConf
+
+instance Has TestConf Env where
+  hasLens = cfg . tConf
+
+instance Has TxConf Env where
+  hasLens = cfg . xConf
+
+instance Has UIConf Env where
+  hasLens = cfg . uConf
+
+instance Has DappInfo Env where
+  hasLens = dapp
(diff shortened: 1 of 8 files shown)


lib/Echidna/Config.hs
```diff
@@ -180,3 +181,34 @@ withDefaultConfig = (`runReaderT` defaultConfig)
 -- | 'withDefaultConfig' but not for transformers
 withDefaultConfig' :: Reader EConfig a -> a
 withDefaultConfig' = (`runReader` defaultConfig)
+
+data Env = Env {
+  _cfg :: EConfig,
```
ggrieco-tob, line 186: How exactly `cfg` is used from `Env`?

snd: It's used by the `Has` implementations below `data Env ...` https://github.com/crytic/echidna/blob/4c4734ac1bd09a70ac63cf4577f1a329b8fcf675/lib/Echidna/Config.hs#L192-L211

There are a lot of class constraints like `Has CampaignConf x`, `Has TestConf x` throughout the codebase where `x` is the environment of the `MonadReader`. 
`EConfig` satisfies all of these and is currently passed as the reader monad environment here https://github.com/crytic/echidna/blob/16834a15e55221fb81cb37fa5cdf9631a0d50cd9/src/Main.hs#L115

To be able to map codehashes to contract names I needed to pass a mapping via the reader monads environment.
`DappInfo` contains such a mapping. 
This could be retrieved via `hasLens` given a `Has DappInfo x`.

So I needed to implement `Has DappInfo x` for reader monad environment.

I didn't want to clutter `EConfig` with `DappInfo` as `EConfig` is concerned with configuration and `DappInfo` isn't configuration in my opinion.
That's why I added the `Env` type to wrap `EConfig` and `DappInfo`, and to satisfy all required `Has` constraints.

It was very ad hoc.
I don't think it's a bad idea to have a dedicated type for the monad reader environment that is nothing concrete like the config and to which new things which are needed from the environment can be added easily.

PR #687, opened 2021-08-22
WIP: GitHub action

smoelius:


smoelius:
Oh, and the "action" test passes when it points to my fork: https://github.com/crytic/echidna/actions/runs/1156055464

I expect that test will pass if this PR is merged.

naszam:
Hi @smoelius,

I'd like to share some feedback on the pr as discussed with @ggrieco-tob.
- Are there any plans to integrate corpus caching (cc: @ggrieco-tob)?
- Does the echidna GitHub action allow to run multiple echidna tests in parallel via `${{ matrix.testName }}` or by specifying the test file and name in the fields `files` and `contract`? Would be nice to add that case scenario in the tests or docs.
- Checking the CI link https://github.com/crytic/echidna/actions/runs/1156055464, I think it should fail as the echidna test fails.
- My assumption is that the echidna GitHub action will install the last echidna version available on each run and I was wondering if would be possible, eventually, to specify a specific version in case of issues with the last version.
- I really like the multi-abi option, would be great to have more docs, examples on how to leverage that.
- Would be also nice to specify solc version directly as github action field (eg. solc).

Overall, the echidna GitHub action looks good and it's a very nice addition to spread the adoption of echidna in the space and I'm looking forward to start playing with it! :slightly_smiling_face:

smoelius:
Hi @naszam,

> * Are there any plans to integrate corpus caching (cc: @ggrieco-tob)?

I think a feature like this would require some thought. For example, the user might supply a corpus (via `corpus-dir`) that is checked into their repository. If additional corpus files are generated by Echidna, it might make sense for the action to create a PR.

> * Does the echidna GitHub action allow to run multiple echidna tests in parallel via `${{ matrix.testName }}` or by specifying the test file and name in the fields `files` and `contract`? Would be nice to add that case scenario in the tests or docs.

I can't see any reason why that shouldn't work.

> * Checking the CI link https://github.com/crytic/echidna/actions/runs/1156055464, I think it should fail as the echidna test fails.

You have the right idea. You'll note that `negate-exit-status` is set to `1` for each of those tests. I added this feature specifically to test the action. CI should fail if either of those Echidna runs do not produce a crash.

But in a normal run, I agree: CI should fail if the action fails.

> * My assumption is that the echidna GitHub action will install the last echidna version available on each run and I was wondering if would be possible, eventually, to specify a specific version in case of issues with the last version.

This was a good idea and surprisingly [easy to implement](https://github.com/crytic/echidna/pull/687/commits/048d46c450fa0e723ed2aa39b7f144f4aeb9a638), thanks to [solc-select](https://github.com/crytic/solc-select)!

> * I really like the multi-abi option, would be great to have more docs, examples on how to leverage that.

👍 

> * Would be also nice to specify solc version directly as github action field (eg. solc).

Done. 🙂

> Overall, the echidna GitHub action looks good and it's a very nice addition to spread the adoption of echidna in the space and I'm looking forward to start playing with it! 🙂

Thanks very much for the feedback and suggestions!

PR #688, opened 2021-09-07
MacOS Compatibility

ESultanik:
Fixes/updates to the macOS build scripts for better compatibility when building outside of CI.

I don't think they should affect anything, but…
- [ ] Ensure that these changes don't break the macOS build in CI before merging!

### DIFF
--- a/.github/scripts/install-libff.sh
+++ b/.github/scripts/install-libff.sh
@@ -25,7 +25,7 @@ if [ "$HOST_OS" = "macOS" ]; then
   export LDFLAGS=-L$OPENSSL_PREFIX/lib
   export CPPFLAGS=-I$OPENSSL_PREFIX/include
   export CXXFLAGS=-I$OPENSSL_PREFIX/include
-  ARGS="$ARGS -DOPENSSL_INCLUDE_DIR=$OPENSSL_PREFIX/include/openssl -DCURVE=ALT_BN128 -DCMAKE_INSTALL_NAME_DIR=$PREFIX/lib"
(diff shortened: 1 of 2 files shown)


ggrieco-tob:
@incertia can you take a look to this?

.github/scripts/install-libff.sh
```diff
@@ -9,23 +9,28 @@ if [ -f $HOME/.local/lib/libff"$EXT" ]; then
...
 git checkout v0.2.1
 
 ARGS="-DCMAKE_INSTALL_PREFIX=$PREFIX -DWITH_PROCPS=OFF"
 CXXFLAGS=""
 if [ "$HOST_OS" = "macOS" ]; then
-  export LDFLAGS=-L/usr/local/opt/openssl/lib
-  export CPPFLAGS=-I/usr/local/opt/openssl/include
-  export CXXFLAGS=-I/usr/local/opt/openssl/include
-  ARGS="$ARGS -DOPENSSL_INCLUDE_DIR=/usr/local/opt/openssl/include/openssl -DCURVE=ALT_BN128 -DCMAKE_INSTALL_NAME_DIR=$PREFIX/lib"
+  OPENSSL_PREFIX=$(brew --prefix openssl)
```
elopez, line 24: I noticed this returns `/opt/homebrew/opt/openssl@3` on my system, even when I only had `openssl@1.1` installed. From [the GH documentation](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md), the runners appear to have `openssl@1.1`. Do we need to take any extra care here to use a specific version?

elopez (approved):
The changes look sensible, thanks! I only have one question

Issue #692, opened 2021-10-19
Add native M1 support


incertia:
Users are able to use Rosetta binary translation to build on Apple M1 but providing some level of native M1 support may be good once the relevant distributions offer M1 packages.

**This issue has a $5,000 USD bounty thanks to an anonymous contributor**. Exit criteria to satisfy the bounty requires a maintainable, non-hacky, solution. Any changes required to fix dependencies should be PR'd upstream, e.g., to HEVM.

None of the Echidna development team has access to an M1 Mac right now (@incertia, @ggrieco-tob, or @arcz). As a result, we're not certain of the exact level of effort required to close this issue. Some dependencies have only ever been tested on x86-64 and compiling them may not be trivial. Specifically, xbyak is an x86 JIT required by ate-pairing, and finding a solution may require creativity.

Here is a non-exhaustive list of what we know will require changes:

* [ ] Building [HEVM](https://github.com/dapphub/dapptools/tree/master/src/hevm). This is by far the largest dependency echidna has and it depends on some external libraries.
* [ ] Investigate building [ate-pairing](https://github.com/herumi/ate-pairing) and [xbyak](https://github.com/herumi/xbyak) for M1 as they are dependencies of [libff](https://github.com/scipr-lab/libff), a dependency of HEVM (this is the issue the anonymous contributor is blocked on)
* [x] ~~Investigate building [boolector](https://github.com/Boolector/boolector), a dependency of [sbv](https://github.com/LeventErkok/sbv), which is in turn a dependency of HEVM~~ (not required according to [this](https://github.com/crytic/echidna/issues/692#issuecomment-979444449))
* [ ] M1 stack binaries (probably optional, because Rosetta stack can install M1 GHCs)

incertia:
There are a few things that are required to get echidna to build properly. @arcz, @ggrieco-tob, and I are currently not in possession of any M1 hardware so it is not so simple for us to get this to build properly. I will update this comment with more checklist items as they are discovered.

* [ ] Building [HEVM](https://github.com/dapphub/dapptools/tree/master/src/hevm). This is by far the largest dependency echidna has and it depends on some external libraries.
* [ ] Investigate building [ate-pairing](https://github.com/herumi/ate-pairing) and [xbyak](https://github.com/herumi/xbyak) for M1 as they are dependencies of [libff](https://github.com/scipr-lab/libff), a dependency of HEVM (this is the issue the anonymous contributor is blocked on)
[...]

PR #700, opened 2021-12-08
Native M1 development support and releases with Nix

arcz:
This is native M1 support for Echidna using Nix and a working answer to https://github.com/crytic/echidna/issues/692. Some dependencies had to be slightly modified, however, those fixes could be upstreamed to nixpkgs, so in the future, no tweaks will be needed. All tests are passing.

Additionally, I added a way to build a release tarball for non-Nix macOS systems with one command - `nix-build macos-release.nix`.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -18,7 +18,7 @@ jobs:
           - ubuntu-latest
           - macos-latest
     steps:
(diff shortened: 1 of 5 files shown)


ggrieco-tob:
Can you also modify the bash scripts to include PR #688 ? (so the same modifications to enable M1 support)

Issue #694, opened 2021-11-01
Echidna GitHub Action Location


naszam:
Hi,

I'd like to know if there are any plans to move the echidna action into a separate repo (e.g. actions) as recommended [here](https://docs.github.com/en/actions/creating-actions/about-custom-actions#choosing-a-location-for-your-action) for custom actions, to allow release versioning and tracking.

This should also allow to reference the echidna action in the workflow job as follow:
```
steps:
	- uses: crytic/actions/echidna
```
and ease action discovery for the community (it could also be published to the GitHub Marketplace).

Thanks!

ggrieco-tob:
@smoelius what is your view on this?

PR #701, opened 2021-12-13
Move echidna GitHub Action to a separate repository

elopez:
This removes the action from the echidna repository. The action will
now live [on its own repository](https://github.com/crytic/echidna-action) where it can be versioned and published
independently.

Additionally, this updates the workflow to refer to the action on its
new location, as well as adds a few lines to the README pointing to the
new action location.

Closes: #694


smoelius (approved):
LGTM. Thanks again for handling this, @elopez!

PR #705, opened 2021-12-27
Memoize getBytecodeMetadata

samalws:
This pull request makes `pointCoverage` (in `Exec.hs`) memoize its calculation of `getBytecodeMetadata`, instead of recalculating it every time. This results in some fairly significant speedups:
- [This](https://github.com/opynfinance/ConvexityProtocol/blob/dev/contracts/echidna/EchidnaOptionsContract.sol) contract ran 6.7x faster on my computer, producing the same output as before.
- [This](https://github.com/makerdao/dss-vest/blob/master/echidna/ChainLog.sol) contract ran 3.8x faster on my computer, producing the same output as before.

Note that memory usage is slightly increased; in these test cases, memory usage went up by 12% and 15% respectively.
Also note that these speedups only happen when coverage is turned on; the coverage-off speed shouldn't be affected.

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -22,8 +22,8 @@ import Control.Monad.Trans (lift)
 import Control.Monad.Trans.Random.Strict (liftCatch)
 import Data.Binary.Get (runGetOrFail)
 import Data.Bool (bool)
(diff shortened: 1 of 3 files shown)


samalws:
Closed because this PR is a subset of [this PR](https://github.com/crytic/echidna/pull/707).

PR #707, opened 2021-12-30
Memoize getBytecodeMetadata; incorporate usingCoverage2

samalws:
This PR makes `pointCoverage` (in `Exec.hs`) memoize its calculation of `getBytecodeMetadata`, instead of recalculating it every time. In addition, this PR adds a cleaned-up version of [this commit](https://github.com/erivas/echidna/commit/1a4bf8b553aa337c640a2b0c9388515775cbd0e7) by @erivas .

This PR results in some fairly significant speedups:
- [This](https://github.com/opynfinance/ConvexityProtocol/blob/dev/contracts/echidna/EchidnaOptionsContract.sol) contract ran 16x faster on my computer, producing the same output as before.
- [This](https://github.com/makerdao/dss-vest/blob/master/echidna/ChainLog.sol) contract ran 10x faster on my computer, producing the same output as before.

(Note that these speedups only happen when coverage is turned on; the coverage-off speed shouldn't be affected.)

In addition, this PR fixes [this problem](https://github.com/crytic/echidna/issues/322#issuecomment-553521046), where a `for` loop that runs for too long can cause a memory leak.

This PR results in `usingCoverage` and `pointCoverage` being unused functions; I'm not quite sure what should be done about this (whether or not they should be removed, whether `usingCoverage2` should be renamed to just `usingCoverage`).

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -44,7 +44,7 @@ import Echidna.Transaction
 import Echidna.Types.Buffer (viewBuffer)
 import Echidna.Types.Campaign
 import Echidna.Types.Coverage (coveragePoints)
-import Echidna.Types.Signature (ByteStringMap, makeBytecodeMemo)
+import Echidna.Types.Signature (makeBytecodeMemo)
 import Echidna.Types.Test (TestState(..), SolTest)
 import Echidna.Types.Tx (TxCall(..), Tx(..), TxConf, getResult, src, call, _SolCall)
 import Echidna.Types.World (World, eventMap)
@@ -158,11 +158,12 @@ updateGasInfo ((t, _):ts) tseq gi = updateGasInfo ts (t:tseq) gi
 
 -- | Execute a transaction, capturing the PC and codehash of each instruction executed, saving the
 -- transaction if it finds new coverage.
-execTxOptC :: (MonadState x m, Has Campaign x, Has VM x, MonadThrow m) => ByteStringMap -> Tx -> m (VMResult, Int)
-execTxOptC bsm t = do
+execTxOptC :: (MonadState x m, Has Campaign x, Has VM x, MonadThrow m) => Tx -> m (VMResult, Int)
+execTxOptC t = do
   let cov = hasLens . coverage
-  og  <- cov <<.= mempty
-  res <- execTxWith vmExcept (usingCoverage2 bsm cov) t
+  og   <- cov <<.= mempty
+  memo <- use $ hasLens . bcMemo
+  res  <- execTxWith vmExcept (execTxWithCov memo cov) t
   let vmr = getResult $ fst res
   -- Update the coverage map with the proper binary according to the vm result
   cov %= mapWithKey (\_ s -> DS.map (set _4 vmr) s)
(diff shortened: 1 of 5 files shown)


lib/Echidna/Types/Signature.hs
```diff
@@ -30,6 +31,8 @@ type SolCall     = (FunctionName, [AbiValue])
 -- | A contract is just an address with an ABI (for our purposes).
 type ContractA = (Addr, NonEmpty SolSignature)
 
+type ByteStringMap = M.Map ByteString ByteString
```
arcz, line 34: Can we name it `BytecodeMemo` or `BytecodeCache` or something similar?

lib/Echidna/Campaign.hs
```diff
@@ -281,11 +283,12 @@ campaign u v w ts d txs = do
   where
     step        = runUpdate (updateTest w v Nothing) >> lift u >> runCampaign
     runCampaign = use (hasLens . tests . to (fmap snd)) >>= update
+    bsm         = makeBytecodeMemo . mapMaybe (viewBuffer . (^. bytecode)) . elems $ (v ^. env . EVM.contracts)
     update c    = do
       CampaignConf tl sof _ q sl _ _ _ _ _ <- view hasLens
       Campaign { _ncallseqs } <- view hasLens <$> get
       if | sof && any (\case Solved _ -> True; Failed _ -> True; _ -> False) c -> lift u
-         | any (\case Open  n   -> n < tl; _ -> False) c                       -> callseq v w q >> step
+         | any (\case Open  n   -> n < tl; _ -> False) c                       -> callseq bsm v w q >> step
```
arcz, line 291: Maybe this should be put inside the Campaign instead of passing it explicitly? Not sure though

arcz (commented):
We should remove unused functions

ggrieco-tob:
This PR looks fine in general (I'm still doing some testing), however, it needs a little bit of refactoring and adding enough documentation to make sure we can modify it in the future.

lib/Echidna/Exec.hs
```diff
@@ -145,33 +144,19 @@ getExecResult h res og cd g t = do
...
   cm :: CoverageMap <- use l
   let (r, vm', cm') = loop vm cm
   hasLens .= vm'
   l       .= cm'
   return r
   where
-    loop vm cm = maybe (loop (stepVM vm) (addCoverage vm cm)) (, vm, cm) (_result vm)
+    loop vm cm = case _result vm of
+      Nothing  -> loop (stepVM vm) (addCoverage vm cm)
+      (Just r) -> (r, vm, cm)
```
arcz, line 159: parens are not needed around `Just r`, hlint might have problem with it later

lib/Echidna/Exec.hs
```diff
@@ -144,22 +144,28 @@ getExecResult h res og cd g t = do
...
+    loop vm cm = case _result vm of
+      Nothing  -> loop (stepVM vm) (addCoverage vm cm)
+      Just r   -> (r, vm, cm)
+    stepVM = execState exec1
+    addCoverage vm = M.alter
+                       (Just . maybe mempty (S.insert $ currentCovLoc vm))
+                       (currentMeta vm)
+    currentCovLoc vm = (vm ^. state . pc, fromMaybe 0 $ vmOpIx vm, length $ vm ^. frames, Success)
+    currentMeta vm = fromMaybe (error "no contract information on coverage") $ do
+      buffer <- vm ^? env . contracts . at (vm ^. state . contract) . _Just . bytecode
```
ggrieco-tob, line 166: These lines need one comment each one, explaining how they work.

arcz (approved):

PR #709, opened 2022-01-07
Add html coverage output

samalws:
This PR adds an extra coverage output (generated in the corpus directory) that is in html, and has color-coded lines - green indicates covered, red indicates not covered.

This PR also modifies the current txt coverage output so that it includes line numbers.

Here's an example of what the html coverage output looks like:
image


ggrieco-tob:
This looks great. However, we still need to use a different color (e.g. yellow?) if we cover with only reverts certain lines. Can you change it?

ggrieco-tob:
Replaced by #816

Issue #424, opened 2020-04-14
Investigate high-memory consumption with coverage enabled


ggrieco-tob:
[FairWin](https://etherscan.io/address/0x5Ec8515d15C758472f3E1A7B9eCa3e996E8Ba902) contract causes echidna to consume a large amount of memory when coverage is enabled. The amount of memory required seems to be around 10 times more than without coverage.

erivas:
I've started looking at this issue. So far, hacking a bit on `usingCoverage` seems to fix the leak: https://github.com/erivas/echidna/commit/1a4bf8b553aa337c640a2b0c9388515775cbd0e7

ggrieco-tob:
Great! Please let us know if this fix deserves a refactoring on the execution and coverage support.

ggrieco-tob:
This was already improved in several PRs.

PR #725, opened 2022-02-15
Mitigate large memory consumption when replaying corpus

ggrieco-tob:
Possible fix for #424


ggrieco-tob:
This helped a little bit reducing some memory peaks when reusing a corpus.

Issue #729, opened 2022-02-22
Mac OS, Couldn't look up terminfo entry


DanielVF:
I'm on Mac OS 11.4, and have just downloaded the latest 2.0.0 release of echidna. I'm running the new executable in place of my existing echidna 1.x setup. However when I run the new echidna, it errors with:

```
Analyzing contract: [..snip..]
echidna-test: setupTerm: Couldn't look up terminfo entry "xterm-256color"
```

I've also tried:
- running it tmux, which changes the error to not being able to look up "screen".
- running it in different kind of shell (fish vs bash)
- export TERMINFO=/usr/share/terminfo/78/xterm-256color

I'm using the stock Mac OS Terminal program.

Any pointers for debugging this?

ggrieco-tob:
@incertia any idea about this? :point_up:

anishnaik:
For a short term fix append `--format text` to your command and it should work.

Example (with Echidna 2.0):
`$ echidna-test MyToken.sol --test-mode property --format text`

DanielVF:
This does work! Thanks, @anishnaik.

I'm running into an unrelated issue now, but I'll mess with it a bit first before reporting.

elopez:
Hi @DanielVF ! Are you using a M1 Mac by chance? Could you run `export TERMINFO=/usr/share/terminfo` and try running echidna again?

I suspect this is related to the way the M1 binaries are built. If you can help me confirm my suspicions that would be awesome 😄

DanielVF:
Yes, M1 MacBook Pro.

And not only does `export TERMINFO=/usr/share/terminfo` fix the first issue, it also fixes the second issue I was seeing as well!

PR #731, opened 2022-03-02
Fix TERMINFO path for Nix release builds

elopez:
ncurses in Nix is built with a TERMINFO path that references `/nix`.
This causes the binaries fail when ran on non-nix systems, unless
TERMINFO=/usr/share/terminfo is exported. This patches the binary
to use a more sensible default TERMINFO path.

Closes: #729

### DIFF
--- a/macos-release.nix
+++ b/macos-release.nix
@@ -12,6 +12,7 @@ in
       macdylibbundler
       darwin.sigtool
       darwin.cctools
+      perl
     ];
   } ''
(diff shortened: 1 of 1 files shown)


elopez:
@arcz, there must be a nicer way of fixing this 🤔

arcz:
I can't think of a better way at the moment. Maybe it's possible to set this during build-time? Let's merge and fix it later if we figure out an improvement. Thanks!

PR #737, opened 2022-03-16
Partial fix for filtering contract functions in multi-abi

ggrieco-tob:
A small fix to avoid removing misclassifying function as view, when in fact, they are not.


arcz:
I'm afraid this is not the right place to fix it. I made another fix here https://github.com/crytic/echidna/pull/950.

PR #740, opened 2022-03-18
fix readme layout (top bit)

piotrostr:


ggrieco-tob:
We love our logo, but there is no need to have it that big. Still, thanks anyway for your contribution.

PR #750, opened 2022-04-04
Allow to use the FFI cheatcode from HEVM in Echidna

ggrieco-tob:
This PR allows to use the FFI cheatcode from HEVM in Echidna if the `allowFFI: true` is used in the configuration yaml. This is disabled by default.

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -11,7 +11,7 @@ import Control.Monad.Catch (MonadCatch(..), MonadThrow(..))
 import Control.Monad.Random.Strict (MonadRandom, RandT, evalRandT, getRandomR, uniform)
 import Control.Monad.Reader.Class (MonadReader)
 import Control.Monad.Reader (runReaderT, asks)
-import Control.Monad.State.Strict (MonadState(..), StateT(..), evalStateT, execStateT, gets)
+import Control.Monad.State.Strict (MonadState(..), StateT(..), evalStateT, execStateT, gets, MonadIO)
 import Control.Monad.Trans (lift)
 import Control.Monad.Trans.Random.Strict (liftCatch)
 import Data.Binary.Get (runGetOrFail)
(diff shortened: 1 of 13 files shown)


patrickd-:
```
testMode: assertion
seqLen: 1
allowFFI: true
initialize: echidna.json
```

After adding an initialization file it's no longer stuck at startup but then the FFI call apparently reverts and the assertion that is supposed to fail does not.

ggrieco-tob:
You don't need to add echidna initiation here. However, hevm seems to be crashing for some reason:

```
echidna-test: VM failed for unhandled reason, Query <EVM.Query: do ffi: ["echo","-n","0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000002676d000000000000000000000000000000000000000000000000000000000000"]. This shouldn't happen. Please file a ticket with this error message and steps to reproduce!
```

I'm investigating this issue.

patrickd-:
Thanks!

I'd like to use the echidna `initialize` contract to deploy some contracts with incompatible solidity version and then fuzz on a contract that also makes use of the FFI cheatcode, that's why I'm asking about initiation

ggrieco-tob:
Uhm, I see. I will keep working on ths PR to include that feature. Btw, it seems that FFI is implemented using oracles in HEVM:

https://github.com/dapphub/dapptools/blob/74a54da1dc77fe71eab5d8b93ed3f63904f151a1/src/hevm/src/EVM/Fetch.hs#L171-L179

This can take a little longer to implement, but it still should be doable soon.

patrickd-:
I don't know anything about Haskell but let me know if I can help somehow

ggrieco-tob:
@patrickd- can you re-try using the latest commit in this branch? It should work with the example (but the .json initialization is still need to be fixed before you can use it)

patrickd-:
The only difference to foundry that I noticed is that in Echidna the stdout hex always needs to be prefixed with `0x` while foundry is fine with omitting that.

But that's not much of an issue.. besides that everything seems to be working as expected!

ggrieco-tob:
Just push the code to allow ffi when using a json initialization. However, the code is very preliminary, so consider this like an a beta code that will be improved in the future before merging it.

patrickd-:
Awesome! Everything seems to work fine at least. I'll let you know if I notice any issues

PR #751, opened 2022-04-07
Enable the setUp() function for every type of test

ggrieco-tob:
This small PR is testing to enable the `setUp()` for every type of test (not only dapptools). It will be always executed, if available if it will still be part of the ABI so it can be re-executed at any time (therefore, it must be idempotent).


yuchangyuan:
Is setUp() function a good place to call "initialize" for openzeppelin upgradeable contracts?

PR #752, opened 2022-04-08
Allow Echidna to collect inputs even after a revert

ggrieco-tob:
An experiment after a discussion with @agroce.


sambacha:
Is this feature still being considered? Would very much like to have this ability in the release offering.

PR #758, opened 2022-04-22
Support for deployment of certain contracts in specific addresses

ggrieco-tob:
This PR adds support for deployment of certain contracts in specific address. A simple example:

```solidity
contract A {
  uint256 public x = 128;
}

contract Test {
  function test(uint256 y) public {
    assert(A(address(0x42)).x() == y);
  }
}
```

with the following config file:

```yaml
deploy: [["0x42", "A"]]
```

To test:

```bash
$ echidna-test deploy.sol --test-mode assertion --config deploy.yaml --contract Test
```


hellwolf:
thank you! do you recommend me to build echidna-test from this branch? Is it cabal based?

ggrieco-tob:
It is using stack (which will somehow use cabal). You can grab the binaries directly from our CI if you want to quickly test it.

ggrieco-tob:
@patrickd- could this PR help you to simplify your use of Echidna?

patrickd-:
hmm, no, haven't had the need for this one yethmm, no, haven't had the need for this one yet

hellwolf:
tried the binary, it works!!

Branch: https://github.com/superfluid-finance/protocol-monorepo/pull/708

How to run (I downloaded the CI build to ~/tmp):
```
$ ECHIDNA=~/tmp/echidna-test ./fuzz.sh CFAFuzzer
```

What I did, in config:

```
deploy: [["0x1820a4B7618BdE71Dce8cdc73aAB6C95905faD24", "ERC1820RegistryMock"]]
```

Caveat though:

Usually the ERC1820Registry is distributed as an exact binary, e.g.:  https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/scripts/artifacts/ERC1820Registry.json But I guess the syntax proposed here wouldn't be able to use a raw binary like this.

Great stuff anyhow! Thank you!

ggrieco-tob:
We can definitely allow to use a precompiled contract, instead of a contract name. It will also be useful for differential fuzzing

lib/Echidna/Solidity.hs
```diff
@@ -223,7 +223,7 @@ loadSpecified name cs = do
       -- main contract deployment
       let deployment = execTx $ createTxWithValue bc d ca (fromInteger unlimitedGasPerBlock) (w256 $ fromInteger balc) (0, 0)
       vm3 <- execStateT deployment vm2
-      when (isNothing $ currentContract vm2) (throwM $ DeploymentFailed ca)
+      when (isNothing $ currentContract vm3) (throwM $ DeploymentFailed ca)
```
hellwolf, line 226: aha, we need linear type!! :D

hellwolf:
Happy with this so far, this has been integrated in our fuzzing testsuite.

Any plan to merge this soon and release?

ggrieco-tob:
We need to add some tests here, but if you don't have more suggestions, it can be merged as soon as next week

Issue #759, opened 2022-04-25
Fuzzings hangs on second .call


EVLRBT-01:
Hey! 

New to Solidity and have managed to implement a test suite with Forge along with Slither analysis. I am now trying out Echidna and seem to be facing a scenario which hangs the fuzzing. 

I have a withdraw function on a contract which uses OpenZeppelin Address to send a split payment, but the second `Address.sendValue()` seems to cause the fuzzing to hang. I also tried using just `payable(address(1).call{value: amount}("");` from the `Address.sendValue()` implementation (i.e. without the `require`s) but the same occurs - unless I comment out the second line. 

A simple example:
```
pragma solidity 0.8.9;

import "@openzeppelin/contracts/utils/Address.sol";

contract Test {

    function withdraw() external {
        uint256 balance = address(this).balance;
        Address.sendValue(payable(address(1)), balance * 500/1000);
        Address.sendValue(payable(address(2)), balance * 500/1000);
    }

    function echidna_Test() public view returns(bool){
         return true;
    }
}
```

Any ideas/workarounds?

ggrieco-tob:
Hi,

A few questions for you:

* Which version of Echidna are you using? 
* What do you mean with 'seems to cause the fuzzing to hang'?
* Have you configured your fuzzing campaign to send ether to the `Test` contract? I believe the balance should be always 0.

Finally: 0x1 and 0x2 are precompile contracts, which are not usually a good idea to send balance. Can you retry with other addresses instead (e.g. 0x42).

EVLRBT-01:
Thanks for your response!

- Latest version: Echidna 2.0.1
- It shows the dashboard but it remains 'blank' and doesn't advance through the fuzzing, like it does when I comment out the second transfer.
With both transfers:
![image](https://user-images.githubusercontent.com/60948618/165257481-84871194-af94-43c4-9b92-7026287a0470.png)
Second transfer commented out:
![image](https://user-images.githubusercontent.com/60948618/165257149-f1dab26e-fd78-4dc8-9c42-e9c61c7457ba.png)
- I have now added `balanceContract: 100000000000000000000 # 100 eth` to the config file and running with `echidna-test . --contract Test --config echidna.config.yml` has the same result (unless I comment out the second transfer).
 
I have tried with new contract addresses as suggested (see below) but same result. I added the payable constructor to get around https://github.com/crytic/echidna/issues/446 when using balanceContract.
```
pragma solidity 0.8.9;

[...]

PR #760, opened 2022-04-27
Handle multiple hevm queries in a single transaction

ggrieco-tob:
A fix for #759, which allows to solve this tests without crashes or invalid results:

```solidity
contract Test {
    bool state1 = true;
    bool state2 = true;
    function withdraw(uint amount, bool positive) external {
        if (positive)
          require(amount > 0);

        bool success; 
        (success,) = payable(address(0x42)).call{value: amount}("");
        require(success);
        (success,) = payable(address(0x43)).call{value: amount}("");
        require(success);

        if (!positive)
          state1 = false;
        else 
          state2 = false;
    }

    function echidna_should_fail() public view returns(bool){
         return state1;
    }

    function echidna_should_never_fail() public view returns(bool){
         return state2;
    }
}
```

### DIFF
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -14,7 +14,7 @@ import Data.Has (Has(..))
 import Data.Maybe (fromMaybe)
 import EVM
 import EVM.Exec (exec, vmForEthrunCreation)
-import EVM.Types (Buffer(..))
+import EVM.Types (Buffer(..), Word)
 import EVM.Symbolic (litWord)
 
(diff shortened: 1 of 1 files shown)


arcz (approved):
I think this looks fine Haskell-wise, I would maybe just reduce the number of ' and go from `vmResult'` to `vmResult` and from `vmResult''` to `vmResult'`. I know it's been like that before but I'm not sure if this is necessary.

ggrieco-tob:
We need to add a test for this (the example here should work fine)

PR #799, opened 2022-07-13
Correctly decode non-utf8 strings from slither printer

ggrieco-tob:
Tests are missing, but this should work.


Boyan-MILANOV:
Tested locally and it works. Echidna successfully solves the following test:
````
contract CoverageString {
    bytes16 key = "\xf8";
    function test(string memory input) public {
       bytes memory x = bytes(input);
       if (x.length > 0)
         assert(key[0] != x[0]);
    }
}
````

PR #807, opened 2022-07-28
Allow to generate arrays/strings sizes up 150 elements

ggrieco-tob:


ggrieco-tob:
This requires a better approach

PR #816, opened 2022-08-19
Colored html output based on @samalws code

ggrieco-tob:
Replaces #709

### DIFF
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -134,7 +134,8 @@ main = do
   saveTxs cd (snd <$> DS.toList (cpg ^. corpus))
 
   -- save source coverage
-  saveCoverage cd sc cs (cpg ^. coverage)
+  saveCoverage False cd sc cs (cpg ^. coverage)
+  saveCoverage True  cd sc cs (cpg ^. coverage)
 
   if not . isSuccess $ cpg then exitWith $ ExitFailure 1 else exitSuccess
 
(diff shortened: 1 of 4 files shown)


arcz:
The last commit fixes coverage display for non-runtime lines.

| Before | After |
|--------|-------|
| before | after |

disconnect3d:
If we are on this, what kind of information are we gathering here? Is it only whether we hit a particular line or is it the number of hits?

Or: can we display the number of hits since this may also be useful?

When fuzzing a complex target and ending up with e.g. 100k corpus inputs, I like to generate coverage report and look for both A) paths that were not visited at all and B) paths that were visited only few times.

So for C/C++/Rust programs I could see sth like:

```
3.5k |  if (...) {
5    | ...
     | }
```
Which tells me that maybe I should also be interested in getting into this condition more often than just 5 times (and then I can e.g. hardcode the harness code to always get there and fuzz again etc).

arcz:
@disconnect3d This is something worth implementing after this PR is merged. I would need to look if we gather the number of hits per line.

ggrieco-tob:
I think @disconnect3d suggestion makes sense, however, for smart contracts is not so easy. We will need several numbers to show how many hits we had per result (e.g. reverted or not). We need to experiment with this too see which is the most useful output here. I agree that we should merge this and then implement more features in future PRs

src/Main.hs
```diff
@@ -133,9 +134,12 @@ main = do
   -- save corpus
   saveTxs cd (snd <$> DS.toList (cpg ^. corpus))
 
+  -- get current time to save coverage
+  t <- getSystemTime
+  let s = fromIntegral $ systemSeconds t
```
arcz, line 139: should `s` be `systemSeconds`?

PR #823, opened 2022-09-07
Initialize sender addresses from etheno transactions

ggrieco-tob:


lib/Echidna/RPC.hs
```diff
@@ -116,6 +117,26 @@ loadEthenoBatch fp = do
       let initVM = mapM execEthenoTxs ethenoInit
       execStateT initVM initialVM
 
+initAddress :: MonadState VM m => Addr -> m ()
+initAddress addr = do
+  cs <- use (env . EVM.contracts)
+  if addr `member` cs then return ()
```
arcz, line 123: we tend to use `pure` instead of `return` in the new code

lib/Echidna/RPC.hs
```diff
@@ -116,6 +117,26 @@ loadEthenoBatch fp = do
...
+  else env . EVM.contracts . at addr .= Just account
+ where account = initialContract (RuntimeCode mempty) & set nonce 0 & set balance (w256 100000000000000000000) -- default balance for EOAs in etheno
+
+showQueryError :: (MonadState VM m, MonadFail m, MonadThrow m) => Query -> Etheno -> m ()
+showQueryError q et =
```
arcz, line 128: hmm, should this be named `crashWithQueryError` instead, for example?

arcz (approved):

Issue #815, opened 2022-08-11
Rename echidna-test to echidna


montyly:
I don't think we have any reason to keep `echidna-test` instead of `echidna`. As far as I can remember we used `echidna-test`, because in its first iteration we had ideas for other tools. As we never explored them, it would probably make sense to just use `echidna` for simplicity

PR #826, opened 2022-09-19
Change executable name from echidna-test to echidna

arcz:
Fixes https://github.com/crytic/echidna/issues/815

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,11 +32,11 @@ RUN /venv/bin/pip3 install --no-cache slither-analyzer solc-select "crytic-compi
 
 
 FROM gcr.io/distroless/python3-debian11:nonroot AS final-distroless
-COPY --from=builder-echidna /root/.local/bin/echidna-test /usr/local/bin/echidna-test
+COPY --from=builder-echidna /root/.local/bin/echidna /usr/local/bin/echidna
 COPY --from=builder-python3 /venv /venv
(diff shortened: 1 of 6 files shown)


ggrieco-tob:
This will break a number of CI tests (including echidna-action). Let's wait for a "major" release to implement it.

ggrieco-tob:
@elopez can you take a quick look to see how this will impact github actions?

docker/Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN apt-get update && \
         && \
     rm -rf /var/lib/apt/lists/*
 RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
-COPY --from=builder-echidna /root/.local/bin/echidna-test /usr/local/bin/echidna-test
+COPY --from=builder-echidna /root/.local/bin/echidna /usr/local/bin/echidna
```
elopez, line 52: I think it will break, as the action invokes `echidna-test` in the docker container. Maybe we can provide a symlink here to maintain compatibility
```suggestion
COPY --from=builder-echidna /root/.local/bin/echidna /usr/local/bin/echidna
RUN ln -s /usr/local/bin/echidna /usr/local/bin/echidna-test
```

elopez: (alternatively we can update the action to check both binary names and choose the one present)

PR #830, opened 2022-09-22
Use Show AbiValue from hevm instead of ppAbiValue

arcz:
`ppAbiValue` is already implemented as `Show` instance in `hevm` 

https://github.com/dapphub/dapptools/blob/hevm/0.49.0/src/hevm/src/EVM/ABI.hs#L105-L119

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -19,7 +19,6 @@ import Data.Foldable (toList)
 import Data.Hashable (Hashable(..))
 import Data.HashMap.Strict (HashMap)
 import Data.HashSet (HashSet, fromList, union)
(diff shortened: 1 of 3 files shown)


ggrieco-tob:
This looks good. I'm just wondering if addresses are printed using the same approach, as the code looks different?

ggrieco-tob:
As I suspected, this code will print address differently: from `f(0x0)` to `f(0x0000000000000000000000000000000000000000)`, which is not going to help users. I recommend not to merge this code. :disappointed:

arcz:
I agree, this is going to change the looks of our UI. I'll close this for now as we'll be updating hevm and it might change even more.

Issue #831, opened 2022-09-23
Long lines are cut off in the UI


arcz:
Wrap long lines so the whole text is readable in the UI:
![Screenshot from 2022-09-23 11-31-08](https://user-images.githubusercontent.com/4679721/191958524-ff38eaa6-14d9-4556-a7cb-00bb6132e8fd.png)

ggrieco-tob:
This is an issue, but difficult to solve, since the user can resize the window and it is non-trivial to know how cut "long" values such as number or strings

arcz:
I think this could be solved with brick's functions such as https://hackage.haskell.org/package/brick-1.1/docs/Brick-Widgets-Core.html#v:txtWrap

PR #832, opened 2022-09-23
Wrap long tx and event lines in UI

arcz:
Closes https://github.com/crytic/echidna/issues/831.
| Before | After |
|--------|------|
|Screenshot from 2022-09-23 14-54-40|Screenshot from 2022-09-23 14-52-44|


nickdammy (commented):
Importing Text.Wrap and apply wrap to the string resolved the issu.

PR #847, opened 2022-11-15
Fix a small typo in README.md

webthethird:
Fixed very small typo in post-installation test run instructions. A stray `/` before `src/tests/solidity/basic/flags.sol` was causing a file not found error.


elopez:
I think it is a valid workaround for the time being, but we should fix crytic/crytic-compile#310

arcz:
Looks like it is fixed in crytic-compile.

PR #858, opened 2022-11-26
Save and load reproducers for every test

ggrieco-tob:


rappie:
Reminder: don't write empty reproducer with only `[]` inside

PR #884, opened 2022-12-31
Update hevm to 0.50.1

arcz:
This also updates dependencies and GHC to 9.2


lib/Echidna/ABI.hs
```diff
@@ -107,34 +109,23 @@ data GenDict = GenDict { _pSynthA    :: Float
                          -- ^ Default seed to use if one is not provided in EConfig
                        , _rTypes     :: Text -> Maybe AbiType
                          -- ^ Return types of any methods we scrape return values from
+                       , _dictValues :: Set W256
```
ggrieco-tob, line 112: Can you document this?

arcz: To add more context, I profiled tests and computing dictValues was a major bottleneck, it's worth having them cached as a Set where we can take random elements quickly

ggrieco-tob: Just add a line before with a comment like the other ones, and this is ready to go.

PR #892, opened 2023-01-05
Experimental power number generator for uints

ggrieco-tob:
This should be considered experimental until we know it works better


rappie:
I have run some tests. It does not seem to be working.

I would expect echidna to be able to break `testAmount` without the helper. It needs a number between `1e16` and `1e20`, which should occur frequently now in the distribution of the random numbers.

```solidity
pragma solidity ^0.8.0;

contract EchidnaTest {
    uint256 private constant min = 16;
    uint256 private constant max = 20;

    function helper(uint256 amount) public {
        amount = (10**min) + (amount % ((10**max) - (10**min)));
        testAmount(amount);
    }

    function testAmount(uint256 amount) public {
        if (amount > (10**min) && amount < (10**max)) {
            assert(false);
        }
    }
}
```

```yaml
testMode: assertion
testLimit: 100000
```

ggrieco-tob:
This is working now, thanks @rappie !

rappie:
I have tested this.

As expected, it is noticably slower to break certain properties than before, but it does manage to break them without helpers.

Personally I'd prefer to have only the logarithmic RNG. I believe this to be superior for fuzzing, but then again I'm no expert :)

lib/Echidna/ABI.hs
```diff
@@ -163,12 +163,19 @@ mkDictValues =
...
 getRandomUint n =
-  getRandomR =<< Random.weighted
-    [ ((0, 1023), 1)
-    , ((0, 2 ^ n - 5), 8)
-    , ((2 ^ n - 5, 2 ^ n - 1), 1)
+  join $ Random.weighted
+    [ (getRandomR (0, 1023), 1)
+    , (getRandomR (0, 2 ^ n - 5), 8)
+    , (getRandomR (2 ^ n - 5, 2 ^ n - 1), 1)
+    , (getRandomPow (n - 5), 4)
```
arcz, line 178: as I understand the chance of this happening will be 4/14 ~ 28%?

arcz (approved):

PR #900, opened 2023-01-10
Use `bit` instead of power of two (2 ^ i)

elopez:
We are computing powers of two, so using `bit` performs better.


ggrieco-tob:
This is great, but can you keep replacing the usage of `2 ^ i` in the same file ? :slightly_smiling_face:

PR #903, opened 2023-01-11
Refactor and simplify code to not use Data.Has

arcz:
We've been using `Data.Has` extensively and it's been a significant pattern in our code. While it allows for more abstract code to be written, it's not a commonly used library and it comes at a cost of substantial mental load as it was often hard to understand what the code is doing.

This PR removes all `Data.Has` from our code and replaces it with more common patterns, in the hope to improve the project's maintenance and development.

I also enabled the `OverloadedRecordDot` extension that comes with GHC 9.2 that somewhat fixes the records so they can be accessed like in "normal" languages 😄.

### DIFF
--- a/lib/Echidna/Mutator/Corpus.hs
+++ b/lib/Echidna/Mutator/Corpus.hs
@@ -5,10 +5,10 @@ import Data.Set qualified as DS
 
 import Echidna.Mutator.Array
 import Echidna.Transaction (mutateTx, shrinkTx)
+import Echidna.Types (MutationConsts)
 import Echidna.Types.Tx (Tx)
 import Echidna.Types.Corpus
 
-type MutationConsts a = (a, a, a, a)
 defaultMutationConsts :: Num a => MutationConsts a
 defaultMutationConsts = (1, 1, 1, 1)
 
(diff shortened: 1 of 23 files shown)


lib/Echidna/Mutator/Corpus.hs
```diff
@@ -40,7 +40,7 @@ mutator Deletion = deleteRandList
 selectAndMutate :: MonadRandom m
                 => ([Tx] -> m [Tx]) -> Corpus -> m [Tx]
 selectAndMutate f ctxs = do
-  rtxs <- weighted $ map (\(i, txs) -> (txs, fromInteger i)) $ DS.toDescList ctxs
+  rtxs <- weighted $ map (\(i, txs) -> (txs, fromIntegral i)) $ DS.toDescList ctxs
```
ggrieco-tob, line 43: Why this was changed?

lib/Echidna/Transaction.hs
```diff
@@ -119,25 +117,23 @@ mutateTx t@(Tx (SolCall c) _ _ _ _ _ _) = do f <- oftenUsually skip mutate
...
-liftSH = stateST . runState . zoom hasLens
-  -- This is the default state function written in terms of get and set:
-  where stateST f = do
-          s <- get
-          let ~(a, s') = f s
-          put s'
-          return a
+
+-- | Transform an EVM action from HEVM to our MonadState VM
+fromEVM :: MonadState VM m => EVM a -> m a
```
ggrieco-tob, line 122: Should we move this to a different file?

arcz: what about moving it to `Echidna.Types`?

lib/Echidna/Mutator/Corpus.hs
arcz: I changed `Integer` to `Int` in the Corpus type `type Corpus = Set (Int, [Tx])` as we don't need an unbounded Integer there.

ggrieco-tob: Great

lib/Echidna/Transaction.hs
ggrieco-tob: Sounds good.

PR #906, opened 2023-01-13
Code cleanup

arcz:
This PR features various cleanup and simplifications in the entirety of code:
- Use `OverloadedRecordDot` as much as possible
- Use `</>` operator for path joining
- Use `Set`s instead of (NE)lists. This is good for generating random data as picking random elements from a large `Set` is way faster than lists
- Rewrite some complex one-liners to something that we can understand 😄
- Remove underscores from most of the structures and Template Haskell generated lenses. I think we can even go entirely without lenses as, in practice, we don't really do that much nested data manipulation and I'm not sure a few more lines of code are worth the added cost of knowing lenses. However, if we still find lenses useful, we should move on to a more modern solution such as `generic-lens` or `optics-core` in the future.

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -6,7 +6,7 @@ module Echidna.ABI where
 
 import Control.Lens
 import Control.Monad (join, liftM2, liftM3, foldM, replicateM)
-import Control.Monad.Random.Strict (MonadRandom, getRandom, getRandoms, getRandomR, uniformMay)
+import Control.Monad.Random.Strict (MonadRandom, getRandom, getRandoms, getRandomR)
 import Control.Monad.Random.Strict qualified as R
 import Data.Binary.Put (runPut, putWord32be)
 import Data.BinaryWord (unsignedWord)
@@ -20,12 +20,11 @@ import Data.Foldable (toList)
 import Data.Hashable (Hashable(..))
 import Data.HashMap.Strict (HashMap)
 import Data.HashMap.Strict qualified as M
-import Data.HashSet (HashSet, fromList, union)
 import Data.Set (Set)
 import Data.Set qualified as Set
 import Data.List (intercalate)
 import Data.List.NonEmpty qualified as NE
-import Data.Maybe (fromMaybe, catMaybes, mapMaybe)
+import Data.Maybe (fromMaybe, catMaybes)
 import Data.Text (Text)
 import Data.Text qualified as T
 import Data.Text.Encoding (encodeUtf8)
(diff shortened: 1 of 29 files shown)


lib/Echidna/Output/Corpus.hs
```diff
@@ -12,21 +12,18 @@ import System.Directory (createDirectoryIfMissing, makeRelativeToCurrentDirector
 import Echidna.Types.Tx
 import Echidna.Output.Utils
 
-saveTxs :: Maybe FilePath -> [[Tx]] -> IO ()
-saveTxs (Just d) txs = mapM_ saveTx txs where
+saveTxs :: FilePath -> [[Tx]] -> IO ()
```
ggrieco-tob, line 15: Is this change correct? :thinking:

lib/Echidna/Solidity.hs
```diff
@@ -104,16 +106,18 @@ removeJsonFiles dir =
         let path = dir </> file
         whenM (doesFileExist path) $ removeFile path
 
-addresses :: SolConf -> NE.NonEmpty AbiValue
```
ggrieco-tob, line 107: NonEmpty has different properties than Set, so I'm not sure.

lib/Echidna/Output/Corpus.hs
arcz: Maybe is now matched earlier:
```haskell
  -- save corpus
  case cfg._cConf._corpusDir of
    Nothing -> pure ()
    Just dir -> do
      saveTxs (dir </> "reproducers") (filter (not . null) $ (.testReproducer) <$> cpg._tests)
      saveTxs (dir </> "coverage") (snd <$> Set.toList cpg._corpus)
```

lib/Echidna/Solidity.hs
arcz: I wish there were a non-empty set, but Set is better for what we do here as we pick random elements and list/non-empty list has slow random access. We would need to convert lists to sets, that takes linear time which kills the performance during fuzzing.

lib/Echidna/ABI.hs
```diff
@@ -309,16 +304,18 @@ mutateAbiCall = traverse f
...
+  let maybeValM = if genDict._pSynthA >= r then fromDict else pure Nothing
+      fromDict = case M.lookup t m of
+                   Nothing -> pure Nothing
+                   Just cs -> Just <$> rElem' cs
   fromMaybe <$> g t <*> maybeValM
 
 -- | Synthesize a random 'AbiValue' given its 'AbiType'. Requires a dictionary.
 genAbiValueM :: MonadRandom m => GenDict -> AbiType -> m AbiValue
-genAbiValueM genDict = genWithDict genDict (toList <$> genDict ^. constants) $ \case
+genAbiValueM genDict = genWithDict genDict genDict._constants $ \case
```
ggrieco-tob, line 318: Should this be faster than using `toList`, right?

arcz: yes, picking random elements from Set is faster (O(log n)) than from a list (O(n))

ggrieco-tob:
I'm getting this error when trying to save a corpus:

```
echidna-test: corpus-new/covered./1674496096/.txt: withFile: does not exist (No such file or directory)
```

arcz:
Fixed. Apparently, I got confused by the variable names so fixed them too.

Issue #865, opened 2022-12-05
Bug: unable to scroll echidna-test results in terminal


0xPhaze:
The test results are cut off and I'm unable to scroll past the terminal size.
<img width="565" alt="image" src="https://user-images.githubusercontent.com/103113487/205631452-297f6008-d40a-45c1-a28c-b3f38e839599.png">

Using apple mac m1 silicon, tested in zshell and VSCode bash terminal.

PR #915, opened 2023-01-23
Scrollable viewport for property list

arcz:
This PR implements a clickable scrollbar to scroll the property list. Closes https://github.com/crytic/echidna/issues/865.
scroll


elopez:
@ggrieco-tob seems to be working fine on interactive docker 👍

PR #920, opened 2023-01-24
Add numeric constants as bytes32

ggrieco-tob:
This can be solved:
```solidity
pragma solidity ^0.8.17;

contract Test {
    uint256 internal flag;

    function set(bytes32 x) external {
        assembly {
            if iszero(xor(x, 0x69)) { sstore(flag.slot, 0x01) }
        }
    }

    function echidna_flag() external view returns (bool) {
        return flag != 1;
    }
}
```


arcz:
It is useful only in rare circumstances. We decided to not pursue it at this time to reduce dictionary pollution. If there is a good argument for it, we can reopen.

PR #924, opened 2023-01-28
Refactor for RPC

arcz:
Some refactoring extracted from the RPC work.


ggrieco-tob:
Is this reducing the memory requirements when shrinking  (since it uses bc memo mapping)?

arcz:
It should be roughly the same, although I observed a slight decrease in CPU/memory usage during tests with those changes.

ggrieco-tob (approved):

PR #929, opened 2023-02-01
Use code hash instead of bytecode metadata

arcz:
~~This preserves the same properties but requires less effort.~~
Apparently, I was wrong and it doesn't work for contracts with immutables:
> It is important to note that the contract creation code generated by the compiler will modify the contract’s runtime code before it is returned by replacing all references to immutables by the values assigned to them. This is important if you are comparing the runtime code generated by the compiler with the one actually stored in the blockchain.
https://ethereum.stackexchange.com/questions/82240/what-is-the-immutable-keyword-in-solidity

### DIFF
--- a/lib/Echidna.hs
+++ b/lib/Echidna.hs
@@ -1,7 +1,6 @@
 module Echidna where
 
 import Control.Monad.Catch (MonadThrow(..))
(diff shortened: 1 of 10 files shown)


arcz:
Gonna try this with another approach in another PR

Issue #859, opened 2022-11-28
Rename multi-abi to something more explicit 


montyly:
`multi-abi` is a confusing name, and we have frequent questions about what it does. We could rename it to something like `--all-contracts`, which would be easier to understand. 

This would require to update all our tutorials (including building-secure-contracts)

ggrieco-tob:
Sure, we can include this in the next "major" release (with breaking changes).

PR #934, opened 2023-02-13
Rename multi-abi to allContracts

arcz:
Closes https://github.com/crytic/echidna/issues/859


ggrieco-tob:
We should be careful with the naming, as `allContract` only include the ones deployed during the constructor or before, but not during the fuzzing campaign.

arcz:
you mean the ones deployed over RPC?

ggrieco-tob:
The ones deployed during after the fuzzing campaign started are not included in `multi-abi`

PR #951, opened 2023-02-22
Make JSON parsing more robust

ggrieco-tob:


lib/Echidna/Processor.hs
```diff
@@ -107,30 +107,18 @@ instance FromJSON SlitherInfo where
...
 
-          "address" ->
-            case AbiAddress . Addr <$> readMaybe v of
-              Nothing -> failure v t
-              a -> pure a
+          "address" -> pure $ AbiAddress . Addr <$> readMaybe v
 
           -- we don't need all the types for now
           _ -> pure Nothing
-        where failure v t = fail $ "failed to parse " ++ t ++ ": " ++ v
```
arcz, line 133: I think that was made on purpose to give a better error message. Is it no longer needed?

ggrieco-tob: We want to do this parsing bulletproof, the reason is simple: slither sometimes output types are user defined, for constants, which can break the parsing. If missing a few constants should not break the tool.

PR #952, opened 2023-02-22
ci: remove `mstksg/get-package`

elopez:
This action has not been updated since 2019, and is currently causing some deprecation warnings to show up. This replaces it with a few run steps for each OS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,12 +23,6 @@ jobs:
         if: runner.os == 'macOS'
         run: brew install automake
 
-      - name: Get Packages (Linux)
-        if: runner.os == 'Linux'
-        run: |
-          sudo apt-get -q update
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -37,19 +30,21 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/.local/
-          key: ${{ runner.os }}-local-v4
+          key: ${{ runner.os }}-local-v5-${{ hashFiles('.github/scripts/install-*') }}
```
arcz, line 33: what happened here?

elopez: I added a hash of the scripts so it gets invalidated in case we ever update a dependency. Right now the cache has a fixed name and it'll still use old binaries even if we update the scripts.

arcz: sounds like a good change 👍

Issue #909, opened 2023-01-18
Echidna won't start if the `ERC20` name or symbol begins with `0x`


ChmielewskiKamil:
The contract below is taken from the [NaughtCoin level 15 of the Ethernaut wargame](https://github.com/OpenZeppelin/ethernaut/blob/master/contracts/contracts/levels/NaughtCoin.sol).

```solidity
// ...
contract NaughtCoin is ERC20 {
    // string public constant name = 'NaughtCoin';
    // string public constant symbol = '0x0';
    // uint public constant decimals = 18;
    uint256 public timeLock = block.timestamp + 10 * 365 days;
    uint256 public INITIAL_SUPPLY;
    address public player;

    constructor(address _player) ERC20("NaughtCoin", "0x0") {
        player = _player;
        INITIAL_SUPPLY = 1000000 * (10**uint256(decimals()));
        _mint(player, INITIAL_SUPPLY);
        emit Transfer(address(0), player, INITIAL_SUPPLY);
    }
// ...
```

If you import this contract into any Echidna test file, Echidna will freeze with the `Starting up, please wait...` log (similar to the [#889]). 

When run with the `--format text` flag it returns:
```
echidna-test: invalid b16 decoding of: "0x0"
CallStack (from HasCallStack):
  error, called at lib/Echidna/Processor.hs:123:31 in echidna-2.0.4-IHy1AKktzUiCzJpHK85UVS:Echidna.Processor
```

### Steps to reproduce the issue
[I've prepared an example repository](https://github.com/ChmielewskiKamil/naught-coin).

1. Clone the repo and cd into it: `git clone https://github.com/ChmielewskiKamil/naught-coin.git && cd naught-coin`
2. Run Echidna: `echidna-test src/TestNaughtCoin.sol --contract TestNaughtCoin --config src/config.yaml`
3. This results in: `echidna-test: invalid b16 decoding of: "0x0"`

As soon as you change the symbol to something different, like so:
```solidity
// change this
constructor(address _player) ERC20("NaughtCoin", "0x0")

// into this
constructor(address _player) ERC20("NaughtCoin", "ABC")
```
It starts fuzzing correctly.

If the token's name starts with `0x`, it does not work either.
```solidity
// this won't work
constructor(address _player) ERC20("0xNaughtCoin", "ABC")
```

This results in the same issue: `echidna-test: invalid b16 decoding of: "0xNaughtCoin"`.

### Additional info
[...]

PR #958, opened 2023-02-24
Fix parsing string constants that start with 0x

arcz:
Closes https://github.com/crytic/echidna/issues/909.


ggrieco-tob:
Is this a fix or a mitigation? :thinking:

Issue #964, opened 2023-02-27
Getting rate-limited when using Infura RPC URL


webthethird:
When trying to reproduce the COMP distribution bug, Echidna is crashing due to rate-limiting from my Infura RPC url endpoint.

![image](https://user-images.githubusercontent.com/6509209/221716450-0edd16db-e7f2-4298-9fa5-ae63385cae94.png)

ggrieco-tob:
@arcz let's fix this crash before the new release

ggrieco-tob:
I think this should be fixed. You will still be rate limited, but it shouldn't crash.

ggrieco-tob:
This was already fixed

PR #971, opened 2023-03-07
Wrap fetch exceptions

arcz:
This is a temporary solution to avoid crashing Echidna when rate limiting occurs. See https://github.com/crytic/echidna/issues/964.


ggrieco-tob:
This looks good, let's merge it.

PR #973, opened 2023-03-08
ci: add solc 0.8.x tests, mark 0.6.x and 0.7.x as not experimental

elopez:


arcz:
@elopez is this ready? I see all tests are passing

elopez:
@arcz CI-wise this is good to go, but I think some of the tests are flaky; I had to re-run boolean.sol on 0.8 and harvey_foo.sol on 0.6 for them to pass. If you're up for investigating why that's the case then go ahead and merge it 👍

Issue #979, opened 2023-03-13
log what echidna is doing at different stages e.g. startup/ shutdown


0xalpharush:
It'd be nice if Echidna gave more feedback on startup like "I'm running this command to compile smart contracts". On shutdown, it'd be nice to say "I'm writing the corpus to the disk here" (assuming a corpus dir was passed). Otherwise, it is unclear that everything is working correctly and progressing and something exceptional hasn't occurred (especially for something time consuming like writing the corpus to disk).

ggrieco-tob:
This is already in master

PR #988, opened 2023-03-20
Measure and log slow external actions

arcz:
This adds simple log messages to improve UX when compilation/running slither takes a long time. Corresponding issue https://github.com/crytic/echidna/issues/979.


ggrieco-tob:
We should be more specific that just "compiling", we could add "and running slither" to make sure people know that the time delay is caused by an external tool. We could add "please wait" to discourage users to abort early and report an issue in slither :disappointed:

PR #993, opened 2023-03-22
Make sure we don't use cheat codes to form fuzzing call sequences

arcz:


ggrieco-tob:
This works, but **requires** to start with an empty corpus in order to ensure that it will never fail again.

Issue #719, opened 2022-02-11
Running a large number of files from a corpus allocates a lot more memory


ggrieco-tob:
When the contracts to run are large and the number of sequences of transactions in the corpus is large (e.g. 500), reloading the corpus produces 2x or 3x mores memory than expected.

ggrieco-tob:
I should re-test this with the current `master`.

ggrieco-tob:
This is still an issue, but the amount of usual memory used during normal usage was reduced

PR #1002, opened 2023-03-29
Force corpus evaluation

arcz:
Should fix https://github.com/crytic/echidna/issues/719.


ggrieco-tob:
This helps to reduce the memory a lot. Let's merge it.

PR #1008, opened 2023-04-04
Show a proper error message when an invalid corpus is used

ggrieco-tob:


ggrieco-tob:
We don't know if a contract will be fetched from RPC.

arcz:
I guess we should store the order of corpus like we do with ncallseqs then replaying it should trigger fetches first

arcz:
Done in https://github.com/crytic/echidna/pull/1177

PR #1009, opened 2023-04-05
Flush stdout every time a new status line is produced

ggrieco-tob:


arcz:
looks good, we could also consider disabling buffering entirely

PR #1013, opened 2023-04-05
ci: enable GHC parallelism

elopez:


arcz:
@elopez do we want to move forward with this?

elopez:
@arcz I don't recall seeing any significant speed improvements when trying this, but I did not look much further into it. I'll close it and we can revisit this in the future if needed.

Issue #936, opened 2023-02-15
feat: generate .lcov coverage report


0xPhaze:
Echidna currently generates a .txt and a .html file when recording coverage. It would be nice if a .lcov file could also be generated. This could then be leveraged by existing tools to display [code coverage in VSCode](https://marketplace.visualstudio.com/items?itemName=semasquare.vscode-coverage-gutters) for example.

hellwolf:
I would be very happy if we could do that.

In our github workflow, we [publish lcovs from both hardhat testing and foundry testing](https://github.com/superfluid-finance/protocol-monorepo/blob/d15b3b2d02656d4a5483b9a0194cb4cb00e7a5fd/.github/workflows/ci.canary.yml#L177) and upload them to codecov; we also test echidna build with 1000 runs for now just for sanity check. If we had lcov, we could do a long running test and upload it to codecov too.

PatStiles:
I'm interested in trying this out

tusuii:
:+1:

PR #1024, opened 2023-04-12
Add .lcov coverage reports; add coverageFormats config option

samalws-tob:
This PR adds support for making .lcov coverage reports (in addition to the currently supported .txt and .html). It also adds a `coverageFormats` config option, which lets you choose which coverage reports should be made (default value: `["txt","html","lcov"]`).

See also https://github.com/crytic/echidna/issues/936


lib/Echidna/Output/Source.hs
```diff
@@ -27,17 +28,37 @@ import Echidna.Types.Signature (getBytecodeMetadata)
...
+                                           cc = ppCoveredCode isLcov isHtml sc cs s
                                      in do
                                        createDirectoryIfMissing True d
                                        writeFile fn cc
 
 -- | Pretty-print the covered code
-ppCoveredCode :: Bool -> SourceCache -> [SolcContract] -> CoverageMap -> Text
-ppCoveredCode isHtml sc cs s | s == mempty = "Coverage map is empty"
+ppCoveredCode :: Bool -> Bool -> SourceCache -> [SolcContract] -> CoverageMap -> Text
+ppCoveredCode isLcov isHtml sc cs s | s == mempty = "Coverage map is empty"
```
arcz, line 61: can we pass a `CoverageFileType` instead of those bools?

lib/Echidna/Config.hs
```diff
@@ -95,6 +96,7 @@ instance FromJSON EConfigWithUsage where
         <*> v ..:? "corpusDir" ..!= Nothing
         <*> v ..:? "mutConsts" ..!= defaultMutationConsts
         <*> v ..:? "coverageReport" ..!= True
+        <*> v ..:? "coverageFormats" ..!= [Txt,Html,Lcov]
```
arcz, line 99: I'm thinking if we should maybe merge this with `coverageReport` option. Like what should happen if `coverageReport` is `true` but `coverageFormats` is empty?

src/Main.hs
```diff
@@ -113,6 +113,7 @@ main = withUtf8 $ withCP65001 $ do
   slotsCache <- readIORef cacheSlotsRef
 
   -- save corpus
+  let saveCoverageInRequestedFormats a b c d e = void $ mapM (\fmt -> saveCoverage fmt a b c d e) cfg.campaignConf.coverageFormats
```
arcz, line 116: can we let saveCoverage to handle multiple formats instead of this helper function?

src/Main.hs
```diff
@@ -135,30 +135,28 @@ main = withUtf8 $ withCP65001 $ do
...
-        -- save source coverage reports
-        saveCoverage False runId dir sourceCache contracts campaign.coverage
-        saveCoverage True  runId dir sourceCache contracts campaign.coverage
+      -- coverage reports for external contracts, we only support
+      -- Ethereum Mainnet for now
+      when (chainId == Just 1) $ do
+        forM_ (Map.toList contractsCache) $ \(addr, mc) ->
+          case mc of
+            Just contract -> do
+              r <- externalSolcContract addr contract
```
arcz, line 150: After removing the condition `when cfg.campaignConf.coverageReport` I think we will now fetch external sources even if the coverageFormats is empty right? Or will laziness prevent that? It might be good to keep this still under a condition for clarity

PR #1030, opened 2023-04-18
Symbolic execution PoC

arcz:
A quick PoC that breaks this contract (included in tests):

```solidity
contract VulnerableContract {
   function func_one(int256 x) public pure {
     if (x / 4 == -20) {
       assert(false); // BUG
     }
   }

   function func_two(int128 x) public payable {
     if ((msg.value >> 30) / 7 == 2) {
       assert(false); // BUG
     }
   }
}
```

Screenshot 2023-04-19 at 00 21 49

Non-nix tests fail because I need to figure out how to install solvers :p

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -49,7 +49,7 @@
         echidna = with pkgs; lib.pipe
           (haskellPackages.callCabal2nix "echidna" ./. { inherit hevm; })
           [
-            (haskell.lib.compose.addTestToolDepends [ haskellPackages.hpack slither-analyzer solc ])
+            (haskell.lib.compose.addTestToolDepends [ haskellPackages.hpack slither-analyzer solc z3])
             (haskell.lib.compose.disableCabalFlag "static")
           ];
       in rec {
--- a/lib/Echidna/SymExec.hs
+++ b/lib/Echidna/SymExec.hs
@@ -30,7 +30,7 @@ exploreContract dst contract = do
     forM methods $ \method -> do
       let
         calldata = mkCalldata (Just (Sig method.methodSignature (snd <$> method.inputs))) []
-        vmSym = abstractVM calldata contract.runtimeCode Nothing AbstractStore
+        vmSym = abstractVM calldata contract.runtimeCode Nothing (ConcreteStore mempty)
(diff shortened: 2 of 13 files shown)


flake.nix
```diff
@@ -41,7 +41,7 @@
         echidna = with pkgs; lib.pipe
           (haskellPackages.callCabal2nix "echidna" ./. { })
           [
-            (haskell.lib.compose.addTestToolDepends [ haskellPackages.hpack slither-analyzer solc ])
+            (haskell.lib.compose.addTestToolDepends [ haskellPackages.hpack slither-analyzer solc z3])
```
d-xo, line 44: Would suggest playing around with cvc5 as well, it can often be faster

lib/Echidna/SymExec.hs
```diff
@@ -0,0 +1,56 @@
...
+import EVM.SMT
+
+import Echidna.Types.Tx
+
+exploreContract :: Addr -> SolcContract -> IO [Tx]
+exploreContract dst contract = do
+  let
+    calldata = mkCalldata Nothing [] -- fully abstract calldata, solution in buffers.txdata
+    -- Alternatively, more concrete calldata with a specific function,
+    -- not sure what is better yet
```
d-xo, line 24: I think you're probably going to make life easier for the solvers if you use more specialized calldata, as it allows us to simplify away quite a lot of abi decoding (amongst other things)

lib/Echidna/SymExec.hs
```diff
@@ -0,0 +1,56 @@
...
+    -- calldata = mkCalldata (Just (Sig method.methodSignature (snd <$> method.inputs))) []
+    vmSym = abstractVM calldata contract.runtimeCode Nothing SymbolicS
+    maxIter = Just 10
+    askSmtIters = Just 5
+    rpcInfo = Nothing
+    timeout = Nothing -- Just 1000 -- is it seconds?
```
d-xo, line 30: seconds on main, but milliseconds in 0.50.4

ggrieco-tob:
Already merged in #1216

Issue #1045, opened 2023-05-11
Save inputs as soon as possible


arcz:
We wait till the end of the campaign to dump reproducers/corpus to files. If Echidna crashes the information is most likely lost. Saving corpus also takes a considerable amount of time though I believe this is an issue on its own. 

With the new event system, we can make it a lot better with the following:

- Listen for `TestFalsified`/`TestOptimized` events and save reproducers
- Listen for `NewCoverage` events and save corpus, this needs the call sequence to be added to `NewCoverage` event
- Skip saving inputs that are replayed, `NewCoverage` is emitted during the replay, so this part most likely needs to look into the initial corpus to figure out if the input is already in there
- Remove saving reproducers and corpus after the campaign in main

arcz:
Done in https://github.com/crytic/echidna/pull/1167

PR #1047, opened 2023-05-17
Save corpus continuously while fuzzing

samalws-tob:
This PR adds a process which continuously saves corpus as new coverage/reproducers are discovered (#1045). This makes echidna crashes less inconvenient, and shortens the corpus saving time at the end.
It also makes the NewCoverage event include the list of transactions as part of the event, since this is needed when saving corpus info.

### DIFF
--- a/lib/Echidna/Output/Corpus.hs
+++ b/lib/Echidna/Output/Corpus.hs
@@ -1,13 +1,18 @@
 module Echidna.Output.Corpus where
 
+import Control.Concurrent (forkFinally, Chan, dupChan, newChan, readChan, writeChan)
+import Control.Monad (unless, void, when)
 import Control.Monad.Extra (unlessM)
 import Data.Aeson (ToJSON(..), decodeStrict, encodeFile)
 import Data.ByteString qualified as BS
 import Data.Hashable (hash)
-import Data.Maybe (catMaybes)
+import Data.Maybe (catMaybes, fromMaybe)
 import System.Directory (createDirectoryIfMissing, makeRelativeToCurrentDirectory, doesFileExist)
 import System.FilePath ((</>), (<.>))
 
+import Echidna.Types.Campaign (CampaignEvent(..), CampaignConf(..))
+import Echidna.Types.Config (Env(..), EConfig(..))
+import Echidna.Types.Test (EchidnaTest(..))
 import Echidna.Types.Tx (Tx)
 import Echidna.Utility (listDirectory, withCurrentDirectory)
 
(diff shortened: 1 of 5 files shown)


lib/Echidna/Output/Corpus.hs
```diff
@@ -26,3 +31,35 @@ loadTxs dir = do
   putStrLn ("Loaded " ++ show (length txSeqs) ++ " transaction sequences from " ++ dir)
   pure txSeqs
   where readCall f = decodeStrict <$> BS.readFile f
+
+-- save to corpus in the background while tests are running
+-- returns a channel that gets sent () when the process is done
+runCorpusSaver :: Env -> IO (Chan ())
```
arcz, line 37: can we use MVar instead of Chan here?

lib/Echidna/Output/Corpus.hs
```diff
@@ -26,3 +31,35 @@ loadTxs dir = do
...
+  Nothing -> do
+    finishChan <- newChan
+    writeChan finishChan ()
+    pure finishChan
+  Just dir -> do
+    -- we want to dupChan *before* forking so we don't miss any events
+    chan <- dupChan env.eventQueue
+    finishChan <- newChan
+    void $ forkFinally (loop dir chan nworkers) (const $ writeChan finishChan ())
+    pure finishChan
```
arcz, line 48: we can move `finishChan <- newChan` and `pure finishChan` out of the case so its not duplicated

lib/Echidna/Output/Corpus.hs
```diff
@@ -26,3 +31,35 @@ loadTxs dir = do
...
+  where
+    nworkers :: Int
+    nworkers = fromIntegral $ fromMaybe 1 env.cfg.campaignConf.workers
+
+    loop !dir !chan !workersAlive = when (workersAlive > 0) $ do
```
arcz, line 53: we have some logic duplication here but we can refactor this with the subscribe system we talked about

lib/Echidna/Output/Corpus.hs
samalws-tob: Is there a way to wait for MVar to be written to without doing an infinite loop? I used a Chan because you can wait for it using `readChan` instead of getting locked in a `while (thead is not finished)` loop

lib/Echidna/Output/Corpus.hs
samalws-tob: Separate PR for that subscribe system? I could add it to this PR if you want

lib/Echidna/Output/Corpus.hs
arcz: yeah, it should be possible

lib/Echidna/Output/Corpus.hs
arcz: let's do it in this PR then

samalws-tob:
#1048

PR #1048, opened 2023-05-18
Save corpus continuously while fuzzing; add TestSimplified and ErrorEvent events; change event system

samalws-tob:
This PR:
- Adds an event handler which continuously saves corpus as new coverage/reproducers are discovered (#1045)
  - Makes the NewCoverage event include the list of transactions as part of the event, since this is needed when saving corpus info
  - Unsimplified tests are saved with the prefix `unsimplified-`
 - Adds a TestSimplified event which is called whenever a test is fully simplified (so that the simplified transactions can be saved)
 - Adds an ErrorEvent event which is called when something goes wrong during event handling
 - Changes the event system to be based on an `IORef [event -> IO ()]` instead of channels. Channels were annoying because you would have to duplicate the channel and make a new thread for each event handler
 - Adds `spawnThread` and `awaitThreads` helper functions to spawn temporary threads and wait for all threads to be finished (in this PR used to save files, so that the worker doesn't have to wait while the file is saved)
 - Takes the absolute value of hash when deciding corpus filenames in order to avoid weird `-12345.txt` files
 
(Most of these changes are just solutions to problems I encountered while trying to make echidna save corpus while fuzzing)

Replaces #1047

### DIFF
--- a/lib/Echidna/Async.hs
+++ b/lib/Echidna/Async.hs
@@ -1,0 +1,39 @@
+module Echidna.Async where
+
+import Control.Concurrent.Thread.Group (forkIO)
+import Control.Monad (void)
+import Control.Monad.Reader (MonadReader, asks, ask)
+import Control.Monad.State.Strict (MonadState, gets)
+import Control.Monad.IO.Class (MonadIO, liftIO)
+import Data.IORef (atomicModifyIORef', readIORef)
+import Data.Time (LocalTime)
+
+import Echidna.Types.Campaign (CampaignEvent, WorkerState(..))
+import Echidna.Types.Config (Env(..))
+import Echidna.Utility (getTimestamp)
+
+spawnThread :: Env -> IO a -> IO ()
+spawnThread env io = void $ forkIO env.threadGroup io
+
+addEventHandler
+  :: (MonadReader Env m, MonadIO m)
+  => ((Int, LocalTime, CampaignEvent) -> IO ())
+  -> m ()
+addEventHandler f = do
+  handlersRef <- asks (.eventHandlers)
+  liftIO $ atomicModifyIORef' handlersRef (\l -> (f:l, ()))
+
+pushEvent
+  :: (MonadReader Env m, MonadState WorkerState m, MonadIO m)
+  => CampaignEvent
+  -> m ()
+pushEvent event = do
+  workerId <- gets (.workerId)
+  env <- ask
+  liftIO $ pushEventIO env workerId event
+
+pushEventIO :: Env -> Int -> CampaignEvent -> IO ()
+pushEventIO env workerId event = do
+  time <- liftIO getTimestamp
+  handlers <- readIORef env.eventHandlers
+  mapM_ (\f -> spawnThread env $ f (workerId, time, event)) handlers
--- a/lib/Echidna/Output/Corpus.hs
+++ b/lib/Echidna/Output/Corpus.hs
@@ -1,5 +1,9 @@
 module Echidna.Output.Corpus where
 
+import Control.Exception (handle, IOException)
+import Control.Monad (unless)
+import Control.Monad.Reader (MonadReader, ask)
+import Control.Monad.IO.Class (MonadIO)
 import Control.Monad.Extra (unlessM)
 import Data.Aeson (ToJSON(..), decodeStrict, encodeFile)
 import Data.ByteString qualified as BS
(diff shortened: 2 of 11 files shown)


lib/Echidna/Async.hs
```diff
@@ -0,0 +1,46 @@
...
+import Data.Time (LocalTime)
+
+import Echidna.Types.Campaign (CampaignEvent, WorkerState(..))
+import Echidna.Types.Config (Env(..))
+import Echidna.Utility (getTimestamp)
+
+awaitThreads :: Env -> IO ()
+awaitThreads env = do
+  n <- readIORef env.numUnfinishedThreads
+  when (n > 0) $ threadDelay 10_000 >> awaitThreads env
```
arcz, line 18: This works but I wonder if there is another concurrency primitive we could use to block instead of threadDelay. I think Chan would work here but maybe there is something simpler. We also have to think about exceptions in those threads so they don't fail silently.

This pattern of spawning short-running jobs and waiting for them to finish sounds like it should be a solved problem 🤔 Maybe something from here https://hackage.haskell.org/package/async-2.2.4/docs/Control-Concurrent-Async.html ?

lib/Echidna/Output/Corpus.hs
```diff
@@ -26,3 +33,19 @@ loadTxs dir = do
...
+  where
+    saveEvent env dir (_, _, event) = maybe (pure ()) (saveFile env dir) $ getEventInfo event
+
+    getEventInfo (TestFalsified test)    = Just ("reproducers", test.reproducer)
+    getEventInfo (TestOptimized test)    = Just ("reproducers", test.reproducer)
+    getEventInfo (TestSimplified test)   = Just ("reproducers", test.reproducer)
+    getEventInfo (NewCoverage _ _ _ txs) = Just ("coverage", txs)
+    getEventInfo _ = Nothing
+
+    saveFile env dir (subdir, txs) = unless (null txs) $ spawnThread env $ saveTxs (dir </> subdir) [txs]
```
arcz, line 51: Like I mentioned earlier, if anything goes wrong while saving the file, we won't know.

lib/Echidna/Types/Campaign.hs
```diff
@@ -44,7 +44,8 @@ data CampaignConf = CampaignConf
 data CampaignEvent
   = TestFalsified !EchidnaTest
   | TestOptimized !EchidnaTest
-  | NewCoverage !Int !Int !Int
+  | TestSimplified !EchidnaTest
```
arcz, line 47: This is fine, but I'm thinking, should we use the "shrink" vocabulary here so it's easier to figure out what's going on? Like `TestShrunk`?

arcz (commented):
Great work! We can do a few improvements to make it more robust.

lib/Echidna/Output/Corpus.hs
```diff
@@ -26,3 +34,21 @@ loadTxs dir = do
...
+
+-- setup a handler to save to corpus in the background while tests are running
+setupCorpusSaver :: (MonadReader Env m, MonadIO m) => m ()
+setupCorpusSaver = do
+  env <- ask
+  maybe (pure ()) (addEventHandler . saveEvent env) env.cfg.campaignConf.corpusDir
+  where
+    saveEvent env dir (workerId, _, event) = maybe (pure ()) (saveFile workerId env dir) $ getEventInfo event
+
+    getEventInfo (TestFalsified test) = Just ("reproducers", "unsimplified-", test.reproducer)
```
arcz, line 46: `unshrunk-` maybe to match the other vocabulary?

lib/Echidna/Types/Config.hs
```diff
@@ -64,9 +64,11 @@ data Env = Env
...
   , dapp :: DappInfo
 
-  -- | Shared between all workers. Events are fairly rare so contention is
-  -- minimal.
-  , eventQueue :: Chan (Int, LocalTime, CampaignEvent)
+  , eventHandlers :: IORef [(Int, LocalTime, CampaignEvent) -> IO ()]
+
+  -- mainly for handling events, but can be used for any purpose
+  -- `wait` is called on this group before echidna closes
+  , threadGroup :: ThreadGroup
```
arcz, line 71: We should move this to `spawnListener` and make all the handlers spawned async

lib/Echidna/Types/Campaign.hs
```diff
@@ -44,8 +44,10 @@ data CampaignConf = CampaignConf
 data CampaignEvent
   = TestFalsified !EchidnaTest
   | TestOptimized !EchidnaTest
-  | NewCoverage !Int !Int !Int
+  | TestShrunk !EchidnaTest
+  | NewCoverage !Int !Int !Int [Tx]
   | TxSequenceReplayed !Int !Int
+  | ErrorEvent !String
```
arcz, line 50: Maybe `HandlerFailed`? We can also add some haddock to make it clearer

src/Main.hs
```diff
@@ -127,6 +131,8 @@ main = withUtf8 $ withCP65001 $ do
 
   tests <- readIORef testsRef
 
+  ThreadGroup.wait threadGroup
```
arcz, line 134: We should move this to `spawnListener` before `putMVar stopVar ()`

lib/Echidna/Types/Config.hs
samalws-tob: Not sure I follow? I got rid of `spawnListener`, and all handlers are spawned async now

arcz:
Done in https://github.com/crytic/echidna/pull/1167.

PR #1064, opened 2023-06-02
Make continuous fuzzing default

### LINKED ISSUE #477: Continuous fuzzing mode
Continuous fuzzing mode

siraben:
Fixes #477


lib/Echidna/Config.hs
```diff
@@ -85,7 +85,7 @@ instance FromJSON EConfigWithUsage where
         pure $ TestConf classify (const psender)
 
       campaignConfParser = CampaignConf
-        <$> v ..:? "testLimit" ..!= defaultTestLimit
```
arcz, line 88: We forgot to remove `defaultTestLimit` function 😄

arcz:
Tests have to be updated too, you can run them with `cabal run tests`

divyaranjan1905:
Is there something specific stopping this from being merged?

elopez:
I'm not sure, but I see there's conflicts on the PR. There's also an alternative PR we could opt for here https://github.com/crytic/echidna/pull/1302

divyaranjan1905:
Maybe that could be merged, it'd be a nice thing to have :)

dguido:
@gustavo-grieco I think we should close this PR, in preference for #1302. What do you think?

PR #1071, opened 2023-06-08
Use contract address as keys for coverageMap

siraben:
Adapts the commit from arcz/no-bememo. N.B. tests do not cover changes to coverage maps


arcz:
Done in https://github.com/crytic/echidna/pull/1160.

PR #1075, opened 2023-06-15
Collect and show the events from all the transactions

ggrieco-tob:
This PR needs some work to make sure the events are collected after the trace is minimized. We should doing it during the minimization, since it is not going to be always used (`shrinkLimit` could be zero).


webthethird:
Remember to use a different field to store events from the constructor, since printing them for every invariant failure is repetitive.

PR #1082, opened 2023-06-21
Init simple installer

### LINKED ISSUE #811: Create a 1-line installer and upgrader
Something like `curl -L URL | bash` that would
- Download the latest version
- Copy it in something like `/home/userName/.echidna`
- Add to the PATH 

In addition, an `echidna-upgrade` script that would download the latest version if available. The script would also be added to `/home/userName/.echidna`

bohendo:
resolves #811 

This script will:
- exit if echidna is already installed
- exit if python3 is not installed
- install pip3 if not already present
- install slither via pip3 if not already present
- download the mac or linux tarball per current system
- verify checksum
- unpack & place echidna executable at `$HOME/.echidna/bin/echidna`
- add PATH update to .bashrc or .zshenv or other shell-specific config

This has been tested manually on MacOS.

TODO:
- upgrade old echidna to latest version (basic layout of `$HOME/.echidna` is well suited to supporting this feature)
- add automated tests for MacOS + Linux
- add Windows support + tests
- optional: host this script somewhere memorable & easy to type out on the command line

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,6 +4,12 @@ set -e
 version="2.2.0"
 
 # Do nothing if echidna is already installed
+if echidna-test --version > /dev/null 2>&1
+then
+  echo "WARNING: $(echidna-test --version) is installed but is out-of-date, we recommend removing it"
+fi
+
+# Do nothing if echidna is already installed
 if echidna --version > /dev/null 2>&1
 then
   echo "$(echidna --version) is already installed" && exit
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,136 @@
+#!/usr/bin/env bash
+set -e
+
+version="2.2.0"
+
+# Do nothing if echidna is already installed
+if echidna --version > /dev/null 2>&1
+then
+  echo "$(echidna --version) is already installed" && exit
+  # TODO: check that the version is up to date, otherwise offer to upgrade
+fi
+
+# In the unlikely scenario that python3 isn't available, let the user install it manually
+if ! python3 --version > /dev/null 2>&1
+then echo "python3 is not available, please install it manually before attempting to install echidna" && exit 1
+fi
+
+# install pip3 if not already installed
+if ! pip3 --version > /dev/null 2>&1
+then
+  echo "pip3 is not available, installing it now.."
+  curl https://bootstrap.pypa.io/get-pip.py | python3
+fi
+
+# install slither if not already installed
+if ! slither --version > /dev/null 2>&1
+then
+  echo "slither is not available, installing it now.."
+  pip3 install slither-analyzer --user
+fi
+
+echo "Installing Echidna v$version.."
+
+BASE_DIR=${XDG_CONFIG_HOME:-$HOME}
+ECHIDNA_DIR=${ECHIDNA_DIR-"$BASE_DIR/.echidna"}
+ECHIDNA_BIN_DIR="$ECHIDNA_DIR/bin"
+ECHIDNA_ARCHIVE_DIR="$ECHIDNA_DIR/v$version"
+ECHIDNA_ARCHIVE_PATH="$ECHIDNA_ARCHIVE_DIR/echidna.tar.gz"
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -0,0 +1,142 @@
...
+then
+  echo "pip3 is not available, installing it now.."
+  curl https://bootstrap.pypa.io/get-pip.py | python3
+fi
+
+# install slither if not already installed
+if ! slither --version > /dev/null 2>&1
+then
+  echo "slither is not available, installing it now.."
+  pip3 install slither-analyzer --user
```
elopez, line 35: It would be cool to have slither updated if it's out of date. PATH may also need to be adjusted to account for slither installed with `--user`.

Maybe we should use a (fresh?) venv instead and just link `slither{,-*}` and `crytic-compile` to the directory we then add to the PATH for echidna?

install.sh
```diff
@@ -0,0 +1,142 @@
...
+ECHIDNA_DIR=${ECHIDNA_DIR-"$BASE_DIR/.echidna"}
+ECHIDNA_BIN_DIR="$ECHIDNA_DIR/bin"
+ECHIDNA_ARCHIVE_DIR="$ECHIDNA_DIR/v$version"
+ECHIDNA_ARCHIVE_PATH="$ECHIDNA_ARCHIVE_DIR/echidna.tar.gz"
+ECHIDNA_DIGEST_PATH="$ECHIDNA_ARCHIVE_DIR/echidna.sha256"
+TRUE_BIN_PATH="$ECHIDNA_ARCHIVE_DIR/echidna/echidna"
+BIN_PATH="$ECHIDNA_BIN_DIR/echidna"
+
+if [[ "$(uname)" == "Darwin" ]]
+then arch="MacOS"
```
elopez, line 50: we should differentiate ARM vs x86 builds, maybe by inspecting `uname -m`. Otherwise M1 users will get slow fuzzing performance.

install.sh
```diff
@@ -0,0 +1,142 @@
...
+elif [[ "$(uname)" == "Linux" ]]
+then arch="Ubuntu"
+else echo "Unsupported system: $(uname)" && exit 1
+fi
+
+TARBALL_URL="https://github.com/crytic/echidna/releases/download/v$version/echidna-$version-$arch.tar.gz"
+DIGEST_URL="https://github.com/crytic/echidna/releases/download/v$version/echidna-$version-$arch.tar.gz.sha256"
```
elopez, line 57: Our digests are not signed, so someone tampering with the releases may also modify the digest. Is this meant only as an integrity check? Otherwise hardcoding the hashes next to the version in the script may provide a bit more of certainty (unless someone tampers with the script as well)

elopez (commented):
Great work @bohendo! 🎉  I left some thoughts inline

ggrieco-tob:
I think we should address @elopez concerns and then merge.

install.sh
bohendo: Is there a separate release for ARM vs x86 macs? Under release assets I only see MacOS vs Ubuntu vs Windows

install.sh
bohendo: Just integrity, hardcoding the hashes seems like a better path forward 👍

install.sh
elopez: @bohendo  yeah, there's MacOS-aarch64 files, with a different structure inside the tarball, e.g see https://github.com/crytic/echidna/releases/tag/v2.1.0

These are manually built (with the nix bundle target in the readme), compressed and uploaded; it seems the latest release is missing them. We used to have a more comprehensive build process that generated a ready to upload archive and fixed TERMINFO on the resulting binaries (the current process has [this bug again](https://github.com/crytic/echidna/pull/731/files)), maybe we can bring that back to aid releases going forward?

install.sh
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env bash
+set -e
+
+version="2.2.0"
+slither_version="0.9.3"
```
0xalpharush, line 5:
```suggestion
slither_version="0.9.6"
```

Issue #1085, opened 2023-06-29
Print a warning if 0.8.20 is used


montyly:
Until we support PUSH0

elopez:
Related: #1049

Issue #1049, opened 2023-05-20
echidna regression test failure while upgrading truffle to 5.9.1


chenrui333:
Looks like the default contract address is not deployable

```
==> /opt/homebrew/Cellar/echidna/2.1.1/bin/echidna --format text /private/tmp/echidna-test-20230520-94978-1wzi1hh
echidna: Deploying the contract 0x00a329c0648769A73afAc7F9381E08FB43dBEA72 failed (revert, out-of-gas, sending ether to an non-payable constructor, etc.):

[2023-05-20 18:20:56.70] Compiling /private/tmp/echidna-test-20230520-94978-1wzi1hh... Done! (36.216785s)
Analyzing contract: /private/tmp/echidna-test-20230520-94978-1wzi1hh/contracts/test.sol:True
/opt/homebrew/Library/Homebrew/ignorable.rb:29:in `block in raise'
```

relates to https://github.com/Homebrew/homebrew-core/pull/131489

chenrui333:
actually the behavior is also different with truffle 5.9.0

```
==> /opt/homebrew/Cellar/echidna/2.1.1/bin/echidna --format text /private/tmp/echidna-test-20230520-67224-hhaaj6
  [2023-05-20 15:59:43.25] Compiling /private/tmp/echidna-test-20230520-67224-hhaaj6... Done! (2.228798s)
  Analyzing contract: /private/tmp/echidna-test-20230520-67224-hhaaj6/contracts/test.sol:True
  [2023-05-20 15:59:45.48] Running slither on /private/tmp/echidna-test-20230520-67224-hhaaj6... Done! (0.743016s)
  [2023-05-20 15:59:49.23] [status] tests: 0/1, values: [], fuzzing: 28129/50000, cov: 92, corpus: 2
  echidna_true:  passed! 🎉
  Unique instructions: 92
  Unique codehashes: 1
  Corpus size: 2
  Seed: 4497770500591259[64](https://github.com/Homebrew/homebrew-core/actions/runs/5034489150/jobs/9029349688?pr=131502#step:3:65)
  Error: echidna: failed
  An exception occurred within a child process:
    Minitest::Assertion: Expected: 1
    Actual: 0
```

See https://github.com/Homebrew/homebrew-core/pull/131502

elopez:
[...]

PR #1090, opened 2023-07-11
Update hevm to 0.51.3

arcz:
Closes https://github.com/crytic/echidna/issues/1085 and https://github.com/crytic/echidna/issues/1049.

### DIFF
--- a/lib/Echidna/Transaction.hs
+++ b/lib/Echidna/Transaction.hs
@@ -16,9 +16,9 @@ import Data.Set (Set)
 import Data.Set qualified as Set
 import Data.Vector qualified as V
 
-import EVM hiding (resetState, VMOpts(timestamp, gasprice))
+import EVM (initialContract, loadContract, bytecode)
 import EVM.ABI (abiValueType)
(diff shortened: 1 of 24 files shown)


lib/Echidna/Transaction.hs
```diff
@@ -116,7 +116,7 @@ canShrinkTx Tx { call, gasprice = 0, value = 0, delay = (0, 0) } =
 canShrinkTx _ = True
 
 removeCallTx :: Tx -> Tx
-removeCallTx t = Tx NoCall 0 t.src 0 0 0 t.delay
+removeCallTx t = Tx NoCall t.src t.dst 0 0 0 t.delay
```
ggrieco-tob, line 119: Why this was changed?

arcz: `t.src` was in the wrong place and having 0 in the src caused a crash in tx generation which was previously not triggered as there was something deployed on the address 0. I think this was changed in hevm and now there is nothing deployed on 0.

rappie:
Does this include the new cheatcodes for `diffusc`?

arcz:
Not yet, unfortunately. It is still in progress https://github.com/ethereum/hevm/pull/259.

PR #1110, opened 2023-09-01
Read assert locations and determinate if they were executed or not

ggrieco-tob:


samalws-tob:
rebased onto master

lib/Echidna/Solidity.hs
```diff
@@ -342,7 +342,7 @@ mkWorld SolConf{sender, testMode} sigMap maybeContract slitherInfo contracts =
   let
     eventMap = Map.unions $ map (.eventMap) contracts
     payableSigs = filterResults maybeContract slitherInfo.payableFunctions
-    as = if isAssertionMode testMode then filterResults maybeContract slitherInfo.asserts else []
```
ggrieco-tob, line 345: This is a TODO

lib/Echidna.hs
```diff
@@ -90,7 +90,7 @@ prepareContract cfg solFiles buildOutput selectedContract seed = do
                      seed
                      (returnTypes contracts)
 
-  pure (vm, env, dict)
+  pure (vm, env, dict, slitherInfo.asserts)
```
arcz, line 93: Can we put this into env?

PR #1113, opened 2023-09-07
Fix broken link

ardislu:
The Homebrew repo split up all the formulae into subdirectories a few weeks ago. This PR updates the link to point to the current location of the `echidna.rb` file.


elopez (approved):
Good catch, thanks for the PR!

PR #1117, opened 2023-09-15
Print bytes32 values as hex

ggrieco-tob:
This small fix makes easier to reproduce property failures in Solidity


aviggiano:
Thanks for this feature!

There's a minor issue, I think it's not correctly padding the output string to 64 characters:

```
  Call sequence:
    liquidate(21033444676813639511009812671505242594468849768767286983973634464167,0x0000000000000000000000000b6840c016347,4)
```

ggrieco-tob:
@aviggiano please take a look to see if it is padding correctly in the last commit

dguido:
@gustavo-grieco Is this done? It looks done-ish.

gustavo-grieco:
I think this is done, but perhaps it a change that hevm wants to do as well. Let me ask them.

PR #1131, opened 2023-10-20
Deliver status information using server-sent events

ggrieco-tob:
A proof of concept of delivering events using SSE. This will open port 3413 and deliver the same events than available in text mode. Still in process of discussion


arcz:
I think we should put it behind a CLI/config option like `--server <port>`

ggrieco-tob:
Yeap, let's do that.

PR #1133, opened 2023-10-26
nix: static builds and release workflow

elopez:
This PR implements (partially) static builds for Echidna using Nix, similarly to [how it's implemented in hevm](https://github.com/ethereum/hevm/pull/185). The Linux build in particular uses musl, to avoid the problems that glibc has with static binaries and multithreading.

It also replaces our current Nix workflow with a new one that builds regular echidna, as well as this new redistributable/static build on the three platforms we can run Nix on Actions:
* linux x86_64
* macOS x86_64
* macOS aarch64

These binaries are then signed (with sigstore) and uploaded to a draft release in case the workflow was triggered by a Git tag, easing our release preparation. If the workflow was triggered by something else (branch, PR), the redistributable executables are available as a workflow artifact.

### TODO (help wanted!)
* [X] See if we can select musl just for echidna-redistributable; rebuilding everything with musl is not quick and it does not seem to be cached by nixos (~2h cold cache build on GH actions).
  * Added Cachix to the workflow, so this should not be an issue 

### Future work
* [ ] Consider adding a Windows build to the release workflow.


arcz (approved):
Amazing!

PR #1136, opened 2023-11-01
docker: make container multi-platform

elopez:
This enables aarch64 builds for our Docker container image. These can then be used on e.g. M1 Macs to get good performance. It also makes the echidna binary static with musl -- previously it was static with glibc, which might be problematic as shown on #962 if the binary is re-used outside this container (I believe using it inside the container should not cause issues as the glibc version matches between binary and runtime environment)

Caveat: The container is built on an x86 runner using emulation, so the build is somewhat slow (~4h total build time).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,8 +14,8 @@ jobs:
   docker:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
+      - name: Set up QEMU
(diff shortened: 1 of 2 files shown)


elopez:
This would allow us to then include the aarch64 binary in eth-security-toolbox.

elopez:
Closing in favor of #1352 which uses native ARM runners to do the build, but we should probably revisit the conversion of the container to alpine/musl at some point

PR #1137, opened 2023-11-06
fix: do not fail if slither fails

0xalpharush:
Pass `--no-fail` to slither to avoid erroring out. This requires at least Slither 0.9.2 https://github.com/crytic/slither/commit/53238600a1ad3b119b26d752460dcfeea75dd20c

This is probably not the robust approach. I would prefer if we did not make running Slither a requirement to run Echidna. We should **only run it if it's installed** (log a warning "install slither for improved ...") and **not rely on Slither succeeding** i.e. continue fuzzing campaign if it fails

### DIFF
--- a/lib/Echidna/Processor.hs
+++ b/lib/Echidna/Processor.hs
@@ -130,7 +130,7 @@ runSlither fp solConf = do
       ProcessorNotFound "slither" "You should install it using 'pip3 install slither-analyzer --user'"
     Just path -> pure path
 
-  let args = ["--ignore-compile", "--print", "echidna", "--json", "-"]
+  let args = ["--ignore-compile", "--print", "echidna", "--json", "--no-fail", "-"]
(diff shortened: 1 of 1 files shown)


lib/Echidna/Processor.hs
```diff
@@ -130,7 +130,7 @@ runSlither fp solConf = do
       ProcessorNotFound "slither" "You should install it using 'pip3 install slither-analyzer --user'"
     Just path -> pure path
 
-  let args = ["--ignore-compile", "--print", "echidna", "--json", "-"]
+  let args = ["--ignore-compile", "--print", "echidna", "--json", "--no-fail", "-"]
```
0xalpharush, line 133:
```suggestion
  let args = ["--ignore-compile", "--print", "echidna", "--json", "-", "--no-fail"]
```

arcz:
Done in https://github.com/crytic/echidna/pull/1159.

PR #1152, opened 2023-12-26
Fix nix flake for development

arcz:
Use regular dynamic build for development.


arcz:
Done in https://github.com/crytic/echidna/pull/1154

Issue #646, opened 2021-04-21
Add a flag to disable slither integration


ggrieco-tob:
We have a vyper workaround that skips slither but ideally we want allow the user to disable slither if it fails for some reason.

avniculae:
Hey! Echidna fails with the following message `echidna: Error running slither:` (no text after colon). What's the current solution for disabling Slither in Echidna? As a side note, I'm not sure why Slither fails to be honest -- it worked well until I've brought in another contract into the fuzzing harness.

avniculae:
Thanks for your quick reply! Here's the full error output.
<details>
  <summary>Error Traceback</summary>

```
'forge clean' running (wd: /path/.../)
'forge build --build-info --force' running
Compiling 150 files with 0.8.17
Solc 0.8.17 finished in 11.69s
Compiler run successful (with warnings)
[...warning info printed...]

Traceback (most recent call last):
  File "/opt/homebrew/Cellar/slither-analyzer/0.9.3/libexec/lib/python3.11/site-packages/slither/__main__.py", line 837, in main_impl
    ) = process_all(filename, args, detector_classes, printer_classes)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/slither-analyzer/0.9.3/libexec/lib/python3.11/site-packages/slither/__main__.py", line 101, in process_all
    ) = process_single(compilation, args, detector_classes, printer_classes)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/slither-analyzer/0.9.3/libexec/lib/python3.11/site-packages/slither/__main__.py", line 79, in process_single
    slither = Slither(target, ast_format=ast, **vars(args))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/slither-analyzer/0.9.3/libexec/lib/python3.11/site-packages/slither/slither.py", line 135, in __init__
    self._init_parsing_and_analyses(kwargs.get("skip_analyze", False))
[...]

PR #1159, opened 2024-01-08
Make slither optional

arcz:
Don't fail, instead print warning instead and continue. Includes a small refactor.

Closes https://github.com/crytic/echidna/issues/646 and https://github.com/crytic/echidna/pull/1137.


arcz:
There is no way to explicitly disable slither, can you open an issue for that?

ggrieco-tob:
@aviggiano I think you can create an empty bash script called `slither` in some directory (e.g. `test`) and then run echidna using `PATH=./test:$PATH`. That should "disable" slither.

PR #1160, opened 2024-01-09
Index CoverageMap and SignatureMap by codehash (for performance)

samalws-tob:
This PR makes it so the CoverageMap and SignatureMap are indexed by contracts' codehashes, rather than their bytecode metadata. As a result:
- [This](https://github.com/opynfinance/ConvexityProtocol/blob/dev/contracts/echidna/EchidnaOptionsContract.sol) contract ran 60% faster on my computer, producing the same output as before.
- [This](https://github.com/makerdao/dss-vest/blob/master/echidna/ChainLog.sol) contract ran 18% faster on my computer, producing the same output as before.

Additional changes:
- Removes the bytecode metadata cache, since this is no longer needed
- Adds a codehash cache. This is because we need to index maps by a contract's "real" / "original" / "compile-time" codehash, rather than its `.codehash` property (which can be different if Solidity's immutables feature, which changes bytecode during compile time, is used). The real codehash is found using hevm's `findSrc` function, and the results are cached in the codehash map
- Adds an additional test for immutables.
- Uses `codeContract` rather than `contract` to determine the current contract while keeping track of coverage (after talking this over with @arcz ). Beforehand, this was done in some places but not in others. I believe this should fix some delegate call behavior.


lib/Echidna/Exec.hs
```diff
@@ -292,35 +285,25 @@ execTxWithCov tx = do
...
-              writeIORef covContextRef (True, Just (vec', pc))
-            else do
-              -- TODO: should we collect the coverage here? Even if there is no
-              -- bytecode for external contract, we could have a "virtual" location
-              -- that PC landed at and record that.
-              pure ()
-          Just vec ->
+            contract = currentContract vm
+
+        maybeMetaVec <- lookupUsingCodehashOrInsert env.codehashMap contract env.dapp env.coverageRef $ do
```
arcz, line 290: can we rename this to `maybeCovVec` or similar as we don't operate on metadata anymore

lib/Echidna/Types/Coverage.hs
```diff
@@ -1,18 +1,19 @@
...
 import Data.Vector.Unboxed.Mutable qualified as V
 import Data.Word (Word64)
+import EVM.Types (W256)
 
 import Echidna.Types.Tx (TxResult)
 
--- | Map with the coverage information needed for fuzzing and source code printing
-type CoverageMap = Map ByteString (IOVector CoverageInfo)
+-- | Map with the coverage information needed for fuzzing and source code printing.
+-- Indexed by contracts' "real" codehash; see `CodehashMap`.
```
arcz, line 15: we can say compile-time codehash, "real" can be understood both ways :D

lib/Echidna/Types/Signature.hs
```diff
@@ -31,24 +28,8 @@ type SolCall = (FunctionName, [AbiValue])
...
-      Nothing     -> bs -- if no metadata is found, return the complete bytecode
-      Just (_, m) -> m
-
-lookupBytecodeMetadata :: MetadataCache -> ByteString -> ByteString
-lookupBytecodeMetadata memo bs = fromMaybe (getBytecodeMetadata bs) (memo M.!? bs)
-
--- | Precalculate getBytecodeMetadata for all contracts in a list
-makeBytecodeCache :: [ByteString] -> MetadataCache
-makeBytecodeCache bss = M.fromList $ bss `zip` (getBytecodeMetadata <$> bss)
+-- | Indexed by contracts' "real" codehash; see `CodehashMap`.
```
arcz, line 31: same as the other comment regarding "real"

lib/Echidna/Transaction.hs
```diff
@@ -53,18 +56,18 @@ getSignatures hmm (Just lmm) =
...
-  :: (MonadRandom m, MonadState WorkerState m)
-  => MetadataCache
-  -> World
+  :: (MonadIO m, MonadRandom m, MonadState WorkerState m, MonadReader Env m)
+  => World
   -> TxConf
   -> Map (Expr EAddr) Contract
   -> m Tx
-genTx memo world txConf deployedContracts = do
+genTx world txConf deployedContracts = do
```
arcz, line 64: if we have `Env` now, we can grab `txConf` from there, no need to pass it

arcz (commented):
This looks great. The immutable.sol seems to fail because `contractAList` in `genTx` is empty, so it looks like a bug with SignatureMap.

arcz (approved):
Amazing work!

PR #1175, opened 2024-01-23
symexec rebase and symbolic VM from concrete

siraben:


ggrieco-tob:
Already merged in #1216

PR #1192, opened 2024-02-28
Speed up shrinking in some cases

### LINKED ISSUE #1028: [Bug-Candidate]: Shrinking extremely slow in some rare situations
### Describe the issue:

In some rare situations shrinking can be really slow. I can work around this by setting `shrinkLimit` really high but it would be nice to have a real fix.

@arcz has access to a private repo where this can be reproduced.

### Version:

Many

### Relevant log output:

image

This `1` can shrink to a `0`:
image

samalws-tob:
This PR changes the process used to shrink a contract call. For a given contract call, instead of trying to shrink all the parameters at once, it picks a subset of the parameters and shrinks those. Subset size is chosen out of `[1,2,rng(0,n),n]`; this means that sometimes we will try to shrink just one parameter, sometimes we will try to shrink all parameters, etc.
I'm fairly sure this fixes #1028 , and it also definitely fixes the following proof-of-concept bug, which the current echidna code on the master branch is unable to shrink:
```solidity
contract Test {
  function foo(bool a, uint256 b) public {
    assert(!a); // `b` won't shrink to 0 because echidna always tries to shrink `a` to False at the same time
  }
}
```

## Summary by CodeRabbit

- **Refactor**
	- Improved the logic for shrinking values in ABI calls, ensuring more efficient and uniform shrinking probabilities.

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -1,7 +1,7 @@
 module Echidna.ABI where
 
 import Control.Monad (liftM2, liftM3, foldM, replicateM)
-import Control.Monad.Random.Strict (MonadRandom, join, getRandom, getRandoms, getRandomR)
+import Control.Monad.Random.Strict (MonadRandom, join, getRandom, getRandoms, getRandomR, uniform, fromList)
 import Control.Monad.Random.Strict qualified as Random
 import Data.Binary.Put (runPut, putWord32be)
 import Data.BinaryWord (unsignedWord)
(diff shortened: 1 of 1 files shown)


arcz:
This is quite a fancy solution however I feel it's a bit too complex. Wouldn't something like this work too:
```haskell
shrinkAbiCall (func, vals) = do
  vals' <- forM vals $ \val -> do
    chance :: Float <- getRandomR (0,1)
    -- 50% chance of shrinking a value
    if chance < 0.5 then shrinkAbiValue val
                    else pure val
  pure (func, vals')
```

ggrieco-tob:
> This is quite a fancy solution however I feel it's a bit too complex. Wouldn't something like this work too:
> 
> ```haskell
> shrinkAbiCall (func, vals) = do
>   vals' <- forM vals $ \val -> do
>     chance :: Float <- getRandomR (0,1)
>     -- 50% chance of shrinking a value
>     if chance < 0.5 then shrinkAbiValue val
>                     else pure val
>   pure (func, vals')
> ```

Actually, we don't need floats here, let's keep everything in integers if possible.

samalws-tob:
The reason I wanted to avoid doing that is if we have a function with a lot of arguments and we can shrink only one of them, the odds of that one (and no others) getting picked by the simplified algo is 1/2^n

arcz:
Ok, let's see how it performs and adjust if we feel the need

PR #1194, opened 2024-02-28
feat: add CLI commands for RPC URL and block number

### LINKED ISSUE #1193: Add `--rpc-url` and `--rpc-block` CLI commands

In order to add RPC configs as github actions input for crytic echidna action, we need to expose them as CLI commands.

Related: https://github.com/crytic/echidna-action/issues/19

penandlim:
Implemented CLI arguments `--rpc-url` and `--rpc-block` to allow overriding the corresponding environment variables.

The precedence for settings is now: Command-Line Arguments > Environment Variables > Configuration Defaults.

Closes #1193 

## Summary by CodeRabbit

- **New Features**
	- Introduced new command-line options for specifying RPC URL and block number, enhancing user control over configuration settings.
- **Enhancements**
	- Improved configuration override mechanism to prioritize command-line inputs over environment variables, offering users more flexibility in setting preferences.

### DIFF
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -214,8 +214,8 @@ versionOption = infoOption
 
 overrideConfig :: EConfig -> Options -> IO EConfig
 overrideConfig config Options{..} = do
-  rpcUrl <- maybe (Onchain.rpcUrlEnv) (pure . Just) cliRpcUrl
-  rpcBlock <- maybe (Onchain.rpcBlockEnv) (pure . Just) cliRpcBlock
+  rpcUrl <- (pure cliRpcUrl) <|> Onchain.rpcUrlEnv
+  rpcBlock <- (pure cliRpcBlock) <|> Onchain.rpcBlockEnv
   pure $
     config { solConf = overrideSolConf config.solConf
            , campaignConf = overrideCampaignConf config.campaignConf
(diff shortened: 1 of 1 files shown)


src/Main.hs
```diff
@@ -206,8 +214,8 @@ versionOption = infoOption
 
 overrideConfig :: EConfig -> Options -> IO EConfig
 overrideConfig config Options{..} = do
-  rpcUrl <- Onchain.rpcUrlEnv
-  rpcBlock <- Onchain.rpcBlockEnv
+  rpcUrl <- (pure cliRpcUrl) <|> Onchain.rpcUrlEnv
```
arcz: This unfortunately won't work since <|> here operates on IO, not on Maybe. Can you rename `rpcUrl` to `envRpcUrl` (same for block) and change this instead: `rpcUrl = cliRpcUrl <|> envRpcUrl <|> config.rpcUrl`

arcz (changes requested):
Thanks for contributing! There is a little mistake in your solution but it's easily fixable.

arcz (approved):
LGTM!

src/Main.hs
penandlim: Updated. Thanks for the quick review!

Issue #1134, opened 2023-10-31
[Bug-Candidate]: Multiple seemingly related performance issues


rappie:
### Describe the issue:

I'm encountering multiple seemingly related performance issues. It is a bit of a mess, but i'll try to over them one by one.

### Using Foundry to compile
Command:
```shell
echidna . --contract PerformanceTest --test-mode assertion
```

**Issue 1:**
When commenting out the `fallback` function
```solidity
fallback() external payable virtual {}
```
performance drops from 10.000 calls/s to 1000 calls/s

![image](https://github.com/crytic/echidna/assets/1430820/bb510071-f361-4c15-988e-612516f2a4d5)
![image](https://github.com/crytic/echidna/assets/1430820/59f53c70-5fc6-41b8-bd1c-15c6d045e31b)

**Issue 2:**
When commenting out the mapping assignment
```solidity
MAPPING[address(0)] = 123;
```
performance drops from 10.000 calls/s to 2000 calls/s

Both of these effects are related to a third issue (see below). Using foundry to compiling (providing `.` to Echidna) needs issues 1 or 2 to trigger issue 3.

### Using single file compilation
Command:
```shell
echidna src/PerformanceTest.sol --contract PerformanceTest --test-mode assertion
```

In this case issues 1&2 are not necessary to trigger issue 3.

**Issue 3:**
Toggling between
```solidity
contract A {
    // B b = new B();

    function a() public view returns (uint256) {
        return 1;
        // return b.a();
    }
}
```
and
```solidity
contract A {
    B b = new B();

    function a() public view returns (uint256) {
        // return 1;
        return b.a();
    }
}
```

drops performance from 3500 calls/s to 1000 calls/s

I hope this makes sense and helps with debugging, it took a lot of work to figure this out. Please let me know if you have any questions or problems reproducing.

### Code example to reproduce the issue:

https://github.com/rappie/echidna-debug-performance

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.20;

abstract contract WithFallback {
    // fallback() external payable virtual {}
}

contract PerformanceTest {
    A a;

    mapping(address => uint256) internal MAPPING;

    constructor() {
        a = new A();
    }

    function test() public {
        // MAPPING[address(0)] = 123;
        uint256 result = a.a();
    }
}

contract A {
    B b = new B();

    function a() public view returns (uint256) {
        // return 1;
        return b.a();
    }
}

contract B {
    C c = new C();

[...]

PR #1200, opened 2024-03-07
Update filterFallbacks to be selective rather than all-or-nothing

samalws-tob:
This PR changes the `filterFallbacks` function so that it filters out fallbacks selectively, rather than either removing fallbacks from all contracts or none. It also removes the first argument of the function, which was unused.

Fixes issue 1 of #1134 (now it's 1000 calls/s in both cases, rather than 1000 in one and 10000 in the other; the performance only improved before because it was trying a bunch of fallback calls, which executed quickly because they didn't do anything)

### DIFF
--- a/lib/Echidna.hs
+++ b/lib/Echidna.hs
@@ -15,7 +15,7 @@ import EVM (cheatCode)
 import EVM.ABI (AbiValue(AbiAddress))
 import EVM.Dapp (DappInfo(..), dappInfo)
 import EVM.Fetch qualified
-import EVM.Solidity (SolcContract(..), BuildOutput)
+import EVM.Solidity (BuildOutput)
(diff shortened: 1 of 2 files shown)


lib/Echidna.hs
```diff
@@ -73,7 +73,7 @@ prepareContract env solFiles specifiedContract seed = do
                                funs
 
     eventMap = Map.unions $ map (.eventMap) contracts
-    world = mkWorld solConf eventMap signatureMap specifiedContract slitherInfo
+    world = mkWorld solConf eventMap signatureMap specifiedContract slitherInfo contracts
```
arcz, line 76: we get `eventMap` from `contracts`, can we do it inside `mkWorkd`?

PR #1201, opened 2024-03-08
Fix mac builds with stack build in nix develop

samalws-tob:
This fixes `stack build` via `nix develop` on my mac, which was previously failing due to "ld: library not found for -lsecp256k1". I'm mostly opening this PR to provide the diff in case somebody has the same issue as me

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -147,6 +147,9 @@
             packages = _: [ (echidna pkgs) ];
             shellHook = "hpack";
             buildInputs = [
+              pkgs.secp256k1
+              pkgs.libff
+              pkgs.gmp
               solc
               slither-analyzer
               haskellPackages.hlint


elopez:
we now have secp256k1 and libff on the `develop` shell, and builds work fine on Mac (with cabal at least, haven't tried with stack). Should we close this or do we also need to add gmp?

samalws-tob:
stack build didn't work for me initially but now I'm having a hard time testing it. I added gmp and zlib to the flake.nix and that fixed the stack build, but afterwards when I took them off of the flake.nix again the stack build still worked, even when tried deleting the echidna directory, re-cloning, and building again. so there's some permanent state sitting in my computer and I don't know where

samalws-tob:
in any case cabal build worked for me out of the box so stack build might not be very necessary

elopez:
I'll close this PR then, we can revisit this if we see failures again in the future

PR #1216, opened 2024-03-14
Add symbolic execution

samalws-tob:
This PR is a continuation of #1030 and #1175. It adds a `symExec` configuration option which enables an extra symbolic execution worker, which runs alongside the existing fuzzing workers. This should allow for bugs which require very specific inputs to be caught more easily (see `tests/solidity/symbolic/sym.sol` for a fairly contrived example). Symbolic execution is mostly implemented by hevm; the bulk of this PR has to do with calling hevm's symbolic executor in the right way, and spawning a new worker to do this.
The sym exec worker runs at startup using the initial corpus, and also runs whenever a fuzz worker finds new coverage. It uses a queue to keep track of this. When the queue is empty, the worker is in a sleep mode waiting for more events to come in.
This PR also adds `symExecTimeout` and `symExecNSolvers` configuration options which allow for choosing the timeout and number of sym exec solvers (i.e. Z3 instances, I think). It's unclear what a good default value for these would be, so I just picked something that looks right (30 and 1, respectively).

This is meant to be an experimental feature, and this PR shouldn't affect users unless they explicitly enable `symExec`. It is turned off by default.

### DIFF
--- a/.github/scripts/install-z3.sh
+++ b/.github/scripts/install-z3.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -eux -o pipefail
+
+if [ "$HOST_OS" = "Linux" ]; then
+  sudo apt-get update
+  sudo apt-get install -y z3
+fi
+# symbolic tests are currently disabled on windows because they fail
+# if [ "$HOST_OS" = "Windows" ]; then
+#   choco install z3
+# fi
--- a/.github/scripts/install-z3.sh
+++ b/.github/scripts/install-z3.sh
@@ -5,7 +5,6 @@ if [ "$HOST_OS" = "Linux" ]; then
   sudo apt-get update
   sudo apt-get install -y z3
 fi
-# symbolic tests are currently disabled on windows because they fail
-# if [ "$HOST_OS" = "Windows" ]; then
-#   choco install z3
-# fi
+if [ "$HOST_OS" = "Windows" ]; then
+  choco install z3
+fi
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -5,7 +5,7 @@ module Echidna.Campaign where
 
 import Control.Concurrent
 import Control.DeepSeq (force)
-import Control.Monad (replicateM, when, void, forM_)
+import Control.Monad (replicateM, when, unless, void, forM_)
 import Control.Monad.Catch (MonadThrow(..))
 import Control.Monad.Random.Strict (MonadRandom, RandT, evalRandT)
 import Control.Monad.Reader (MonadReader, asks, liftIO, ask)
@@ -16,6 +16,7 @@ import Control.Monad.Trans (lift)
 import Data.Binary.Get (runGetOrFail)
 import Data.ByteString.Lazy qualified as LBS
 import Data.IORef (readIORef, atomicModifyIORef')
+import Data.Foldable (foldlM)
 import Data.List qualified as List
 import Data.Map qualified as Map
 import Data.Map (Map, (\\))
@@ -28,6 +29,7 @@ import System.Random (mkStdGen)
 
 import EVM (cheatCode)
 import EVM.ABI (getAbi, AbiType(AbiAddressType), AbiValue(AbiAddress))
+import EVM.Solidity (SolcContract)
 import EVM.Types hiding (Env, Frame(state), Gas)
 
 import Echidna.ABI
@@ -35,6 +37,7 @@ import Echidna.Exec
 import Echidna.Mutator.Corpus
 import Echidna.Shrink (shrinkTest)
 import Echidna.Symbolic (forceAddr)
+import Echidna.SymExec (createSymTx)
 import Echidna.Test
 import Echidna.Transaction
 import Echidna.Types (Gas)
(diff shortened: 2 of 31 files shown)


ggrieco-tob:
This is great!. I have some suggestions, but this first needs some testing. Btw, I think this code will not work as expected when  `allContracts` is enabled, right?

ggrieco-tob:
Some UI improvements we need (as simple textual events):

* Show when a new coverage is result from using symbolic execution (or fuzzing)
* Show when a worker is hitting a solver timeout
* Show when a solver is returning unfeasible
* Show when a worker has finished and has nothing else to do
* Show a message when the campaign is terminated because there is no concrete workers and all the symbolic workers are waiting.

samalws-tob:
> This is great!. I have some suggestions, but this first needs some testing. Btw, I think this code will not work as expected when allContracts is enabled, right?

Yeah, it won't be able to call any contracts aside from the "main one"

lib/Echidna/Campaign.hs
```diff
@@ -78,10 +81,121 @@ replayCorpus vm txSeqs =
...
+        continue
+    where
+    continue = lift callback >> run objs
+
+  symexecTxs txs = do
+    cfg <- asks (.cfg)
+    vm' <- foldlM (\vm' tx -> snd <$> execTx vm' tx) vm txs
+    (threadId, symTxsChan) <- liftIO $ createSymTx cfg name cs vm'
+
+    modify' (\ws -> ws { runningThreads = [threadId] })
```
arcz, line 185: Won't this override other `runningThreads`? Otherwise, why a list here?

lib/Echidna/Campaign.hs
```diff
@@ -78,10 +81,121 @@ replayCorpus vm txSeqs =
...
+  continueLoopMVar <- liftIO $ newMVar () -- used to pause `run` when nothing is happening
+  threadsLeftRef <- liftIO $ newIORef (getNFuzzWorkers cfg.campaignConf)
+  newCovTxs <- liftIO $ newIORef ([] : map snd initialCorpus)
+  txsToApplyToState <- liftIO $ newIORef []
+  let objs = (continueLoopMVar, threadsLeftRef, newCovTxs, txsToApplyToState)
+
+  void $ spawnListener (listenerFunc objs)
+
+  flip runStateT initialState $
+    flip evalRandT (mkStdGen effectiveSeed) $ do -- unused but needed for callseq
```
arcz, line 126: Yeah this is a bit unfortunate, `callseq` requires `MonadRandom` because we shrink in `updateTest` but we most likely shouldn't. Can you add this note so we remember to clean it up?

lib/Echidna/Campaign.hs
```diff
@@ -78,10 +81,121 @@ replayCorpus vm txSeqs =
...
+
+    modify' (\ws -> ws { runningThreads = [threadId] })
+    lift callback
+
+    symTxs <- liftIO $ takeMVar symTxsChan
+
+    modify' (\ws -> ws { runningThreads = [] })
+    lift callback
+
+    mapM_ (\symTx -> void $ callseq vm (txs ++ [symTx])) symTxs
```
arcz, line 193: We can start from `vm'` and just execute `[symTx]`, will save a good chunk of work

ggrieco-tob: This is a great optimization.

ggrieco-tob:
A few small additional improvements:
* Allow to set symExec., symExecTimeout and symExecNSolvers from command line. This will allow a quick experimentation without having to change the base config file
* (If it is not too hard): Allow to select the number of symboilc transactions to execution. This will usually take a very small positive number, by default is 1, but it can be 2 or 3 as well. I believe this will impact here: ` mapM_ (\symTx -> void $ callseq vm (txs ++ [symTx])) symTxs`

lib/Echidna/Campaign.hs
```diff
@@ -78,10 +81,121 @@ replayCorpus vm txSeqs =
...
+runSymWorker callback vm dict workerId initialCorpus name cs = do
+  cfg <- asks (.cfg)
+
+  continueLoopMVar <- liftIO $ newMVar () -- used to pause `run` when nothing is happening
+  threadsLeftRef <- liftIO $ newIORef (getNFuzzWorkers cfg.campaignConf)
+  newCovTxs <- liftIO $ newIORef ([] : map snd initialCorpus)
+  txsToApplyToState <- liftIO $ newIORef []
+  let objs = (continueLoopMVar, threadsLeftRef, newCovTxs, txsToApplyToState)
+
+  void $ spawnListener (listenerFunc objs)
```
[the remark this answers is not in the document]
samalws-tob: Currently this only supports one symbolic worker at a time

lib/Echidna/Campaign.hs
samalws-tob: There's only ever 1 (or 0) runningThreads at a time

samalws-tob:
- I added @ggrieco-tob 's suggestions, with one exception: I didn't add an event for when a symbolic worker is idle, since this would require peeking from a channel, which as far as I can tell isn't possible.
- I disabled windows tests for symbolic execution, since these were failing. We can ship this as an experimental feature with linux and mac support for now, and fix windows later.
- I’m unsure about the ToJSON implementation for SymNoNewCoverage, someone please check that this is fine.

samalws-tob:
Also added config options for setting the `maxIters` and `askSMTIters` variables passed to hevm

rappie:
> Also added config options for setting the `maxIters` and `askSMTIters` variables passed to hevm

Could this be used to lower the memory consumption?

samalws-tob:
Yep, although it only affects memory consumption of symbolic execution, not the fuzz workers

tests/solidity/basic/default.yaml
```diff
@@ -91,3 +91,17 @@ rpcBlock: null
 workers: 1
 # events server port
 server: null
+# whether to add an additional symbolic execution worker
+symExec: false
+# timeout for symbolic execution SMT solver
+# only relevant if symExec is true
+symExecNSolvers: 1
+# number of SMT solvers used in symbolic execution
```
rappie, line 99: Comments for these 2 variables are flipped around

elopez:
This PR on hevm may resolve the issues on Windows https://github.com/ethereum/hevm/pull/484

rappie:
> Yep, although it only affects memory consumption of symbolic execution, not the fuzz workers

I've ran some tests and it is using a LOT less memory with the default config now.

```solidity
    function testSymExec(uint256 magicNumber) public {
        if (magicNumber == 1256984789498465878979846549849854654984)
            t(false, "you found the magic number");
    }
```

What is a good approach to test the symexec itself? I've tried the above code but it does not break this in a couple of minutes.

samalws-tob:
Related note here: I'm working on implementing concolic execution, which should significantly lower the memory usage & runtime when executing on functions with a lot of branches (I kept running into this issue when testing this PR). I should have a proof-of-concept available in a day or two. Most of the code for concolic execution will be shared with the code from this PR, so any bug you find here will probably also be a bug in the concolic version.

samalws-tob:
@ your test example: It shouldn't take that long to find the test case, unless there's other functions in the contract that it tries executing first

samalws-tob:
> This PR on hevm may resolve the issues on Windows https://github.com/ethereum/hevm/pull/484

Thanks @elopez !

samalws-tob:
Just pushed a commit adding support for concolic execution. It works, but:
- We need to add some tests for it
- There's a few minor TODOs to work out
...so I'm going to convert this PR back to a draft PR

Any testing done on both the symexec and concolic execution features is appreciated :)

samalws-tob:
Fixed up the todos, now all that's left is:
- Adding tests
- Reverting stack.yaml (I have it set to a fork of hevm, awaiting my & Emilio's PRs getting merged)
- Review

.github/scripts/install-z3.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -eux -o pipefail
+
+if [ "$HOST_OS" = "Linux" ]; then
+  sudo apt-get update
+  sudo apt-get install -y z3
+fi
+# symbolic tests are currently disabled on windows because they fail
```
elopez, line 8:
```suggestion

```

samalws-tob: whoops, thanks

m4k2:
Hi, I want to try out the new features, I'm currently using Echidna version 2.2.3 and encountering issues. It seems that some options are not supported yet ? 

```bash
Warning: unused option: symExec
Warning: unused option: symExecAskSMTIters
Warning: unused option: symExecConcolic
Warning: unused option: symExecMaxIters
Warning: unused option: symExecNSolvers
Warning: unused option: symExecTargets
Warning: unused option: symExecTimeout
```

Maybe it's for the next release ?

rappie:
> Maybe it's for the next release ?
Correct.

You can download one of the latest builds here (scroll down)
https://github.com/crytic/echidna/actions/runs/9522275975

Or manually build the latest `master` branch (what I prefer):
https://github.com/crytic/echidna?tab=readme-ov-file#building-using-nix-works-natively-on-apple-m1-systems

PR #1245, opened 2024-04-29
Upgrade build to GHC 9.6

elopez:


lib/Echidna/UI.hs
```diff
@@ -247,20 +241,19 @@ ui vm world dict initialCorpus cliSelectedContract cs = do
...
     liftIO $ mapM_ killThread (threadId : workerState.runningThreads)
 
-vtyConfig :: IO Config
+vtyConfig :: IO VtyUserConfig
 vtyConfig = do
-  config <- Vty.standardIOConfig
-  pure config { inputMap = (Nothing, "\ESC[6;2~", EvKey KPageDown [MShift]) :
-                           (Nothing, "\ESC[5;2~", EvKey KPageUp [MShift]) :
-                           inputMap defaultConfig }
+  pure defaultConfig { configInputMap = [
```
hellwolf, line 253: an alternative is to use:

```
config <- Vty.userConfig
```

which is empty config or load some user settings. Per doc, I have never tried though!

ggrieco-tob:
Haskell part looks good to me!

PR #1265, opened 2024-05-28
coverage: count number of executions per line

elopez:


arcz (commented):
This looks good, I would just benchmark if this doesn't add too much overhead

ggrieco-tob:
I think either the workers will be frequently sending events (quite bad) or some counter will not be properly updated (not so bad, still not great).

ggrieco-tob:
@elopez can you also include the changes from here? https://github.com/crytic/echidna/pull/1110

rappie:
This is awesome, I want this feature so bad :laughing:

elopez:
I did a second attempt at implementing this a safer way. The code is quite ugly right now but if anyone wants to test/review it before I spend some more time on it I'd appreciate it 👍

rappie:
I don't see this branch under Actions to download the binary :thinking:

ggrieco-tob:
I think it is because it is a draft, let me convert it a proper PR

elopez:
@rappie There was a small git conflict, hopefully it builds now that I resolved it 👍

lib/Echidna/Types/Coverage.hs
```diff
@@ -17,9 +18,20 @@ import Echidna.Types.Tx (TxResult)
 -- Indexed by contracts' compile-time codehash; see `CodehashMap`.
 type CoverageMap = Map W256 (IOVector CoverageInfo)
 
+-- | Map with the statistic information needed for source code printing.
+-- Indexed by contracts' compile-time codehash; see `CodehashMap`.
+type StatsMap = Map W256 (IOVector StatsInfo)
+
+-- | Map with the statistic information needed for source code printing.
+-- Indexed by contracts' compile-time codehash; see `CodehashMap`.
```
samalws-tob, line 26: might want to mention here that StatsMap is used during runtime and StatsMapV is used at the end when generating coverage

lib/Echidna/Exec.hs
```diff
@@ -304,13 +314,13 @@ execTxWithCov tx = do
             -- bug in another place, investigate.
             -- ... this should be fixed now, since we use `codeContract` instead
             -- of `contract` for everything; it may be safe to remove this check.
-            when (pc < VMut.length vec) $
+            when (pc < VMut.length vec) $ do
+              VMut.modify (fromJust maybeStatsVec) (\(execQty, revertQty) -> (execQty + 1, revertQty)) opIx
```
samalws-tob, line 318: I don't see anywhere where we increment revertQty

lib/Echidna/Output/Source.hs
```diff
@@ -29,8 +32,25 @@ import EVM.Solidity (SourceCache(..), SrcMap, SolcContract(..))
...
+  return [(exec1 + exec2, revert1 + revert2) | (exec1, revert1) <- vec1 | (exec2, revert2) <- vec2]
+
+mvToList :: (VU.Unbox a) => VU.IOVector a -> IO [a]
+mvToList = fmap U.toList . U.freeze
+
+combineStats :: TLS (IORef StatsMap) -> IO StatsMapV
+combineStats statsRef = do
+  threadStats' <- allTLS statsRef
+  threadStats <-  mapM readIORef threadStats' :: IO [StatsMap]
+  statsLists <- pure $ map (\(m :: StatsMap) -> Map.map (\(x :: VU.IOVector StatsInfo) -> mvToList x) m) threadStats :: IO [Map EVM.Types.W256 (IO [StatsInfo])]
```
samalws-tob, line 52:
```suggestion
  let statsLists = map (Map.map mvToList) threadStats :: IO [Map EVM.Types.W256 (IO [StatsInfo])]
```

lib/Echidna/Output/Source.hs
```diff
@@ -103,7 +125,7 @@ ppCoveredCode fileType sc cs s | null s = pure "Coverage map is empty"
   pure $ topHeader <> T.unlines (map ppFile allFiles)
 
 -- | Mark one particular line, from a list of lines, keeping the order of them
-markLines :: CoverageFileType -> V.Vector Text -> S.Set Int -> Map Int [TxResult] -> V.Vector Text
+markLines :: CoverageFileType -> V.Vector Text -> S.Set Int -> Map Int ([TxResult], ExecQty) -> V.Vector Text
```
samalws-tob, line 128: looks like we never print out number of reverts

lib/Echidna/Output/Source.hs
```diff
@@ -167,8 +194,14 @@ srcMapCov sc covMap contracts = do
...
                   innerUpdate =
                     Map.alter
-                      (Just . (<> unpackTxResults txResults) . fromMaybe mempty)
+                      updateLine
                       line
+                  updateLine (Just (r, q)) = Just ((<> unpackTxResults txResults) r, max q execQty)
+                  updateLine Nothing = Just (unpackTxResults txResults, execQty)
+                  fileStats = Map.lookup c.runtimeCodehash statMap
+                  idxStats | isJust fileStats = fromJust fileStats U.! opIx
+                           | otherwise        = (0, 0)
```
samalws-tob, line 203:
```suggestion
                  idxStats = maybe (0, 0) (U.! opIx) fileStats
```

lib/Echidna/Exec.hs
elopez: yeah, I have not implemented counting of reverts. I meant to do it at first but then it wasn't clear to me where I should be counting them. If you have any pointers there I'd appreciate them, otherwise I may remove the extra var at this time.

lib/Echidna/Exec.hs
```diff
@@ -296,6 +297,15 @@ execTxWithCov tx = do
             forM_ [0..size-1] $ \i -> VMut.write vec i (-1, 0, 0)
             pure $ Just vec
 
+        statsRef <- getTLS env.statsRef
+        maybeStatsVec <- lookupUsingCodehashOrInsert env.codehashMap contract env.dapp statsRef $ do
+          let size = BS.length . forceBuf . fromJust . view bytecode $ contract
+          if size == 0 then pure Nothing else do
+            -- IO for making a new vec
+            vec <- VMut.new size
```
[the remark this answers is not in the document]
elopez: I'll replace the other instance of new + for we have right next to this one as well

samalws-tob:
I benchmarked this, looks like it cuts performance in half

samalws-tob:
in #1305 tried doing `type StatsInfo = ExecQty` instead of `type StatsInfo = (ExecQty, RevertQty)` because I thought the issue might've had to do with tuples being non-strict, but it's still far slower than master

rappie:
I'm still very interested in this, despite the performance loss. It would be great to have this in `master` as a togglable feature :slightly_smiling_face:

PR #1274, opened 2024-06-14
Emit "saved reproducer" log message as event rather than putStrLn

samalws-tob:


arcz:
You wrote "rather than putStrLn" but I don't see any `putStrLn` removed 🤔

Issue #1105, opened 2023-08-16
[Bug-Candidate]: Echidna multicore produces slower shrinking behavior


aviggiano:
### Describe the issue:

I am running echidna with 8 workers, but I am getting unexpected/unwanted results for the shrinking step when a bug is found:

Instead of shrinking _faster_, when I add more workers the process is significantly _slower_.

Sometimes I see the shrink step go down, and then up, then down, and then up. It seems like increasing the number of workers adds a bunch of (duplicate) transactions that do not contribute to the smallest sequence possible.

I did a small 1-min test, reutilizing the corpus, with 1 worker vs 8 workers. You can see below that using more workers makes echidna run much slower when a bug is found.

### 1 worker

<img width="1109" alt="Screenshot 2023-08-16 at 15 52 02" src="https://github.com/crytic/echidna/assets/3029017/b5ea3351-8472-423a-8a57-e5cf3f830608">

### 8 workers

<img width="1319" alt="Screenshot 2023-08-16 at 15 49 21" src="https://github.com/crytic/echidna/assets/3029017/f12ea6f8-e482-4225-99e5-3356b1df0561">

### Code example to reproduce the issue:

N/A

### Version:

Echidna 2.2.1
slither 0.9.2

### Relevant log output:

```shell
N/A
```

rappie:
I often encounter the same thing. I think this is because the different workers can overwrite each others "best shrunk" version and kind of work against each other that way.

arcz:
Yes, this is a known issue to be fixed

Issue #1249, opened 2024-05-06
Shrinking multicore


aviggiano:
### Describe the desired feature

Currently, it seems like Echidna uses a single core/process/thread to shrink failed sequences.

In some cases, however, we're interested in using 100% of the machine's resources to extract the results of that particular sequence. For example, this is often the case when I am using `stopOnFail: true` and `workers: N` in a multicore setup. I don't care about _other_ failed properties, I only care about that particular one that I know to have failed. The problem is that this takes forever even if I bump up N or the number of cores.

It seems like trying to shrink a sequence on a `c5.large` instance takes about the same amount of time as on a `c5.4xlarge` (benchmark pending), which is unexpected.

ggrieco-tob:
It seems it was already reported here: #1105.. by you :smile:

ggrieco-tob:
Shrinking with multiple cores is hard with our current (stochastic) approach, since the number of messages required between the core is likely to be high, and that will cause an overhead that will offset the gains for multiple workers. If you can @aviggiano, please test https://github.com/crytic/echidna/pull/1250 to see if that improved the shrinking speed on a single worker.

PR #1280, opened 2024-06-27
Shrink on one worker

arcz:
There was a high contention while shrinking with multiple workers at the same time. This locks shrinking to the same worker which solved the test. Fixes https://github.com/crytic/echidna/issues/1105.

Changes and implications:
1. Only the worker which solved a test will shrink the test. Worker number is displayed in the UI while shrinking.
3. If there is shrinking work, do it until the shrink limit and then go back to fuzzing. Fuzzing will be put on pause when all workers are shrinking.
5. Test limit is unevenly distributed. When worker shrinks it doesn't work on its portion of test limit. This will be fixed in another PR.
6. Refactored `selectMainContract`, `mkTests` and `mkSignatureMap` out of `loadSpecified`. This allows for more granular control to easier construct `Env`. Moved `world` to the `Env` and tests can be created before `Env`.
7. Merged `Campaign.updateTest` and `Test.updateOpenTest` into `Campaign.updateOpenTest`.

TODO: Look into reintroducing multicore shrinking with better synchronization (https://github.com/crytic/echidna/issues/1249).

### DIFF
--- a/lib/Echidna.hs
+++ b/lib/Echidna.hs
@@ -3,7 +3,7 @@ module Echidna where
 import Control.Concurrent (newChan)
 import Control.Monad.Catch (MonadThrow(..))
 import Control.Monad.ST (RealWorld)
-import Data.IORef (writeIORef, newIORef)
+import Data.IORef (newIORef)
 import Data.List (find)
 import Data.List.NonEmpty (NonEmpty)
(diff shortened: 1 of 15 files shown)


ggrieco-tob:
Let's update the number of workers to 3. We can probably pick a better number, but it should work while we determinate how to do that.

PR #1282, opened 2024-06-28
Refactors in Exec.hs

samalws-tob:
Factors a few pieces of code out into separate functions to make it more readable

### DIFF
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -11,7 +11,7 @@ import Control.Monad.ST (ST, stToIO, RealWorld)
 import Control.Monad.State.Strict (MonadState(get, put), execState, runStateT, MonadIO(liftIO), gets, modify', execStateT)
 import Data.Bits
 import Data.ByteString qualified as BS
-import Data.IORef (readIORef, newIORef, writeIORef, modifyIORef')
+import Data.IORef (IORef, readIORef, newIORef, writeIORef, modifyIORef')
 import Data.Map qualified as Map
 import Data.Maybe (fromMaybe, fromJust)
 import Data.Text qualified as T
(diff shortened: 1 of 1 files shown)


gustavo-grieco:
@elopez fixed conflict and rebased this one. I think we can merge it, it will make Exec.hs a bit better.

gustavo-grieco:
@elopez it seems we forgot about this one for the release, I think you had a comment about it?

elopez:
from Claude:

# Review: `simplifyExecTx`

Single commit `183e1420` "Simplify Exec.hs" (+136/−131), touching only
`lib/Echidna/Exec.hs`. Static review — nothing was built or run.

## Summary

The refactor is a faithful restructuring in almost every respect: the
contract-fetch, slot-fetch, env-read and coverage paths are behaviorally
identical, and `execTxWithCov`'s body is unchanged. Pulling the query handlers
out of the `case` and into a `handleQuery` function is an improvement — handlers
can no longer forget to resume execution, which is exactly the bug the old FFI
arm had.

There is **one real behavior change**, plus a handful of smaller concerns.

## The behavior change: empty-argument `ffi` now crashes the worker

`lib/Echidna/Exec.hs:199`

```haskell
handleQuery (PleaseDoFFI [] _ _) = error "Malformed FFI call"
```

### Before

`PleaseDoFFI []` did not match `Just (PleaseDoFFI (cmd : args) ...)` and fell
through the catch-all `_ -> pure vmResult`. That returned
`HandleEffect (Query …)` as the transaction result, which is a *supported* path:

- `getResult` maps it to the first-class `ErrorQuery` constructor
  (`lib/Echidna/Types/Tx.hs:244`)
- `handleErrorsAndConstruction` hits its `_ -> pure ()` arm
- the next `setupTx` resets `result = Nothing`
  (`lib/Echidna/Transaction.hs:172`)

The campaign kept going.

### After

`error` throws `ErrorCall`, which is caught at `lib/Echidna/UI.hs:248` as
`Crashed (show e)`. That worker dies.

### Reachability

hevm builds `cmd` directly from the decoded `string[]` with no non-empty check
(`hevm/src/EVM.hs:1913-1924`), so `vm.ffi(arr)` with a zero-length array
produces `PleaseDoFFI []`. This requires `allowFFI` to be on, but the array
length can easily be fuzzer-influenced — and this repo recently grew Foundry
test support, where `vm.ffi` is idiomatic. Net effect: a contract under test can
kill a fuzzing worker where it previously just recorded an `ErrorQuery`
transaction.

### Suggested fix

hevm's own oracle does `internalError` in this case
(`hevm/src/EVM/Fetch.hs:585`), so failing loudly is defensible — but a bare
`error` inside the exec loop is the wrong mechanism. Either keep
`pure vmResult` for this case, or raise it as an `ExecException` via
`throwM`/`vmExcept` so it is classified rather than stringified into `Crashed`.

## Non-blocking concerns

### `handleQuery` has no type signature and pattern-matches a GADT

`GADTs` implies `MonoLocalBinds`, so `handleQuery` should infer as
`Query Concrete -> m ()` from its use in `runFully`, and `Query Concrete` really
does have exactly these four constructors — `PleaseAskSMT` and `PleaseGetSols`
are `Query Symbolic` (`hevm/src/EVM/Types.hs:619-625`). Exhaustiveness should
therefore check out.

But CI compiles with `-Werror` (`.github/workflows/ci.yml:171`), so this rests
on GHC's coverage checker doing the GADT-index reasoning for an unannotated
local binding. An explicit signature (requires adding `forall m.` to
`execTxWith`) removes the doubt and documents intent. Worth confirming CI is
green before merging.

### Losing the catch-all trades a runtime fallback for a compile-time one

Any future `Query Concrete` constructor used to degrade gracefully to
`ErrorQuery`; now it becomes a build failure under `-Werror`. That is arguably
the better failure mode — noting it only because it is a deliberate change in
posture, not a no-op.

### Redundant constraint on `execCov`

`lib/Echidna/Exec.hs:253` adds `MonadThrow m`, but the body only uses `get`,
`put` and `liftIO`.

### Performance: probably a wash, but this is the hottest code in the fuzzer

`execCov` moved from a closed local binding with a single call site to a
top-level exported polymorphic function. The inner `loop` / `stepVM` /
`addCoverage` are still monomorphic `IO`, and the module builds with
`-O2 -fspecialize-aggressively -fexpose-all-unfoldings`, so the dictionary
indirection should land once per transaction rather than per opcode. Still worth
a before/after benchmark rather than an argument.

### `getRpcInfo` runs per query instead of once per `runFully`

Negligible — a reader `asks`, dwarfed by the RPC round-trip. Minor style point:
it uses `return` where the rest of the file uses `pure`.

### Dropped commented-out diagnostics

Two commented-out `logMsg` lines (the "Requested RPC but it is not configured"
ones) were removed. Trivial, but they were the only inline note that a missing
`rpcUrl` silently substitutes an empty account / zero slot. The `TODO`s survive.

## Verdict

Merge-able after two changes:

1. Make the empty-FFI case fail through the existing exception machinery instead
   of `error` (or restore the previous `pure vmResult` behavior).
2. Add an explicit type signature to `handleQuery`.

PR #1283, opened 2024-07-01
Show trace on UnknownFailure

### LINKED ISSUE #938: feat: more visibility for unsupported cheatcodes (BadCheatCode)
Using unrecognized cheatcodes results in a failure.

```solidity
import "forge-std/Test.sol";

contract TestContract is Test {
    function test123(uint256 x) public {
        vm.assume(false);
        vm.expectRevert();
    }
}
```

```
echidna-test . --contract TestContract --test-mode dapptest --format text --test-limit 100
Analyzing contract: /Users/lain/git/eth/template/test/TestContract.sol:TestContract
echidna-test: VM failed for unhandled reason, BadCheatCode (Just 1281615202). This shouldn't happen. Please file a ticket with this error message and steps to reproduce!
```

It would be great to know what cheatcode was executed (and perhaps where), as these can be difficult to find in larger codebases. Additionally, maybe unsupported cheatcodes could just be ignored (after printing an error message).

samalws-tob:
Partially resolves #938

### DIFF
--- a/lib/Echidna/Etheno.hs
+++ b/lib/Echidna/Etheno.hs
@@ -178,7 +178,7 @@ execEthenoTxs et = do
       (_        , AccountCreated _)  -> pure ()
       (Reversion,   _)               -> void $ put vm
       (HandleEffect (Query q), _)    -> crashWithQueryError q et
-      (VMFailure x, _)               -> vmExcept x >> M.fail "impossible"
+      (VMFailure x, _)               -> vmExcept Nothing x >> M.fail "impossible"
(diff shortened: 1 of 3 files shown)


ggrieco-tob:
Why this is partially resolving 938?

samalws-tob:
Because it doesn't explicitly provide a trace message of "you used an unknown cheatcode `vm.expectRevert`, which echidna doesn't support" but you can figure out what the unknown cheatcode is from the stack trace

ggrieco-tob:
This is good enough.

Issue #1252, opened 2024-05-10
update number of works to a decent default 


0xalpharush:
### Describe the desired feature

The current default is one but we should try to use all available CPU by default to maximize performance

ggrieco-tob:
This was upgraded to 4 (if cores are available). We can increase it, but using all the available CPU means that the user experience can be degraded if the developer is using short fuzzing campaign while they are developing new code.

gustavo-grieco:
Let's close this one for now, as we have a better number.

PR #1288, opened 2024-07-15
Raise default number of workers

elopez:
Now the default number of workers will be equal to the number of cores on the system, with a minimum of 1 and a maximum of 4.

### DIFF
--- a/lib/Echidna/Types/Campaign.hs
+++ b/lib/Echidna/Types/Campaign.hs
@@ -3,10 +3,10 @@ module Echidna.Types.Campaign where
 import Control.Concurrent (ThreadId)
 import Data.Aeson
 import Data.Map (Map)
-import Data.Maybe (fromMaybe)
 import Data.Text (Text)
 import Data.Text qualified as T
 import Data.Word (Word8, Word16)
+import GHC.Conc (numCapabilities)
 
 import Echidna.ABI (GenDict, emptyDict, encodeSig)
 import Echidna.Types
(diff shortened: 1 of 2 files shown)


lib/Echidna/Types/Campaign.hs
```diff
@@ -211,9 +211,14 @@ defaultSymExecAskSMTIters :: Integer
...
 
 -- | Get number of fuzzing workers (doesn't include sym exec worker)
--- Defaults to 1 if set to Nothing
+-- Defaults to (N-1) if set to Nothing, where N is Haskell's -N value,
+-- usually the number of cores, clamped between 1 and 3.
 getNFuzzWorkers :: CampaignConf -> Int
-getNFuzzWorkers conf = fromIntegral (fromMaybe 1 (conf.workers))
+getNFuzzWorkers conf = maybe defaultN fromIntegral conf.workers
+  where
+    n = numCapabilities
```
arcz, line 219: is it the same as `getNumCapabilities`?

lib/Echidna/Types/Campaign.hs
```diff
@@ -211,9 +211,14 @@ defaultSymExecAskSMTIters :: Integer
...
+    maxN = max 1 (n - 1) -- one dedicated thread for shrinking
```
arcz, line 220: we shrink with fuzz workers so this will not serve the purpose

lib/Echidna/Types/Campaign.hs
```diff
@@ -211,9 +211,14 @@ defaultSymExecAskSMTIters :: Integer
...
+    defaultN = min 3 maxN -- capped at 3 by default
```
arcz, line 221: why do we cap at 3?

lib/Echidna/Types/Campaign.hs
elopez: yes, but this is ["pure"](https://hackage.haskell.org/package/ghc-internal-9.1001.0/docs/src/GHC.Internal.Conc.Sync.html#numCapabilities), `getNumCapabilities` is IO

lib/Echidna/Types/Campaign.hs
elopez: @ggrieco-tob suggested the value

ggrieco-tob: it's a reasonable number for default, we can increase if we want later.

0xalpharush:
Does this close https://github.com/crytic/echidna/issues/1252

lib/Echidna/Types/Campaign.hs
elopez: I switched it to 4, considering I dropped the `-1`

Issue #1240, opened 2024-04-18
[Bug-Candidate]: no fore color support for terminal xterm-256color


aviggiano:
### Describe the issue:

In some cases, Echidna fails to run because of the error attached. I can't confirm, but I am under the impression that this happens when I minimize my terminal.

### Code example to reproduce the issue:

The codebase is private

### Version:

Echidna 2.2.3
Slither 0.9.3

### Relevant log output:

```shell
[I] ➜ echidna . --contract CryticTester --config echidna.yaml --test-mode assertion --corpus-dir echidna --test-limit 500000
[2024-04-18 18:25:18.86] Compiling .... itDone! (83.459169s)
Analyzing contract: CryticTester.sol:CryticTester
[2024-04-18 18:26:44.47] Running slither on .... Done! (126.154415s)
Loaded 23 transaction sequences from echidna/reproducers
Loaded 72 transaction sequences from echidna/coverage
echidna: no fore color support for terminal xterm-256color
CallStack (from HasCallStack):
  error, called at src/Graphics/Vty/Output/TerminfoBased.hs:119:39 in vty-5.38-IjxqNDNH7C1mXkr8nIcBw:Graphics.Vty.Output.TerminfoBased
```

ggrieco-tob:
@elopez any idea about this? :thinking:

elopez:
hm no idea here, it seems to be an error from vty. What OS / terminal emulator are you using @aviggiano? does it happen if you use a different terminal app?

If you can try to reproduce it with the binaries from [the GHC 9.6 PR](https://github.com/crytic/echidna/pull/1245) that would be nice as well; that branch uses a newer version of the UI stack (brick, vty) so it may fix things. You can get binaries from the CI run here: https://github.com/crytic/echidna/actions/runs/8915729894

elopez:
Please let us know if you see this happening once you upgrade to 2.2.5 😄

gustavo-grieco:
I haven't this in while, let's close it.

Issue #1135, opened 2023-10-31
[Bug-Candidate]: no back color support for terminal screen.xterm-256color


rappie:
### Describe the issue:

This happens very rarely at random moments. No way to reproduce.

Might be similar to the earlier "cups" error I kept getting in the past.

### Version:

Latest 2.2.1 `master` built with `nix-env`.

### Relevant log output:

```shell
echidna: no back color support for terminal screen.xterm-256color
CallStack (from HasCallStack):
  error, called at src/Graphics/Vty/Output/TerminfoBased.hs:127:32 in vty-5.38-Dbd4gS2zkOy8uP4PqnkPG0:Graphics.Vty.Output.TerminfoBased
```

rappie:
Possibly related:
```
echidna: user error (Terminal does not define required capability "sgr0")
```

ggrieco-tob:
Is this solved in 2.2.3?

rappie:
I have not had these TERM related errors for a while now since updating to the latest version

ggrieco-tob:
Great, let's close. Feel free to reopen if you see these again.

rappie:
```
echidna: user error (Terminal does not define required capability "clear")
```
Just had this one. Should we reopen it or create a new issue?

elopez:
@rappie if you're able to reproduce it, can you check if running `stty sane` or `tput reset` on the terminal help fix it?

rappie:
I've only had it once so far, but I regularly get a situation where CTRL+C is not working anymore after running Echidna. If I press it, the ^C character appears but it does not stop/cancel anything, it is just a character that I can remove with backspace.

To fix it, I do `reset`. I assume this is the same as `tput reset`?. I will try your commands next time.

rappie:
```
echidna: no back color support for terminal screen.xterm-256color
CallStack (from HasCallStack):
  error, called at src/Graphics/Vty/Output/TerminfoBased.hs:127:32 in vty-5.38-t4uBxudScr4gwaMP3IHWw:Graphics.Vty.Output.TerminfoBased
```

Just had the exact same error as mentioned in the original issue. It is very rare now though.

rappie:
I haven't had any issues with this in a very long time. Shall we close the issue?

gustavo-grieco:
Good idea, I was not able to reproduce this as well, probably some faulty dependency.

PR #1290, opened 2024-07-15
Support cross platform vty 6.x with vty-crossplatform

hellwolf:
This is a draft to make the patch visible during fixing [echidna](https://github.com/NixOS/nixpkgs/pull/327511) in nixpkgs.

I have not updated this repo's flake.nix and fix the build yet..

### DIFF
--- a/lib/Echidna/UI.hs
+++ b/lib/Echidna/UI.hs
@@ -8,7 +8,8 @@ import Brick
 import Brick.BChan
 import Brick.Widgets.Dialog qualified as B
 import Data.Sequence ((|>))
-import Graphics.Vty (Config, Event(..), Key(..), Modifier(..), defaultConfig, inputMap, mkVty)
+import Graphics.Vty (VtyUserConfig (..), Event(..), Key(..), Modifier(..))
+import Graphics.Vty.CrossPlatform (mkVty)
 import Graphics.Vty qualified as Vty
 import System.Posix
 import Echidna.UI.Widgets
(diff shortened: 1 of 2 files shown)


lib/Echidna/UI.hs
```diff
@@ -243,12 +244,12 @@ stopWorkers :: MonadIO m => [(ThreadId, a)] -> m ()
...
   forM_ workers $ \(threadId, _) -> liftIO $ killThread threadId
 
-vtyConfig :: IO Config
+vtyConfig :: IO VtyUserConfig
 vtyConfig = do
-  config <- Vty.standardIOConfig
-  pure config { inputMap = (Nothing, "\ESC[6;2~", EvKey KPageDown [MShift]) :
-                           (Nothing, "\ESC[5;2~", EvKey KPageUp [MShift]) :
-                           inputMap defaultConfig }
+  config <- Vty.userConfig
```
hellwolf, line 249: Note:

The standardIOConfig has some unix specific configurations. From 6.x, it is gone. The closest is a function from vty-unix package.

elopez:
Hi! to avoid duplicating efforts, please take a look at #1245 which also implements the newer vty

ggrieco-tob:
Uhm, I'm wondering if this will solve either #1135  or #1240 :thinking:

hellwolf:
> Hi! to avoid duplicating efforts, please take a look at #1245 which also implements the newer vty

Oh, I see! I will try the patch there and see if it passes the nixpkgs build, then I can abandon my own.

hellwolf:
I got echidna 2.2.3 fixed for nixpkgs.

I saw 2.2.4 release, but that requires a newer commit of hevm, which is not in nixpkgs yet.

ggrieco-tob:
is this still needed?

hellwolf:
As long as  the patch in

https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/tools/security/echidna/default.nix#L11

```
      url = "https://github.com/crytic/echidna/commit/2913b027d7e793390ed489ef6a47d23ec9b3c800.patch";
```

Is still valid?

PR #1300, opened 2024-08-26
Upgrade `hevm` to upstream `main`

elopez:


arcz:
Do we need to use our fork still? I checked and all our patches seem to be merged into hevm now

elopez:
We have the tstore/tload patches in our fork, once those get merged upstream we can switch back

PR #1302, opened 2024-08-30
Add support for unlimited tests

elopez:
This makes `--test-limit 0` run indefinitely, in line with medusa's behavior.


dguido:
@gustavo-grieco I think this looks good?

gustavo-grieco:
I think we should wait for the new CLI to be merged before allowing this. We need a specific mode to do replay and shrinking and using `test-limit == 0` was one workaround.

PR #1309, opened 2024-09-12
symExec support for tuple type (hevm PR 522)

samalws-tob:
This PR adds support for [HEVM PR 522](https://github.com/ethereum/hevm/pull/522), which is ahead of the version used as a dependency by echidna. PR 522 changes symAbiArg so that it supports tuples and has slightly different behavior around arrays. Once we bring our hevm dependency up to the latest version we need to apply the changes to SymExec made in this PR.


arcz:
@samalws-tob is this good to go now?

PR #1314, opened 2024-09-17
Add labels to call sequence

elopez:
hevm already supports the `vm.label(address, string)` cheatcode, but Echidna has not made use of the labels so far in its output. This adds some basic support when printing call sequences. When available, it will show the corresponding label next to any from/to address, or next to any address arguments.

image

Test contract:
```solidity
abstract contract Hevm {
    function label(address addr, string calldata label) external virtual;
}

contract C {
    constructor() public {
        Hevm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D).label(
            address(0x20000),
            "hacker"
        );
    }
    bool hacked;
    function hack() public {
        if (msg.sender == address(0x20000)) hacked = true;
    }

    function hack2(address x) public {
        if (msg.sender == address(0x10000) && x == address(0x20000))
            assert(!hacked);
    }
}
```

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -65,11 +65,11 @@ makeArrayAbiValues b =
      fmap (\n -> AbiBytes n . BS.append b $ BS.replicate (n - size) 0) [size..32]
 
 -- | Pretty-print some 'AbiValue'.
-ppAbiValue :: AbiValue -> String
-ppAbiValue = \case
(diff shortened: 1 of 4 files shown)


arcz (approved):
This is good, just please fix hlint stuff. For the long term we should re-use trace printing from hevm but this depends on https://github.com/ethereum/hevm/issues/496.

PR #1328, opened 2024-11-06
Print total calls as part of the stats on exit

### LINKED ISSUE #1322: Print total calls as part of the stats after closing Echidna

I track my total calls for reporting later on. It would be nice to have this in the terminal log instead of having to copy it from the TUI and potentially losing it when accidentally pressing ESC.

It could be part of these stats:
```
Unique instructions: 103968
Unique codehashes: 32
Corpus size: 27
Seed: 4563243713121521234
Total calls: 123456789
```

elopez:
Echidna will now print the number of calls done as part of the stats printed when exiting.

    Unique instructions: 287
    Unique codehashes: 1
    Corpus size: 1
    Seed: 830942768189866820
    Total calls: 10

Closes #1322


arcz (approved):
lgtm

PR #1350, opened 2025-02-10
Expanded the constant mining to get more nearby values

gustavo-grieco:


gustavo-grieco:
@elopez I'm wondering why this was never merged? 🤔

elopez (approved):

PR #1357, opened 2025-03-28
fix issue 1354

figtracer:


elopez:
Hi! Thanks for working on this! I see you moved the `tstore.sol`, but right now it is unused. Could you integrate it into the test runner so it is run as well? https://github.com/crytic/echidna/blob/master/src/test/Tests/Integration.hs

figtracer:
Thanks for the feedback! It should be integrated now.

elopez (approved):

PR #1369, opened 2025-05-02
Optimize `maybeLit{Byte,Word,Addr}Simp` and `maybeConcStoreSimp`

elopez:


elopez:
Merged upstream, replaced by #1346

PR #1373, opened 2025-05-09
Storable memory experiment

elopez:


elopez:
Merged upstream in hevm, closing in favor of #1346

PR #1376, opened 2025-06-03
HashMap storage experiment

elopez:
HashMap storage experiment


elopez:
Changes merged upstream, replaced with #1377

Issue #1224, opened 2024-03-20
[Bug-Candidate]: Echidna can get stuck while shrinking


rappie:
### Describe the issue:

In certain cases, Echidna can get stuck while shrinking.

Example below:
![image](https://github.com/crytic/echidna/assets/1430820/69e4b5c7-725d-43cf-b653-a2c9d06f7ef5)

When setting `maxBlockDelay` and `maxTimeDelay` to `1`, it is easily shrunk further:
![image](https://github.com/crytic/echidna/assets/1430820/48331f5f-ee25-440f-b37e-07271dde735b)

After further research I was able to determine that timestamps/blocks actually don't play a role in the breaking of this invariant at all. So it should have been able to shrink it further.

### Code example to reproduce the issue:

The code is private, I can share the repo if you DM me on Slack

### Version:

Echidna 2.2.3 (custom build from `master`)

ggrieco-tob:
I think we need to be more aggressive reducing the block numbers? In this case, this required at least a block each operation?

rappie:
No, there are no block/timestamp increments necessary at all to break this invariant.

Besides block/timestamp shrinking being slow or getting stuck, there are also uneccesary transactions that don't get shrunk away. It seems to me that there is some deeper issue going on.

ggrieco-tob:
@rappie can you please re-test the shrinking using the latest `master`?

rappie:
Still the same issue :disappointed:

elopez:
@rappie can you confirm if this is fixed on 2.2.7 or current master?

rappie:
Will check this soon 🫡

gustavo-grieco:
Any updates? 🤔

rappie:
Not yet sadly. I've been super busy but it is still on my todo list!

rappie:
@elopez @gustavo-grieco Tested this and it is indeed fixed 🎉

gustavo-grieco:
Awesome, let's close this one. Thanks to everyone involved to close this ticket!

PR #1385, opened 2025-06-20
Fix and reorganize shrinking

elopez:
The current shrinking implementation had a bug/limitation: it always tries to shrink the sender. In other words, if a transaction has a sender that is not minimal, Echidna will try to reduce it. However, the transaction might require a different sender to still cause the assertion failure, and the fact that Echidna will unconditionally try to lower it might cause the shrinking to get 'stuck'.

This commit reworks the shrinking process to try and favor a more incremental shrinking, with multiple passes of smaller changes. The old process consisted of:

 - 50% chance of removing one tx
 - 50% chance of (reducing sender _and_ shrinking) _all_ txs

The new process consists of, for each transaction independently:

 - 35% chance of shrinking the transaction
 - 30% chance of removing the transaction
 - 30% chance of keeping the transaction as-is
 -  5% chance of reducing sender

"wait" calls are treated differently:

 - 60% chance of reducing or removing delay
 - 40% chance of keeping it as-is

Fixes: #1224

### DIFF
--- a/lib/Echidna/Shrink.hs
+++ b/lib/Echidna/Shrink.hs
@@ -1,8 +1,7 @@
 module Echidna.Shrink (shrinkTest) where
 
-import Control.Monad ((<=<))
 import Control.Monad.Catch (MonadThrow)
-import Control.Monad.Random.Strict (MonadRandom, getRandomR, uniform)
+import Control.Monad.Random.Strict (MonadRandom, uniform, weighted)
 import Control.Monad.Reader.Class (MonadReader (ask), asks)
 import Control.Monad.State.Strict (MonadIO)
 import Control.Monad.ST (RealWorld)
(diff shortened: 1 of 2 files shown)


radubahmata:
Hi @elopez! Just checking in on this PR. If it’s still on hold, I can give it a shot and see if I can make progress. It would be helpful to have the contract and invariants to reproduce #1224, and to diff-check it before and after #1399.

elopez:
Hi @BowTiedRadone! We merged https://github.com/crytic/echidna/pull/1399 to fix https://github.com/crytic/echidna/issues/1224 so that's no longer an issue. On the other hand, this is a larger refactor that kind of changes the approach to shrinking and that I feel might lead to better results. But as it is a significant change, it requires more experimentation and testing to find a sweet spot for the different weights and such, and to make sure it doesn't regress shrinking performance. You're more than welcome to play with the weights on some real life shrinking problems and report your observations if you want 😄

PR #1389, opened 2025-06-26
Library auto-linking support through crytic-compile

elopez:


elopez:
CI is broken due to https://github.com/crytic/slither/issues/3042

PR #1392, opened 2025-07-07
Show gas/s metric in non-interactive mode

elopez:


lib/Echidna/UI.hs
```diff
@@ -380,17 +383,23 @@ isTerminal = hNowSupportsANSI stdout
 statusLine
   :: Env
   -> [WorkerState]
-  -> LocalTime  -- ^ Campaign start time
+  -> IORef (LocalTime, Int)  -- ^ Last update time and total gas
```
gustavo-grieco: Maybe a type of this will be good, otherwise, the PR looks good to me.

lib/Echidna/Transaction.hs
```diff
@@ -177,25 +179,49 @@ setupTx tx@Tx{call} = fromEVM $ do
...
+      SolCalldata cd -> cd
 
 advanceBlock :: Block -> (W256, W256) -> Block
 advanceBlock blk (t,b) =
   blk { timestamp = Lit (forceWord blk.timestamp + t)
       , number = Lit (forceWord blk.number + b) }
+
+-- | Calculate transaction gas cost for Echidna Tx
+-- Adapted from HEVM's txGasCost function
+txGasCost :: FeeSchedule Gas -> Bool -> BS.ByteString -> Gas
```
gustavo-grieco, line 218: This function should be moved to hevm I guess, but let´s keep it here for this PR and move it later.

gustavo-grieco (approved):

PR #1394, opened 2025-07-08
Reworked symbolic execution code

gustavo-grieco:
This PR brings a good number of important changes to the symbolic execution code:

* It removes the concolic execution: this code should be moved to hevm directly, since it is too hard to maintain from our side, and we don´t know exactly when it is useful. 
* It adds a new type of test result: unsolvable. This new value is used to signal that a symbolic exploration finished to review all possible states without any counter example to trigger an assertion failure.
* It improves the full symbolic execution of a single transaction, as this seems to be the best approach to explore the code using symbolic execution, some notable changes:
   + A corpus is shuffled before selecting a sequence for the symbolic exploration
   + Bitwulza is the new default smt solver, but other solvers can be selected
   + Support for handling symbolic block.number and block.timestamp
   + Support for avoiding doing symbolic exploration for useless (e.g. no arguments) or unsupported inputs (e.g. dynamic arguments)
   + Two modes for using symbolic execution: verification mode (stateless) and exploration (fuzzing+sym-exec)
   + Two strategies are used for introducing symbolic transactions in the exploration mode: (1) a random transaction is inserted at a random position and (2) the last transaction of a sequence is symbolized.
   + Support for RPC query when doing symbolic exploration (including cache that will be queried but not updated)
   + Added some more tests for symbolic execution (verification mode)  
   + The symbolic execution mode is moved into its own directory

Some unrelated changes included in this PR:

* Removed etheno related code (obsolete)
* Removed tests related with gas usage (covered by hevm per opcode) and self-destruction (no longer relevant in the current fork)
* Removed usage of very old solc version in the tests

Known issues:
* UI needs to be adapted in verification mode since the symbolic worker is busy and it can block the refresh of the UI. Text mode is advised.
* There is a detection to some cases where hevm produces a counter example, but it is not reproduced correctly. This is a bug that will be eventually solved upstream, but if we manage to detect a counter example that does not increase coverage (e.g. reached an assertion), echidna will trigger a failure so it can be reported upstream. This detection is not perfect, and there cases where it does not properly work.


gustavo-grieco:
It seems that the latest merge broke it: https://github.com/crytic/echidna/actions/runs/16471079633/job/46560187173?pr=1394#step:14:34 😅

lib/Echidna/Campaign.hs
```diff
@@ -108,20 +116,23 @@ runSymWorker
   -- ^ Initial corpus of transactions
   -> Maybe Text -- ^ Specified contract name
   -> m (WorkerStopReason, WorkerState)
-runSymWorker callback vm dict workerId initialCorpus name = do
+runSymWorker callback vm dict workerId _ name = do
```
elopez, line 119: if this argument is not used, then let's delete it and adjust runWorker to not pass it

lib/Echidna/Campaign.hs
```diff
@@ -108,20 +116,23 @@ runSymWorker
...
+    flip runStateT initialState $
+      flip evalRandT (mkStdGen effectiveSeed) $ do -- unused but needed for callseq
+        verifyMethods -- No arguments, everything is in this environment
+        pure SymbolicVerificationDone
+  else
+    flip runStateT initialState $
+      flip evalRandT (mkStdGen effectiveSeed) $ do -- unused but needed for callseq
+        lift callback
+        listenerLoop listenerFunc chan nworkers
+        pure SymbolicExplorationDone
```
elopez: can we invert the conditionals and the flips to reduce duplication?

lib/Echidna/Types/Campaign.hs
```diff
@@ -43,10 +41,12 @@ data CampaignConf = CampaignConf
     -- ^ Server-Sent Events HTTP port number, if missing server is not ran
   , symExec            :: Bool
     -- ^ Whether to add an additional symbolic execution worker
-  , symExecConcolic    :: Bool
-    -- ^ Whether symbolic execution will be concolic (vs full symbolic execution)
-    -- Only relevant if symExec is True
+  , symExecSMTSolver   :: Solver
+    -- ^ SMT solver to use for symbolic execution.
+    -- Supported solvers: "cvc5", "z3", "yices", "boolector"
```
elopez: this is inaccurate, right?

```suggestion
    -- Supported solvers: "cvc5", "z3", "bitwuzla"
```

src/test/Tests/Integration.hs
```diff
@@ -60,8 +60,6 @@ integrationTests = testGroup "Solidity Integration Testing"
       [ ("echidna_library_call failed",            solved      "echidna_library_call")
       , ("echidna_valid_timestamp failed",         passed      "echidna_valid_timestamp")
       ]
-  , testContractV "basic/fallback.sol"   (Just (< solcV (0,6,0))) Nothing
-      [ ("echidna_fallback failed",                solved      "echidna_fallback") ]
```
elopez: this test looks useful, let's keep it

tests/solidity/basic/default.yaml
```diff
@@ -95,21 +95,22 @@ workers: null
...
+# only relevant if symExec is true
 symExecAskSMTIters: 1
-# List of whitelisted functions for using symbolic/concolic exploration
+# List of whitelisted functions for using symbolic exploration
 # only relevant if symExec is true
 symExecTargets: null
+# List the max number of states to explore when using symbolc execution
+# only relevant if symExec is true
+symExecMaxExplore: 10
+symExecSMTSolver: bitwuzla
```
elopez, line 118: Can you add a comment for symExecSMTSolver explaining what it does, valid values?

lib/Echidna/Campaign.hs
gustavo-grieco: I reduced the duplication moving the `flip` outside of the if. Why you want to invert the conditional?

tests/solidity/basic/default.yaml
gustavo-grieco: Done

src/test/Tests/Integration.hs
gustavo-grieco: This test is only valid when solc versions is less than 0.6.x. All the old solc versions tests were disabled (and removed) in this PR. If I enabled this, it will be skipped 100% of the time.

lib/Echidna/Campaign.hs
elopez: yeah, that's what I meant, sorry if I wasn't clear 😄

gustavo-grieco: Done!

lib/Echidna/Campaign.hs
gustavo-grieco: I suspect these initial corpus will be used sooner than later, that´s why I prefer to keep it for now

lib/Echidna/Types/Campaign.hs
gustavo-grieco: Fixed

src/test/Tests/Integration.hs
```diff
@@ -86,9 +86,7 @@ integrationTests = testGroup "Solidity Integration Testing"
   , testContract "basic/gaslimit.sol"  Nothing
       [ ("echidna_gaslimit passed",                passed      "echidna_gaslimit") ]
   , testContract "basic/gasleft.sol"     (Just "basic/gasleft.yaml")
-      [ ("unexpected gas left",                   passed      "echidna_expected_gasleft") ]
-  ,  testContractV "basic/killed.sol"      (Just (< solcV (0,8,0))) (Just "basic/killed.yaml")
```
elopez, line 90: Can you remove `killed.sol` and `killed.yaml` as well?

lib/Echidna/Campaign.hs
```diff
@@ -18,38 +18,46 @@ import Data.ByteString.Lazy qualified as LBS
...
-import Data.Maybe (isJust, mapMaybe)
+import Data.Maybe (isJust, mapMaybe, fromJust)
 import Data.Set (Set)
 import Data.Set qualified as Set
-import Data.Text (Text)
+import Data.Text (Text, unpack)
 import Data.Time (LocalTime)
 import System.Random (mkStdGen)
 
+
```
elopez:
```suggestion
```

lib/Echidna/Server.hs
```diff
@@ -52,7 +53,8 @@ runSSEServer serverStopVar env port nworkers = do
                     TestFalsified _ -> "test_falsified"
                     TestOptimized _ -> "test_optimized"
                     NewCoverage {} -> "new_coverage"
-                    SymNoNewCoverage -> "sym_no_new_coverage"
+                    SymExecLog _ -> "sym_tx_verified"
```
elopez:
```suggestion
                    SymExecLog _ -> "sym_exec_log"
```

lib/Echidna/Types/Cache.hs
```diff
@@ -0,0 +1,7 @@
+module Echidna.Types.Cache where
+
+import Data.Map (Map)
+import EVM.Types (W256, Addr, Contract)
+
+type ContractCache = Map Addr (Maybe Contract)
+type SlotCache = Map Addr (Map W256 (Maybe W256))
```
elopez: Can you use these new types in Types/Config.hs:Env definition?

gustavo-grieco: Done

src/test/Tests/Integration.hs
gustavo-grieco: Done!

PR #1399, opened 2025-07-11
Allow address to stay the same during shrinking

elopez:
The current shrinking implementation has a bug/limitation: it always tries to shrink the sender. In other words, if a transaction has a sender that is not minimal, Echidna will try to reduce it. However, the transaction might require a different sender to still cause the assertion failure, and the fact that Echidna will unconditionally try to lower it might cause the shrinking to get 'stuck'.

This allows the sender shrinker to keep the transaction source the same with some probability, allowing the shrinking to proceed in such cases.


elopez:
This is a lighter version of #1385 to unblock users while we discuss the larger refactoring.

PR #1404, opened 2025-07-23
fix: reduce html class names to their first letters

### LINKED ISSUE #1337: Reduce HTML coverage size by compressing class names

From looking at:
https://github.com/crytic/echidna/blob/1ef14f02219dd2585a9a5bc2d1339496d5becf95/lib/Echidna/Output/Source.hs#L87

And html reports it seems that echidna uses only 4 classes:
.executed
.neutral
.reverted
.unexecuted

These classes are repeated on each line, contributing to file-size for close to zero advantage

You could rename each class to use it's first letter, this would save file size (and may make the process of generating the report faster)

divyaranjan1905:
Fixes #1337 

I didn't find a way to test this, so please tell me how I can ensure that indeed this is working as expected. Though since it's mostly a few strings changed, it should just work as intended.

Regards,

Divya.

### DIFF
--- a/lib/Echidna/Output/Source.hs
+++ b/lib/Echidna/Output/Source.hs
@@ -85,10 +85,10 @@ ppCoveredCode fileType sc cs s | null s = "Coverage map is empty"
     topHeader = case fileType of
       Lcov -> "TN:\n"
       Html -> "<style> code { white-space: pre-wrap; display: block; background-color: #eee; }" <>
-              ".e { background-color: #afa; }" <>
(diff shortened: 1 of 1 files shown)


elopez:
hi, thanks for the contribution! could we add something (a few comments maybe) to keep the meaning that corresponds to each class visible?

gustavo-grieco (approved):

divyaranjan1905:
Hello @elopez , just added some comments. Hopefully that works :)

elopez (approved):
Thanks!

PR #1405, opened 2025-07-24
Store event changes into Echidna's corpus

### LINKED ISSUE #883: Feature Request: Add values from state changes and/or events to corpus
**Feature**
It would be nice to have echidna add values from state changes and/or emitted events to the corpus.

**State changes**
A simple example of state changes would be an ERC20 transfer. The new balance(s) might be interesting numbers to find edge cases.

image
https://etherscan.io/tx/0x483b6382644c114dba9268fa8a45a8c8921666043ea75ae2fc6e09381a5363f2#statechange

**Events**
An example for events would also be transfering ERC20 or ERC721 tokens. Here you have the value transfered or the id of the NFT.

In my opinion many events hold a treasure trove of interesting numbers. For example:
```solidity
event ConditionCreated(
	uint256 indexed oracleConditionId,
	uint256 indexed conditionId,
	uint64 timestamp
);
event ConditionResolved(
	uint256 indexed oracleConditionId,
	uint256 indexed conditionId,
	uint64 outcomeWin,
	uint8 state,
	uint256 amountForLp
);
```
https://github.com/Azuro-protocol/Bug-Bounty/blob/main/contracts/interface/ICore.sol#L37

divyaranjan1905:
Fixes #883. Only implements the storing of events, not of the entire state's variables etc.

Regards,

Divya.


lib/Echidna/Campaign.hs
```diff
@@ -366,8 +367,14 @@ callseq vm txSeq = do
       diffs = Map.fromList [(AbiAddressType, Set.fromList $ AbiAddress . forceAddr <$> newAddrs)]
       -- Now we try to parse the return values as solidity constants, and add them to 'GenDict'
       resultMap = returnValues results workerState.genDict.rTypes
+      -- compute the new events to be stored
+      eventDiffs = extractEventValues vm.dappInfo results
```
divyaranjan1905: I'm not sure if this is how it should be done with regards to the `dappInfo`, since there's `vm.dapp` and `vm.dappInfo`. Where can I find more information about them?

lib/Echidna/Events.hs
```diff
@@ -72,6 +78,35 @@ extractEvents decodeErrors dappInfo vm =
         Just $ humanPanic $ decodePanic d
       _ -> Nothing
 
+-- | Extract all non‑indexed event values emitted between two VM states.
+extractEventValues :: DappInfo -> VM Concrete s -> VM Concrete s -> Map AbiType (Set AbiValue)
+extractEventValues dappInfo vm vm' =
+  let
+    oldLogs = vm.logs
```
gustavo-grieco, line 83: I think logs are cleaned from one van to the next one 🤔

gustavo-grieco:
Please provide a small solidity code that shows an assertion failure that needs this new feature. We will show you how to add a test

divyaranjan1905:
> Please provide a small solidity code that shows an assertion failure that needs this new feature. We will show you how to add a test

Something like this?

```solidity
contract EventAssert {
    event ValueSet(uint256 indexed value);

    function setValue(uint256 value) public {
        emit ValueSet(value);
        assert(value != 1295);
    }
}
```

gustavo-grieco:
Your example needs to have an event emmited with a value that can be used to trigger the assertion, but it needs to be essential for that. Usually you need to do some computation and emit some value from it, that is normally hidden from the fuzzer.

divyaranjan1905:
How about this @gustavo-grieco?
```solidity
contract EventAssert {
    uint private secret = 123;
    event Output(uint result);

    function check(uint x) public {
        uint result = x % 200;
        emit Output(result);
        if (result == secret) {
            assert(false);
        }
    }
}
```

gustavo-grieco:
Uhm it looks bette but I'm not sure the value from the event is useful to break the assertion. Try using a hashed value instead 🤔

divyaranjan1905:
So, something like this:

```solidity
contract EventAssert {
    uint private secret;

    event Computed(uint result);

    constructor() {
        secret = uint(keccak256(abi.encodePacked(block.timestamp))) % 1000;
    }

    function check(uint x) public {
        uint result = (x * 3 + 7) % 1000;
        emit Computed(result);
        if (result == secret) {
            assert(false);
        }
    }
}
```

gustavo-grieco:
So if you remove `emit Computed(result)`, echidna will not able to find the value to break it, right? (I haven't tested it)

divyaranjan1905:
> So if you remove `emit Computed(result)`, echidna will not able to find the value to break it, right? (I haven't tested it)

I think so, because if we don't `emit`, Echidna has no way of knowing about it.

Let me know how to go further with the testing.

divyaranjan1905:
Changed the contract following @elopez's suggestion:

```solidity
contract EventAssert {
    uint private secret;
    event Secret(uint result);
    function reset() public {
        secret = uint(keccak256(abi.encodePacked(block.timestamp)));
        emit Secret(secret);
    }
    function check(uint x) public {
        if (x == secret) {
            assert(false);
        }
    }
}
```

divyaranjan1905:
This is the result:

image


On the left is upstream Echidna without the changes, and on the right is my fork:

gustavo-grieco:
This looks good, let's add it as a test in the "value" section

divyaranjan1905:
The fork doesn't reproduce the revert deterministically though, but I guess that's expected because of the hash being quite a huge one.

gustavo-grieco:
You mean that this test is not reliable ?

divyaranjan1905:
> You mean that this test is not reliable ?

I wasn't sure if that's expected ir not. It only gives the output I sent a few times and at other times it doesn't revert and says everything passes.

lib/Echidna/Events.hs
divyaranjan1905: Isn't that how it should be?

tests/solidity/values/campaign.yaml
```diff
@@ -0,0 +1,2 @@
+testLimit: 100000
```
gustavo-grieco: Reduce the testLimit, do we really need these amount now?

tests/solidity/values/campaign.sol
gustavo-grieco, line 1: change the name to events.sol, so it is easy to understand what kind of value we are trying to get from the filename

lib/Echidna/Campaign.hs
```diff
@@ -366,8 +367,18 @@ callseq vm txSeq = do
       diffs = Map.fromList [(AbiAddressType, Set.fromList $ AbiAddress . forceAddr <$> newAddrs)]
       -- Now we try to parse the return values as solidity constants, and add them to 'GenDict'
       resultMap = returnValues results workerState.genDict.rTypes
+      -- compute the new events to be stored
+      eventDiffs = extractEventValues env.dapp vm vm'
       -- union the return results with the new addresses
-      additions = Map.unionWith Set.union diffs resultMap
+      additions = 
+        -- fold all of the individual maps into one
+        List.foldl' (Map.unionWith Set.union)
```
elopez: I think using `Map.unionsWith` would make the code a bit nicer to read. Also, according to the documentation, it's preferable to do `bigMap union smallMap`, so let's order these in descending size (likely resultsMap and eventDiffs first, then diffs (always 1 map key)).

lib/Echidna/Campaign.hs
```diff
@@ -366,8 +367,18 @@ callseq vm txSeq = do
...
+               Map.empty
+               [ diffs
+               , resultMap
+               , eventDiffs
+               ]
+
       -- append to the constants dictionary
       updatedDict = workerState.genDict
         { constants = Map.unionWith Set.union additions workerState.genDict.constants
```
elopez: we should probably swap the arguments here for performance as well, the accumulated constants map is likely to be bigger most of the time.

```suggestion
        { constants = Map.unionWith Set.union workerState.genDict.constants additions
```

lib/Echidna/Campaign.hs
divyaranjan1905: Indeed, that should've occurred to me. `unionsWith` is certainly more elegant to read. Didn't know about the ordering efficiency, thanks!

lib/Echidna/Campaign.hs
divyaranjan1905: Fixed.

gustavo-grieco (approved):

PR #1406, opened 2025-07-26
Add values from tuple elements into the dictionary

gustavo-grieco:
This should allow echidna to solve this:

```solidity
pragma solidity ^0.8.0;

contract Marketplace {
    uint256 internal currentKey = 1375566;
    struct Item {
        uint16 useless;
        uint256 keyId;
        bytes32 data;
    }

    mapping(uint256 => Item) public items;

    function addItem(bytes32 _data) public returns (Item memory) {
    currentKey += 174; 

        // Store the item
        items[currentKey] = Item({
            useless: 16,
            keyId: currentKey,
            data: _data
        });
    return items[currentKey];
    }

    function getItem(uint256 _keyId) public view returns (Item memory) {
        assert(items[_keyId].keyId == 0);
    }
}
```

We need to add this into a test.

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -25,10 +25,11 @@ import Data.Set (Set)
 import Data.Set qualified as Set
 import Data.Text (Text)
 import Data.Time (LocalTime)
+import Data.Vector qualified as V
 import System.Random (mkStdGen)
 
 import EVM (cheatCode)
-import EVM.ABI (getAbi, AbiType(AbiAddressType), AbiValue(AbiAddress))
+import EVM.ABI (getAbi, AbiType(AbiAddressType, AbiTupleType), AbiValue(AbiAddress, AbiTuple), abiValueType)
 import EVM.Dapp (DappInfo(..))
 import EVM.Types hiding (Env, Frame(state), Gas)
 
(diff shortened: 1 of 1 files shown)


tests/solidity/values/struct.sol
```diff
@@ -0,0 +1,30 @@
...
+        // Store the item
+        items[currentKey] = Item({
+	    nonsense: 43,
+            keyId: currentKey,
+            data: _data
+        });
+	return items[currentKey];
+    }
+
+    function getItem(uint256 _keyId) public view returns (Item memory) {
```
elopez:
```suggestion
    function getItem(uint256 _keyId) public view {
```

this does not seem to return anything

tests/solidity/values/struct.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
elopez:
```suggestion
pragma experimental ABIEncoderV2;
```

this should get the tests running

lib/Echidna/Campaign.hs
```diff
@@ -395,20 +396,32 @@ callseq vm txSeq = do
...
           type' <- returnTypeOf fname
           case runGetOrFail (getAbi type') (LBS.fromStrict buf) of
-            -- make sure we don't use cheat codes to form fuzzing call sequences
-            Right (_, _, abiValue) | abiValue /= AbiAddress (forceAddr cheatCode) ->
-              Just (type', Set.singleton abiValue)
+            Right (_, _, abiValue) ->
+              if isTuple type'
+                then Just $ Map.fromListWith Set.union
+                      [ (abiValueType val, Set.singleton val)
+                      | val <- V.toList $ getTupleVector abiValue
```
elopez: should these be filtered so no `cheatCode` makes its way into the dictionary?

gustavo-grieco: Let me fix that

gustavo-grieco: Should we move the cheat code detection deeper into the dictionary handling to avoid duplication?

gustavo-grieco: This is done, even at the cost of duplicating the logic a bit

elopez (approved):

PR #1420, opened 2025-08-15
Enable onlyDeployed and updated hevm

gustavo-grieco:
Echidna needs to enabled the new `onlyDeployed` option in order to get correct results when an address is symbolic (and it has some constraints). However, there is some issue compared to hevm, and we currently don´t know why it works in a different way: https://github.com/ethereum/hevm/issues/828. Once that's solved, a unit case should be added here and this is ready to go.


gustavo-grieco:
No longer need, replaced by #1441

PR #1423, opened 2025-08-20
[RFC] MCP support during a fuzzing campaign

gustavo-grieco:
This is an experimental PR to add MCP commands in Echidna. These commands allow LLMs to query the state of fuzzing campaign, and perhaps make adjustments to guide or direct it.

Proposed commands to implement:
- [x] Show number of corpus elements
- [x] Show transactions in the corpus (using sequence number and pages of transactions)
- [x] Search transactions in the corpus
- [ ] Insert element in the initial dictionary 
- [ ] Change mutation weights
- [ ] Inspect coverage of a function
- [ ] Prioritize function  

As expected, these commands can be used directly by LLM using a number of MCP clients:

image

Feel free to comment with more suggestions.

Note that this branch will not compile until we fork `haskell-mcp-server` (or use another similar library supported in stack/nix).

### DIFF
--- a/lib/Echidna/MCP.hs
+++ b/lib/Echidna/MCP.hs
@@ -1,29 +1,78 @@
 module Echidna.MCP (runMCPServer) where
 
 import Control.Monad.IO.Class (liftIO)
+import Data.Aeson (encode)
 import Data.IORef (readIORef)
-import Data.Text (pack)
+import Data.List (find, isInfixOf)
+import Data.Maybe (fromMaybe)
+import Data.Set (toList)
+import Data.Text (pack, unpack)
+import Data.Text.Lazy (toStrict)
+import Data.Text.Lazy.Encoding (decodeUtf8)
 import MCP.Server
 import Network.Wai.Handler.Warp (Port)
 
 import Echidna.Types.Config (Env(..))
 import Echidna.Types.Corpus (corpusSize)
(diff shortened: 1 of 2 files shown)


aviggiano:
Is it possible to extend the server to support the following additional commands?

- Save the current corpus/reproducers

This would be useful when we want to manually inspect coverage. Also, if we need to feed the output of one fuzzer into another tool. For example, imagine an LLM orchestrating parallel runs between Echidna and Medusa, exchanging interesting inputs between them. I guess using the disk for I/O is not a pre-requisite, but could be easier to do depending on the other tool's architecture, without making too many assumptions into how they operate.

- Stop shrinking

This is mostly a convenience feature to allow the fuzzer to resume exploration without spending cycles on shrinking. I don't have a strong practical use case: usually if I want this behavior, I simply stop the fuzzer and restart it later—but given that adding new commands is likely far simpler than building the server itself, it seems reasonable to include it as an extra tool.

gustavo-grieco:
Yes, these are good candidates. This feature is blocked until we rewrite some central part of the campaign (I'm writing an issue describing what exactly)

gustavo-grieco:
Blocked by https://github.com/crytic/echidna/issues/1490

gustavo-grieco:
Replaced by #1502

PR #1428, opened 2025-08-20
Add independent coverage directory configuration option

BowTiedRadone:
This PR adds a new `coverageDir` configuration option that allows users to specify a separate directory for saving coverage reports, independent from the corpus directory.

Previously, coverage reports were saved to the corpus directory, which mixed different types of output files. This PR introduces a dedicated configuration option with automatic fallback to maintain backward compatibility.

**Usage Example**
```yaml
corpusDir: "corpus-xyz"
coverageDir: "coverage-xyz"
```

### DIFF
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -101,7 +101,7 @@ main = withUtf8 $ withCP65001 $ do
         corpus <- readIORef env.corpusRef
         saveTxs env (dir </> "coverage") (snd <$> Set.toList corpus)
 
-  -- save coverage reports (now using separate directory)
+  -- save coverage reports
   let coverageDir = cfg.campaignConf.coverageDir <|> cfg.campaignConf.corpusDir
   case coverageDir of
     Nothing -> pure ()
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -101,20 +101,21 @@ main = withUtf8 $ withCP65001 $ do
         corpus <- readIORef env.corpusRef
         saveTxs env (dir </> "coverage") (snd <$> Set.toList corpus)
 
-      -- TODO: We use the corpus dir to save coverage reports which is confusing.
-      -- Add config option to pass dir for saving coverage report and decouple it
-      -- from corpusDir.
-      unless (null cfg.campaignConf.coverageFormats) $ measureIO cfg.solConf.quiet "Saving coverage" $ do
-        -- We need runId to have a unique directory to save files under so they
-        -- don't collide with the next runs. We use the current time for this
-        -- as it orders the runs chronologically.
-        runId <- fromIntegral . systemSeconds <$> getSystemTime
+  -- save coverage reports (now using separate directory)
+  let coverageDir = cfg.campaignConf.coverageDir <|> cfg.campaignConf.corpusDir
+  case coverageDir of
+    Nothing -> pure ()
+    Just dir -> unless (null cfg.campaignConf.coverageFormats) $ measureIO cfg.solConf.quiet "Saving coverage" $ do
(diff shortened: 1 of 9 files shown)


radubahmata:
@arcz Could you please take a look at this when you have a chance? Let me know if any further action is needed.

gustavo-grieco:
I think the code looks fine in general but I'm wondering about the use case. Why exactly you need the fuzzer to write the reports elsewhere?

radubahmata:
First of all, this PR addresses the TODO in the codebase: https://github.com/crytic/echidna/blob/9870b1ac0e47bc251afeccbe73014e1eb43a87d5/src/Main.hs#L104-L106

It separates coverage reports from the corpus directory, which makes things clearer since they serve different purposes. It also gives users more flexibility—for example, someone tracking coverage over time in version control can now keep those reports cleanly organized by choosing a dedicated path.

src/Main.hs
```diff
@@ -101,20 +101,21 @@ main = withUtf8 $ withCP65001 $ do
...
 
-      -- TODO: We use the corpus dir to save coverage reports which is confusing.
-      -- Add config option to pass dir for saving coverage report and decouple it
-      -- from corpusDir.
-      unless (null cfg.campaignConf.coverageFormats) $ measureIO cfg.solConf.quiet "Saving coverage" $ do
-        -- We need runId to have a unique directory to save files under so they
-        -- don't collide with the next runs. We use the current time for this
-        -- as it orders the runs chronologically.
-        runId <- fromIntegral . systemSeconds <$> getSystemTime
+  -- save coverage reports (now using separate directory)
```
elopez:
```suggestion
  -- save coverage reports
```

src/test/Common.hs
```diff
@@ -249,3 +252,18 @@ countCorpus :: Int -> (Env, WorkerState) -> IO Bool
...
+  -- verify configuration: corpusDir set, coverageDir not set
+  case (env.cfg.campaignConf.corpusDir, env.cfg.campaignConf.coverageDir) of
+    (Just corpusDir, Nothing) -> pure $ corpusDir == expectedCorpusDir
+    _ -> pure False
+
+-- | Check that the effective coverage directory matches the expected path.
+-- This tests the actual fallback logic: coverageDir <|> corpusDir.
+checkEffectiveCoverageDir :: FilePath -> (Env, WorkerState) -> IO Bool
+checkEffectiveCoverageDir expectedDir (env, _) = do
+  let effectiveDir = env.cfg.campaignConf.coverageDir <|> env.cfg.campaignConf.corpusDir
```
elopez: This kind of duplicates the same logic you're trying to test, I don't think it adds much value. I think asserting the expected state is enough 👍

elopez (commented):
Thanks for the PR!

src/test/Common.hs
radubahmata: Addressed in 7b37d96.

gustavo-grieco:
Oh, btw, I noted that you are first time contributor. Would you like some suggestions on other small issues to address so you can get more exposure to the code?

radubahmata:
@gustavo-grieco Thanks, I’d be glad to help when time permits. I’m very interested in fuzzing tools.

In fact, together with @moodmosaic, I built a similar tool for the Stacks blockchain called [Rendezvous](https://github.com/stacks-network/rendezvous). In its [Foreword](https://stacks-network.github.io/rendezvous/chapter_1.html), we made sure to credit Echidna 🙂

gustavo-grieco:
Merged! It is great to see people inspired by our code! BTW, if you want to discuss some next steps or potential collaboration hit me on Telegram

PR #1431, opened 2025-08-24
Only verify suitable methods

gustavo-grieco:
This PR adds a missing feature from the new symbolic execution capabilities: run some basic checks to verify functions to make sense to verify. In verification mode, we excluded:

* Functions without arguments (e.g. unit tests)
* Function without assert detected. If no asserts are detected, all the methods are considered valid.
* Functions that have dynamic data structures are parameters.

### DIFF
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -44,7 +44,7 @@ import Echidna.Solidity (chooseContract)
 import Echidna.SymExec.Common (extractTxs, extractErrors)
 import Echidna.SymExec.Symbolic (forceAddr)
 import Echidna.SymExec.Exploration (exploreContract, getTargetMethodFromTx, getRandomTargetMethod)
-import Echidna.SymExec.Verification (verifyMethod)
+import Echidna.SymExec.Verification (verifyMethod, isSuitableToVerifyMethod)
 import Echidna.Test
 import Echidna.Transaction
 import Echidna.Types.Campaign
@@ -274,7 +274,12 @@ runSymWorker callback vm dict workerId _ name = do
     let cs = Map.elems dapp.solcByName
     contract <- chooseContract cs name
     let allMethods = contract.abiMap
-    mapM_ (symExecMethod contract) allMethods
+    mapM_ (\method -> do
+           isSuitable <- isSuitableToVerifyMethod contract method
+           if isSuitable 
+            then (symExecMethod contract method) 
+            else pushWorkerEvent $ SymExecError ("Skipped verification of method " <> unpack method.methodSignature)
+          ) allMethods
 
   symExecMethod contract method = do
     lift callback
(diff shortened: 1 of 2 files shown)


lib/Echidna/Campaign.hs
```diff
@@ -274,7 +273,13 @@ runSymWorker callback vm dict workerId _ name = do
     let cs = Map.elems dapp.solcByName
     contract <- chooseContract cs name
     let allMethods = contract.abiMap
-    mapM_ (symExecMethod contract) allMethods
+    conf <- asks (.cfg)
+    mapM_ (\method -> do
```
elopez: nit: forM_ is probably easier to follow

lib/Echidna/Campaign.hs
```diff
@@ -274,7 +273,13 @@ runSymWorker callback vm dict workerId _ name = do
...
+           isSuitable <- isSuitableToVerifyMethod contract method conf.campaignConf.symExecTargets
+           if isSuitable
+            then symExecMethod contract method
+            else pushWorkerEvent $ SymExecError ("Skipped verification of method " <> unpack method.methodSignature)
```
elopez, line 281: should this be an error?

lib/Echidna/SymExec/Verification.hs
```diff
@@ -19,16 +21,32 @@ import Control.Monad.ST (RealWorld)
...
 import Echidna.Worker (pushWorkerEvent)
 
+isSuitableToVerifyMethod :: (MonadIO m, MonadReader Echidna.Types.Config.Env m) => SolcContract -> Method -> [Text.Text] -> m Bool
+isSuitableToVerifyMethod contract method symExecTargets = do
+  env <- ask
+  let allMethods = assocs contract.abiMap
+      assertSigs = env.world.assertSigs
+      (selector, _) = case filter (\(_, m) -> m == method) allMethods of
+        [] -> error $ "Method " ++ show method.name ++ " not found in contract ABI"
+        (x:_) -> x
```
elopez: something using find is probably more readable, like

```suggestion
      (selector, _) = case find (\(_, m) -> m == method) allMethods of
        Nothing -> error $ "Method " ++ show method.name ++ " not found in contract ABI"
        Just x -> x
```

or some variation with maybe

`maybe (error ".....") snd $ find (snd >>> (== method)) allMethods`

lib/Echidna/Campaign.hs
gustavo-grieco: No, verification skipped means that it cannot verify it or the user wants to verify something else

lib/Echidna/Campaign.hs
gustavo-grieco: Done

lib/Echidna/SymExec/Verification.hs
gustavo-grieco: Done. I avoided using a one liner, since the error message is already long.

PR #1433, opened 2025-08-28
docs: fix incorrect Docker volume mount command

zkpepe:
noticed that the provided Docker commands don’t work as written because `pwd` is treated as a literal directory name instead of being evaluated. this makes the examples unusable out of the box.

changed `-v pwd:/src` → `-v "$(pwd)":/src` so that the current path is correctly mounted.

both affected commands are updated - now they run as expected without errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ something like the following command. It will map the current directory as
 `echidna`:
 
 ```sh
-$ docker run --rm -it -v `pwd`:/src ghcr.io/crytic/echidna/echidna
+$ docker run --rm -it -v "$(pwd)":/src ghcr.io/crytic/echidna/echidna
 ```
 
 Otherwise, if you want to locally build the latest version of Echidna, we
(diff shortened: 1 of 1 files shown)


elopez:
Hi @zkpepe ! What OS/shell were you seeing this behavior on? In my experience backticks should be equivalent to the $(...) syntax 🤔  https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03

zkpepe:
[elopez](https://github.com/elopez), I observed this behavior on macOS 14 / zsh. The issue occurs when the current path contains spaces, e.g., `/Users/John Doe/...`. Using `$(pwd)` with quotes ensures that the path is correctly expanded and handled by Docker, which backticks without quotes do not.

elopez:
Ah, so it's a matter of word splitting rather than backticks not working. Sounds good then!

PR #1436, opened 2025-08-28
Remove misleading opcode comment from gas.sol test

VolodymyrBg:
The comment described opcodes that didn't match the actual assembly code,
likely leftover from previous versions. Removing it eliminates confusion.

### DIFF
--- a/tests/solidity/cheat/gas.sol
+++ b/tests/solidity/cheat/gas.sol
@@ -39,8 +39,6 @@ contract TestCheatGas {
   }
 
   function echidna_gas_zero() public returns (bool){
-    // 0x14 as measured from opcodes, but let's leave some leeway in case solc changes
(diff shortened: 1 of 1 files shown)


elopez:
Hi! this comment is still useful to understand what we're trying to measure in the test and why it's that value. I went and checked the opcodes on more modern solcs, and they're still roughly the same as on the comment.

```
# 0.8.30
GAS PUSH0 PUSH0 PUSH1 0x24 PUSH1 0x9C PUSH0 DUP7 GAS CALLCODE GAS

# 0.8.25
GAS PUSH0 DUP1 PUSH1 0x24 PUSH1 0x9C PUSH0 DUP7 GAS CALLCODE GAS

# 0.8.10
GAS PUSH1 0x0 DUP1 PUSH1 0x24 PUSH1 0x9C PUSH1 0x0 DUP7 GAS CALLCODE GAS
```

PR #1437, opened 2025-08-29
Added initial support for generation of foundry reproducers

gustavo-grieco:
This PR add a basic support for the generation of Foundry reproducers, with the code inspired by [Runes](https://github.com/Enigma-Dark/runes).

### DIFF
--- a/lib/Echidna/Output/Foundry.hs
+++ b/lib/Echidna/Output/Foundry.hs
@@ -5,13 +5,13 @@
 module Echidna.Output.Foundry (foundryTest) where
 
 import Data.Aeson (Value(..), object, (.=))
-import Data.List (nub)
+import Data.List (elemIndex, nub)
 import Data.Text (Text, unpack)
 import Data.Text.Encoding (decodeUtf8)
 import qualified Data.Text.Lazy as TL
 import Data.Text.Lazy (fromStrict)
 import Data.Maybe (fromMaybe)
-import Data.Vector as V hiding ((++), map, zipWith)
+import Data.Vector as V hiding ((++), map, zipWith, elemIndex)
 import EVM.ABI (AbiValue(..))
 import EVM.Types (W256, Addr)
 import Numeric (showHex)
@@ -39,7 +39,7 @@ createTestData mContractName test =
   let
     senders = nub $ map (.src) test.reproducer
     actors = zipWith actorObject senders [1..]
-    repro = map foundryTx test.reproducer
+    repro = map (foundryTx senders) test.reproducer
     cName = fromMaybe "YourContract" mContractName
   in
   object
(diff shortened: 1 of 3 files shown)


elopez:
This could be a good use case for mustache templates as well if we're adding the dependency for #1415

gustavo-grieco:
@elopez changed the code to use mustache templates, similar to the report generation. Please review.

lib/Echidna/Output/Foundry.hs
```diff
@@ -0,0 +1,89 @@
...
+formatAddr :: Addr -> String
+formatAddr addr = "address(0x" ++ showHex (fromIntegral addr :: W256) "" ++ ")"
+
+-- | Generate a single transaction line for the reproducer.
+foundryTx :: Tx -> String
+foundryTx tx =
+  case tx.call of
+    SolCall (name, args) ->
+      let
+        sender = "_setUpActor(address(0x" ++ showHex (fromIntegral tx.src :: W256) "" ++ "));"
```
elopez: I think this should use the constants (USER1 etc) for better readability

lib/Echidna/Output/Foundry.hs
```diff
@@ -0,0 +1,89 @@
...
+  ]
+
+-- | Format an address for Solidity.
+formatAddr :: Addr -> String
+formatAddr addr = "address(0x" ++ showHex (fromIntegral addr :: W256) "" ++ ")"
+
+-- | Generate a single transaction line for the reproducer.
+foundryTx :: Tx -> String
+foundryTx tx =
+  case tx.call of
```
elopez, line 66: should this also produce warp/roll calls as appropriate?

gustavo-grieco: Done!

lib/Echidna/Output/Foundry.hs
```diff
@@ -0,0 +1,96 @@
...
+        senderName =
+          case elemIndex tx.src senders of
+            Just i -> "USER" ++ show (i + 1)
+            Nothing -> formatAddr tx.src
+        prelude =
+          (if time > 0 || blocks > 0 then "    _delay(" ++ show time ++ ", " ++ show blocks ++ ");\n" else "") ++
+          "    _setUpActor(" ++ senderName ++ ");"
+        call = "    Target." ++ unpack name ++ "(" ++ foundryArgs (map abiValueToString args) ++ ");"
+      in object ["prelude" .= prelude, "call" .= call]
+    _ -> object ["prelude" .= ("" :: String), "call" .= ("" :: String)]
```
elopez: instead of these empty objects, you should be able to `mapMaybe (foundryTx senders) test.reproducer` and return Nothing here

lib/Echidna/Output/Foundry.hs
gustavo-grieco: Done!

lib/Echidna/Output/Foundry.hs
gustavo-grieco: Done!

lib/Echidna/Output/Foundry.hs
```diff
@@ -0,0 +1,96 @@
...
+-- | Generate a Foundry test from an EchidnaTest result.
+foundryTest :: Maybe Text -> EchidnaTest -> TL.Text
+foundryTest mContractName test =
+  case test.testType of
+    AssertionTest{} ->
+      let testData = createTestData mContractName test
+      in fromStrict $ substituteValue template (toMustache testData)
+    _ -> ""
+
+-- | Create an Aeson Value from test data for the Moustache template.
```
elopez:
```suggestion
-- | Create an Aeson Value from test data for the Mustache template.
```

src/Main.hs
```diff
@@ -101,6 +103,25 @@ main = withUtf8 $ withCP65001 $ do
...
+            (AssertionTest{}, state) | isLargeOrSolved state ->
+              do
+              let
+                reproducerHash = (show . abs . hash) test.reproducer
+                fileName = foundryDir </> "Test." ++ reproducerHash <.> "sol"
+                content = foundryTest cliSelectedContract test
+              liftIO $ writeFile fileName (TL.unpack content)
+            _ -> pure ()
+
+      -- TODO: We use the corpus dir to save coverage reports which is confusing.
```
elopez: what does this floating comment mean? Is it from a mismerge?

gustavo-grieco: Yes, my fault, let me remove it.

elopez (approved):

src/Main.hs
gustavo-grieco: Done!

PR #1439, opened 2025-08-29
[Do Not Merge] Avoid crashing when cheatcode is invalid

gustavo-grieco:
A draft PR with the fix from hevm applied from my branch for testing. It should not crash with invalid cheatcodes.


gustavo-grieco:
Replaced by #1467

PR #1441, opened 2025-09-08
Symbolic execution fixes to make sure no counter example is missed

gustavo-grieco:


lib/Echidna/Onchain.hs
```diff
@@ -76,8 +77,8 @@ data FetchedContractData = FetchedContractData
   }
   deriving (Generic, ToJSON, FromJSON, Show)
 
-instance ToJSONKey W256 where
-  toJSONKey = toJSONKeyText (Text.pack . show)
+--instance ToJSONKey W256 where
+--  toJSONKey = toJSONKeyText (Text.pack . show)
```
elopez: Please drop this if it is no longer needed

lib/Echidna/SymExec/Common.hs
```diff
@@ -224,26 +147,30 @@ getUnknownLogs = mapMaybe (\case
...
-                    & #state % #caller .~ SymAddr "caller"
-                    & #state % #calldata .~ cd
-                    & #env % #contracts .~ (Map.union vmSym'.env.contracts vm.env.contracts)
+  let
+    vm' = vmReset & execState (loadContract (LitAddr dst))
+                  & #tx % #isCreate .~ False
+                  & #state % #callvalue .~ TxValue
+                  & #state % #caller .~ SymAddr "caller"
+                  & #state % #calldata .~ cd
+                  -- & #env % #contracts .~ (Map.union vmSym'.env.contracts vm.env.contracts)
```
elopez: Why are contracts not loaded on the symbolic VM?

package.yaml
```diff
@@ -76,6 +76,7 @@ library:
     - vty-crossplatform
     - wai-extra
     - warp
+    - wreq == 0.5.4.3
```
elopez: do we need the strict pin here?

lib/Echidna/SymExec/Common.hs
gustavo-grieco: The concrete VM already have all the contracts and we symbolize the state.

package.yaml
gustavo-grieco: I'm not sure. Using no pin will fail, so any suggestion is appreciated.

lib/Echidna/Onchain.hs
gustavo-grieco: Actually, hevm used some instance (and therefore, I had to comment it). I'm not sure if this affects our code or not 🤔

gustavo-grieco: Removed this comment

lib/Echidna/SymExec/Common.hs
gustavo-grieco: Removed this comment. All the tests seem to be working fine, even with multiple contracts. If we found some contract is not loaded, we can restore this line.

flake.nix
```diff
@@ -54,8 +54,8 @@
           (pkgs.haskellPackages.callCabal2nix "hevm" (pkgs.fetchFromGitHub {
             owner = "ethereum";
             repo = "hevm";
-            rev = "9982c580ed19b88ebab9744d29d940fd2f0bd8c6";
-            sha256 = "sha256-B60aJo9EfIaTWuJaN06Wl1Br+rstzUAYGFBE5rX72Fs=";
+            rev = "87fe0eec5abb69c0b54c097784dfd8712a36de7";
```
elopez:
```suggestion
            rev = "87fe0eec5abb69c0b54c097784dfd8712a36de70";
```

let's use full commit hashes 😄

PR #1444, opened 2025-09-11
Support for encoding valid calls for callbacks

gustavo-grieco:
This PR add an experimental feature that allows to automatically encode valid callback data in any `bytes` argument using non-view/non-pure functions to call. For instance, now echidna can automatically solve this:

```solidity
pragma solidity ^0.8.0;

contract MultiDelegatecall {
   
    function multicall(bytes[] memory data)
        external
        payable
        returns (bytes[] memory results)
    {
        results = new bytes[](data.length);

        for (uint256 i; i < data.length; i++) {
            (bool ok, bytes memory res) = address(this).delegatecall(data[i]);
            results[i] = res;
            if (!ok) {
               return results;
            }
        }
    }
}

contract TestMultiDelegatecall is MultiDelegatecall {
    event Log(address caller, string func, uint256 i);

    function func1(uint256 x, uint256 y) external {
        emit Log(msg.sender, "func1", x + y);
    }

    function func2() external returns (uint256) {
        emit Log(msg.sender, "func2", 2);
        return 111;
    }

    mapping(address => uint256) public balanceOf;
    function mint() external payable {
        balanceOf[msg.sender] += msg.value;
    }
}

contract Test {
  TestMultiDelegatecall c = new TestMultiDelegatecall();
  bool once = false;

  function testMultiDelegatecall(bytes[] memory data)
        external
        payable
    { 
      require(!once);
      once = true;
      c.multicall{value: msg.value}(data);
      assert(c.balanceOf(address(this)) <= msg.value);
    }
}
```

With this PR, echidna will try to use possible callbacks encoded in the `bytes[]` argument from `testMultiDelegatecall`. It will immediately found a way to break the invariant:
```
testMultiDelegatecall(bytes[]): failed!💥  
  Call sequence:
    Test.testMultiDelegatecall(["\DC2I\197\139", "\DC2I\197\139"]) Value: 0x1

Traces: 
call TestMultiDelegatecall::multicall(bytes[])([0x1249c58b, 0x1249c58b]) (/Users/g/Code/echidna/multicall.sol:50)
 ├╴call TestMultiDelegatecall::mint()() (/Users/g/Code/echidna/multicall.sol:13)
 │  └╴← 0x
 ├╴call TestMultiDelegatecall::mint()() (/Users/g/Code/echidna/multicall.sol:13)
 │  └╴← 0x
 └╴← ([0x, 0x])
call TestMultiDelegatecall::balanceOf(address)(Test) (/Users/g/Code/echidna/multicall.sol:51)
 └╴← (2)
```

### DIFF
--- a/lib/Echidna/ABI.hs
+++ b/lib/Echidna/ABI.hs
@@ -434,7 +434,7 @@ genAbiCallM genDict (name, types) = do
 -- | Given a list of 'SolSignature's, generate a random 'SolCall' for one,
 -- possibly with a dictionary.
 genInteractionsM
-  :: (MonadRandom m)
+  :: MonadRandom m
   => GenDict
   -> NonEmpty SolSignature
   -> m SolCall
--- a/lib/Echidna/Campaign.hs
+++ b/lib/Echidna/Campaign.hs
@@ -426,7 +426,7 @@ randseq
 randseq deployedContracts = do
   env <- ask
   let world = env.world
-  
+
   let
     mutConsts = env.cfg.campaignConf.mutConsts
     seqLen = env.cfg.campaignConf.seqLen


lib/Echidna.hs
```diff
@@ -73,22 +75,28 @@ prepareContract cfg solFiles buildOutput selectedContract seed = do
...
-                     Set.empty
-                     seed
-                     (returnTypes contracts)
-
+                      -- make sure we don't use cheat codes to form fuzzing call sequences
+                      (Set.delete (AbiAddress $ forceAddr cheatCode) constants)
+                      Set.empty
+                      seed
+                      (returnTypes contracts)
+                      nonViewPureSigs
```
elopez: why the extra spaces here? 😄

lib/Echidna.hs
```diff
@@ -32,6 +33,7 @@ import Echidna.Types.Tx
 import Echidna.Types.World
 import Echidna.Types.Test (EchidnaTest)
 import Echidna.Types.Signature (ContractName)
+import Echidna.SourceMapping (findSrcForReal)
```
elopez: please keep imports ordered

lib/Echidna/ABI.hs
```diff
@@ -121,6 +121,7 @@ data GenDict = GenDict
     -- ^ Return types of any methods we scrape return values from
   , dictValues :: !(Set W256)
     -- ^ A set of int/uint constants for better performance
+  , allSigs    :: ![SolSignature]
   }
```
elopez, line 126: please add a doc string for what this field is

lib/Echidna/ABI.hs
```diff
@@ -197,7 +199,7 @@ getRandomInt n =
 -- | Synthesize a random 'AbiValue' given its 'AbiType'. Doesn't use a dictionary.
 -- Note that we define the dictionary case ('genAbiValueM') first (below), so
 -- recursive types can be generated using the same dictionary easily
-genAbiValue :: MonadRandom m => AbiType -> m AbiValue
+genAbiValue :: (MonadRandom m) => AbiType -> m AbiValue
```
elopez: are these extra parens something needed for hlint? or an unrelated change?

lib/Echidna.hs
gustavo-grieco: Removed!

lib/Echidna.hs
gustavo-grieco: Done

lib/Echidna/ABI.hs
gustavo-grieco: Done!

lib/Echidna/ABI.hs
gustavo-grieco: Fixed!

elopez (approved):

Issue #1452, opened 2025-09-17
[Bug-Candidate]: Internal Error: Call target doesn't exist


gustavo-grieco:
### Describe the issue:

There is a very strange issue crashing echidna (`master`). To reproduce:
1. `forge init crash`
2. `cd crash`
3. Open `foundry.toml` add `via_ir = true`.
4. Add the following contract (e.g. `src/Crash.sol`):
```solidity
pragma solidity ^0.8.13;
import {Test} from "forge-std/Test.sol";

contract Crash is Test {
    function crash(
        address _creator,
        address _quoteToken,
        address _underlyingToken,
        uint256 _r0,
        uint256 _r_star,
        uint256 _k0,
        uint256 _cashIn,
        uint256 _initLPShares,
        address _vault
    ) public asActor {
        return;
        t(false, "");
    }

    modifier asActor() {
        vm.prank(address(this));
        _;
    }

    event Log(string);
    function t(bool b, string memory reason) internal {
        if (!b) {
            emit Log(reason);
            assert(false);
        }
    }
}
```
5. Run `echidna . --contract Crash --test-limit 100000000000 --format text --test-mode assertion`.

Result:
```
[2025-09-17 09:13:01.81] [Worker 0] Crashed:

Internal Error: Call target doesn't exist -- CallStack (from HasCallStack):
  internalError, called at src/EVM.hs:1576:9 in hevm-0.55.1-FvmSjNlhmqw6zGwQpkjqs9:EVM
CallStack (from HasCallStack):
  error, called at src/EVM/Types.hs:1552:19 in hevm-0.55.1-FvmSjNlhmqw6zGwQpkjqs9:EVM.Types
  internalError, called at src/EVM.hs:1576:9 in hevm-0.55.1-FvmSjNlhmqw6zGwQpkjqs9:EVM

Please report it to https://github.com/crytic/echidna/issues
[2025-09-17 09:13:01.81] [Worker 2] Crashed:

Internal Error: Call target doesn't exist -- CallStack (from HasCallStack):
  internalError, called at src/EVM.hs:1576:9 in hevm-0.55.1-FvmSjNlhmqw6zGwQpkjqs9:EVM
CallStack (from HasCallStack):
  error, called at src/EVM/Types.hs:1552:19 in hevm-0.55.1-FvmSjNlhmqw6zGwQpkjqs9:EVM.Types
[...]

PR #1453, opened 2025-09-17
Set up proper reversion state when loading transactions

gustavo-grieco:
Hot fix for #1452

### DIFF
--- a/lib/Echidna/Transaction.hs
+++ b/lib/Echidna/Transaction.hs
@@ -192,6 +192,8 @@ setupTx tx@Tx{call} = fromEVM $ do
   modify' $ \vm ->
     let intrinsicGas = txGasCost vm.block.schedule isCreate calldata
         burned = min intrinsicGas vm.state.gas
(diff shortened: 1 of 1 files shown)


lib/Echidna/Transaction.hs
```diff
@@ -191,8 +192,10 @@ setupTx tx@Tx{call} = fromEVM $ do
   modify' $ \vm ->
     let intrinsicGas = txGasCost vm.block.schedule isCreate calldata
         burned = min intrinsicGas vm.state.gas
+        reversionState = EVM.Transaction.setupTx vm.tx.origin vm.block.coinbase vm.tx.gasprice vm.tx.gaslimit vm.env.contracts
```
elopez, line 197:
```suggestion
        -- This might not be 100% accurate (it does not revert the balance transfers)
        -- but Echidna resets the vm on revert so it should not make a difference
        reversionState = EVM.Transaction.setupTx vm.tx.origin vm.block.coinbase vm.tx.gasprice vm.tx.gaslimit vm.env.contracts
```

Issue #1459, opened 2025-09-28
[Bug-Candidate]: missing info in transactions from JSON output


monperrus:
### Describe the issue:

I run echidna in JSON output

### Expected behavior

The JSON object contains arguments and msg.value

```
Contract.rescueETH(4370000) from: 0x0000000000000000000000000000000000020000 Time delay: 400981 seconds Block delay: 32147
```
...
```json
"transactions":[{"arguments":[4370000],"value":0,"contract":"","function":"rescueETH","gas":12500000,"gasprice":0}],"type":"property"}]
```

### Actual behavior:

- The JSON object contains an always empty argument list.
- The JSON object does not contain any msg.value

```json
"transactions":[{"arguments":[],"contract":"","function":"rescueETH","gas":12500000,"gasprice":0}],"type":"property"}]
```

### Code example to reproduce the issue:

n-a

### Version:

Echidna 2.2.7

### Relevant log output:

```shell

```

elopez:
Hi, thanks for the report! I believe this was recently fixed by PR https://github.com/crytic/echidna/pull/1445 , can you retry with a build from the current master branch and let us know if you're still experiencing issues? You can get prebuilt binaries from CI if you need them: https://github.com/crytic/echidna/actions/runs/18060860641

monperrus:
thanks a lot @elopez 

I confirm that https://github.com/crytic/echidna/pull/1445 fixes the arguments problem.

msg.value is still missing.

gustavo-grieco:
@monperrus please test #1460 (prebuilt binaries are available for every PR in case you need them)

monperrus:
Thanks a lot @elopez  and @gustavo-grieco 

This is perfectly solved!

Now the next blocker regarding JSON is https://github.com/crytic/echidna/issues/1462

PR #1460, opened 2025-09-29
Include value into json output and use string instead of integer

gustavo-grieco:
A fix for #1459, which include converting gas and gasprice fields to strings. This additional change is necessary to make sure the values are preserved correctly since the json specification does not allow values larger than 2 ** 80 (aprox.)


elopez (approved):

gustavo-grieco:
Let merge this but note that we need to update the documentation as well.

PR #1461, opened 2025-09-29
Downgrade z3 to 4.12.6

gustavo-grieco:
Some symbolic execution related tests are failing when using z3 (4.15) but work correctly with 4.12.6 for some unknown reason. While we investigate, we will downgrade z3 to have the tests passing (they are also passing with other solvers such as bitwuzla).


gustavo-grieco:
Done!

PR #1464, opened 2025-10-01
chore: fix subprocess silence logic in compiler selection

mdqst:
noticed a small issue in how we were setting `silent` for the compiler subprocess.
previously it used a combination of `and/or` which can lead to unexpected results.

changed it to a clear ternary expression:

```python
silent = subprocess.DEVNULL if os.getenv("SOLC_SELECT_SILENT", "1") == "1" else None
```

now the behavior is straightforward and consistent.

### DIFF
--- a/docker/solc-install.py
+++ b/docker/solc-install.py
@@ -15,7 +15,7 @@ if len(sys.argv) < 2:
 
 solc_version = os.getenv('SOLC_VERSION', None)
 if solc_version:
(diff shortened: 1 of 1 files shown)


elopez:
hi! out of curiosity, what prompted this change? do you use this script somewhere? and what issues did you see? (I'm not sure how the and/or pseudo-ternary could fail here, but I agree the if/else is clearer)

I'm mainly asking as we only copy this script to the distroless docker target, and that is not a docker target we ever built on CI or people use. I've thought about removing it altogether.

Issue #1465, opened 2025-10-02
bugs in foundry test generation


monperrus:
Here, I collect bugs in foundry test generation

### contract name

```
Error (2333): Identifier already declared.
 --> test/Test.1904265591758374514.sol:6:1:
  |
6 | contract Test is Test {
```

should be for example `Test1 is Test`

### fallback 

call to fallback function is written as an empty string

```
Error (2314): Expected identifier but got '('
  --> test/Test.8771174626403390381.sol:19:16:
   |
19 |         Target.();
```

### null byte
 
files contains \\0 (written as <nul>  below)
```
Error (8936): Expected even number of hex-nibbles.
  --> test/Test.2952294588457979761.sol:23:29:
   |
23 |         Target.renounceRole(hex"<nul><nul><nul>", address(0xa329c0648769a73afac7f9381e08fb43dbea72));
   |  
```

gustavo-grieco:
Great report! You can try using the sub issue GitHub feature to keep them in a list

gustavo-grieco:
You can also ask ChatGPT/Claude/Gemini to create PRs for each one, as they are capable to do it.

radubahmata:
Hi @monperrus. I will give this a shot, if you have anything that would help reproduce it would be highly appreciated 🙏

monperrus:
Hi @BowTiedRadone!

Awesome.

```
contract bugfallback {
  event AssertionFailed();
  function f() external {} // workaround ABI is empty, are you sure your constructor is right?
  fallback() external {
    emit AssertionFailed();
  }
}
```

`$ ./echidna --test-mode assertion --corpus-dir foobar --format text bugfallback.sol`

For some reason, no forge test is written with this.

gustavo-grieco:
"assertion failed" tests are not supported. For the first step, we will omit that case, since I think it is a bit harder to fix

monperrus:
@gustavo-grieco I'm puzzled. you told me in https://github.com/crytic/echidna/issues/1462#issuecomment-3361025611 that  " Foundry generation only works in assertion mode" and I indeed go some.

gustavo-grieco:
My mistake, the assertion mode using events is not supported yet. 😅

radubahmata:
[...]

PR #1484, opened 2025-11-27
Fix Foundry test generation contract name collision

BowTiedRadone:
This PR changes generated Foundry tests contract name from "Test" to "FoundryTest" to avoid identifier collision with the `forge-std/Test.sol` base class. The previous output `contract Test is Test` caused a Solidity compilation error.

Part of #1465 report.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -249,6 +249,9 @@ jobs:
           HOST_OS: ${{ runner.os }}
           SOLC_VER: ${{ matrix.solc }}
 
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+
       - name: Download testsuite
         uses: actions/download-artifact@v6
         with:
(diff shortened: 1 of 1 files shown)


gustavo-grieco:
This looks good, but I would like to have the test for this feature to actually call foundry with a minimal example and verify it compiles correctly. While it is a trivial to know that it works here, this test functionality will be useful for future PRs (there are a couple of foundry issues that also will need a unit test).

radubahmata:
@gustavo-grieco Updated the test so it runs `forge init`. This required adding a Foundry installation step in CI.

radubahmata:
@gustavo-grieco Anything left to be covered in this PR?

gustavo-grieco:
I think this one looks good, but it will be merged after the release of 2.3 (which can be as close as next week). I'm very interested to develop this test code into a full test of the foundry unit-test capabilities (e.g. breaking an invariant, generating the foundry test and then run `forge test` to check that the test can be reproduced. Are you interested in keep expanding this in a future PR?

radubahmata:
@gustavo-grieco Thanks for the update and looking forward to 2.3! 🚀 

> Are you interested in keep expanding this in a future PR?

Sure! Also, two more Foundry test generation fixes that I'm planning to resolve are left (#1465).

PR #1487, opened 2025-11-29
Update `hevm` to `9ba5e52fc7ec7ae6f7f3a25d5ee426625d2aa9d`

elopez:
This updates hevm and sorts the imports across the codebase.

### DIFF
--- a/lib/Echidna/Onchain.hs
+++ b/lib/Echidna/Onchain.hs
@@ -180,5 +180,5 @@ fetchChainIdFrom (Just url) = do
     EVM.Fetch.Latest -- this shouldn't matter
     (EVM.Fetch.fetchWithSession url sess)
     EVM.Fetch.QueryChainId
-  pure $ either (const Nothing) Just $ res
+  pure $ either (const Nothing) Just res
 fetchChainIdFrom Nothing = pure Nothing


lib/Echidna/Exec.hs
```diff
@@ -144,13 +141,14 @@ execTxWith executeTx tx = do
...
+                when (status == EVM.Fetch.Fresh) $ logMsg $ "Fetched new slot: " <> show q
                 fromEVM (continuation value)
-              Nothing -> do
-                -- TODO: How should we fail here? It could be a network error,
-                -- RPC server returning junk etc.
-                logMsg $ "ERROR: Failed to fetch slot: " <> show q
+              EVM.Fetch.FetchFailure _ -> do
                 fromEVM (continuation 0)
+              EVM.Fetch.FetchError e -> do
+                error $ "ERROR: Failed to fetch slot: " <> show q <> " " <> T.unpack e
```
elopez, line 149: is a full error what we want here? or should this be a logMsg + also have a `fromEVM (continuation 0)` like we do earlier in the file?

lib/Echidna/Onchain.hs
```diff
@@ -173,8 +176,11 @@ saveCoverageReport env runId = do
...
 fetchChainIdFrom (Just url) = do
   sess <- Session.newAPISession
-  EVM.Fetch.fetchQuery
+  res <- EVM.Fetch.fetchQuery
     EVM.Fetch.Latest -- this shouldn't matter
     (EVM.Fetch.fetchWithSession url sess)
     EVM.Fetch.QueryChainId
+  pure $ case res of
+    Right val -> Just val
+    Left _ -> Nothing
```
elopez: looks like `rightToMaybe`

lib/Echidna/Exec.hs
gustavo-grieco: The issue is that if we add that zero slot and the slot was something else, we are introducing an invalid value into the cache,  and we are risking to false positives or false negatives. I think it is safer to crash, but ideally we need to wait a random amount of second and re-try as this is caused 99% by sending too many request to a RPC.

elopez: Is that not also the case in PleaseFetchContract's FetchError? it does not crash in that case.

gustavo-grieco: Good catch!

lib/Echidna/Onchain.hs
gustavo-grieco: Not available 😞

lib/Echidna/Exec.hs
gustavo-grieco: Done

Issue #1486, opened 2025-11-29
Add support for chainId > 1 in source code fetcher / RPC mode


elopez:
Currently chainid / etherscan is hardcoded when fetching contract code and sourcemaps. We should support other chains as well

gustavo-grieco:
Fixed

Issue #1230, opened 2024-04-05
Feature Request: Option to disable Etherscan downloads with onchain fuzzing


rappie:
### Describe the desired feature

During the development of a fuzzing campaign I am rarely interested in coverage. It takes an extra CTRL+C to close Echidna, and it is a waste of unused resources :slightly_smiling_face:

ggrieco-tob:
This seems fair, I think Etherscan should be used only if RPC is enabled.

rappie:
Actually, I would like the option to fully disable it. In practice I would have it disabled by default in all cases and only enable it when I'm actively investing coverage or doing production runs.

ggrieco-tob:
If you do not specify a corpus directory, it will not write any coverage.

gustavo-grieco:
Actually, what we should do here is to make the etherscan API key required and never fetch if the key is not available.

gustavo-grieco:
Fixed in 2.3. If the Etherscan API key is not set, then it will not fetch the coverage maps.

PR #1492, opened 2025-12-13
Do not fetch if etherscan key is not available and support different chain ids

gustavo-grieco:
Fix for #1230  #1486

### DIFF
--- a/lib/Echidna/Onchain.hs
+++ b/lib/Echidna/Onchain.hs
@@ -1,12 +1,18 @@
-{-# LANGUAGE DeriveAnyClass #-}
-
-module Echidna.Onchain where
+module Echidna.Onchain
+  ( etherscanApiKey
+  , fetchChainIdFrom
+  , rpcBlockEnv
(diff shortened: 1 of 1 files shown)


lib/Echidna/Onchain.hs
```diff
@@ -115,63 +115,64 @@ readFileIfExists path = do
...
 
 
 saveCoverageReport :: Env -> Int -> IO ()
 saveCoverageReport env runId = do
   case env.cfg.campaignConf.corpusDir of
     Nothing -> pure ()
     Just dir -> do
-      -- coverage reports for external contracts, we only support
-      -- Ethereum Mainnet for now
-      when (env.chainId == Just 1) $ do
```
elopez, line 160: I'm not sure we can allow other chains yet, as fetchContractSourceMap only supports eth mainnet

elopez: I implemented support for it on 5fd1eed

PR #1498, opened 2025-12-16
Feat: SSE POST Events for lcov dump and 

GalloDaSballo:
## Feature:

Intoduces a way to trigger an lcov coverage report to be generated, as well as having echidna reload all corpus files.

## Usage:

NOTE: Both require the corpus folder config and SSE to be active:
`ECHIDNA_COMMAND --server PORT --corpus echidna`

`POST: /dump_lcov` will return an lcov file that was added to the corpus folder

`POST: `/reload_corpus` trigger a reload of all corpus files to update the in-memory corpus

## Testing

I've built from our own fork and ran here:
https://staging.getrecon.xyz/shares/74b7786e-71f0-4878-903b-ce9213aba25c

The lcov allows us to build the coverage chart you see

### DIFF
--- a/.github/workflows/release-fork.yml
+++ b/.github/workflows/release-fork.yml
@@ -1,0 +1,95 @@
+name: "Build & Release"
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version name (e.g., 2.3.0-custom)'
+        required: true
+        default: 'dev'
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
+
+jobs:
+  build:
+    name: Build Linux (x86_64)
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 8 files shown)


.github/workflows/release-fork.yml
gustavo-grieco, line 1: Is this file necessary?

GalloDaSballo: > Is this file necessary?

No, I added it to generate builds on main

gustavo-grieco: I think once it is here as a PR, you don't need this file and can be removed. Perhaps @elopez can confirm this.

gustavo-grieco:
This will be replaced by #1502, once all the commands are there.

gustavo-grieco:
Both commands were already implemented in #1502, so I'm closing this one.

PR #1499, opened 2025-12-16
Improved support for foundry tests

gustavo-grieco:
Foundry tests should be able to be used, as much as possible. In particular, we could re-use all the test infrastructure, for stateless tests.  We want to eventually reach some decent level of compatibility.

### DIFF
--- a/lib/Echidna/Solidity.hs
+++ b/lib/Echidna/Solidity.hs
@@ -1,6 +1,7 @@
 module Echidna.Solidity where
 
 import Control.Monad (when, unless, forM_)
+import Control.Monad.IO.Class (liftIO)
 import Control.Monad.Catch (MonadThrow(..))
 import Control.Monad.Extra (whenM)
 import Control.Monad.Reader (ReaderT(runReaderT))
@@ -231,6 +232,11 @@ loadSpecified env mainContract cs = do
                                            solConf.contractAddr
                                            unlimitedGasPerBlock
                                            (0, 0)
+    -- Call setUp() for any contract that has IS_TEST() in its ABI, regardless of
+    -- test mode. Contracts following the Foundry/dapptools convention (IS_TEST)
+    -- expect setUp() to run even in assertion or property mode.
+    when (isFoundryMode solConf.testMode && is_testFunction `notElem` abi) $
+      liftIO $ putStrLn "Warning: running in Foundry mode but contract does not have IS_TEST(). setUp() will not be called."
     vm4 <- if is_testFunction `elem` abi && setUpFunction `elem` abi
               then snd <$> transaction
               else pure vm3
--- a/tests/solidity/foundry-basic/basic.sol
+++ b/tests/solidity/foundry-basic/basic.sol
@@ -20,6 +20,7 @@ contract Greeter {
 }
 
 contract GreeterTestSetup {
+    bool public IS_TEST = true;
     Greeter greeter;
 
     function greeting() public view returns (string memory) {


gustavo-grieco:
@BowTiedRadone do you want to own this PR? We need to add some test integration with Foundry so we can verify we are catching correctly the failures (at least, for stateless tests)

radubahmata:
@gustavo-grieco Sure! Thanks.

lib/Echidna/SymExec/Common.hs
```diff
@@ -27,6 +31,24 @@ import Echidna.Types.Config (EConfig(..))
...
+    Failure {} -> PBool False
+    {-Failure _ _ (Revert msg) -> case msg of
+      ConcreteBuf b ->
+        -- NOTE: assertTrue/assertFalse does not have the double colon after "assertion failed"
+        let assertFail = (selector "Error(string)" `BS.isPrefixOf` b) &&
+              ("assertion failed" `BS.isPrefixOf` BS.drop txtOffset b)
+        in if assertFail || b == panicMsg 0x01 then PBool False
+        else PBool True
+      _ -> error "Non-concrete revert message in assertion check"
+    -}
```
elopez: is this needed? if not, can you remove the commented out code?

lib/Echidna/SymExec/Common.hs
```diff
@@ -137,7 +159,7 @@ exploreMethod method contract sources vm defaultSender conf veriOpts solvers rpc
   let
     vm' = vmReset & execState (loadContract (LitAddr dst))
                   & #tx % #isCreate .~ False
-                  & #state % #callvalue .~ TxValue
+                  & #state % #callvalue .~ Lit 0 --TxValue
```
elopez: why is callvalue forced to zero here?

lib/Echidna/SymExec/Verification.hs
```diff
@@ -56,6 +56,7 @@ verifyMethod method contract vm = do
     askSmtIters = conf.campaignConf.symExecAskSMTIters
     rpcInfo = RpcInfo (rpcFetcher conf.rpcUrl (fromIntegral <$> conf.rpcBlock))
     defaultSender = fromJust $ Set.lookupMin conf.solConf.sender <|> Just 0
+    --testMode = conf.campaignConf.testMode
```
elopez:
```suggestion
```

lib/Echidna/Campaign.hs
```diff
@@ -288,6 +288,8 @@ runSymWorker callback vm dict workerId _ name = do
     lift callback
 
     (symTxs, partials) <- liftIO $ takeMVar symTxsChan
+    liftIO $ print symTxs
+    liftIO $ print partials
```
elopez: is this used? It looks like it might be debug prints

lib/Echidna/Solidity.hs
```diff
@@ -146,11 +146,11 @@ filterMethods contractName (Whitelist ic) ms =
 filterMethods contractName (Blacklist ig) ms =
   NE.filter (\s -> encodeSigWithName contractName s `notElem` ig) ms
 
--- | Filter methods with arguments, used for dapptest mode
+-- | Filter methods with arguments, used for foundry mode
 filterMethodsWithArgs :: NonEmpty SolSignature -> NonEmpty SolSignature
 filterMethodsWithArgs ms =
-  case NE.filter (\(n, xs) -> T.isPrefixOf "invariant_" n || not (null xs)) ms of
-    [] -> error "No dapptest tests found"
+  case NE.filter (\(n, xs) -> T.isPrefixOf "test" n || (T.isPrefixOf "invariant_" n || not (null xs))) ms of
```
elopez, line 161: is this a foundry convention?

lib/Echidna/Test.hs
```diff
@@ -52,13 +52,13 @@ createTest m = EchidnaTest Open m v [] Stop Nothing Nothing
...
 validateTestModeError =
-  "Invalid test mode (should be property, assertion, dapptest, optimization, overflow or exploration)"
+  "Invalid test mode (should be property, assertion, foundry, optimization, overflow or exploration)"
 
 validateTestMode :: String -> TestMode
 validateTestMode s = case s of
   "property"     -> s
   "assertion"    -> s
-  "dapptest"     -> s
+  "foundry"     -> s
```
elopez:
```suggestion
  "foundry"      -> s
```

lib/Echidna/Test.hs
```diff
@@ -76,18 +76,19 @@ isPropertyMode :: TestMode -> Bool
 isPropertyMode "property" = True
 isPropertyMode _          = False
 
-isDapptestMode :: TestMode -> Bool
-isDapptestMode "dapptest"  = True
-isDapptestMode _           = False
+isFoundryMode :: TestMode -> Bool
+isFoundryMode "foundry"  = True
+isFoundryMode _           = False
```
elopez:
```suggestion
isFoundryMode _          = False
```

lib/Echidna/Test.hs
```diff
@@ -99,9 +100,13 @@ createTests m td ts r ss = case m of
...
-  "dapptest" ->
-    map (\s -> createTest (AssertionTest True s r))
-        (filter (\(n, xs) -> T.isPrefixOf "invariant_" n || not (null xs)) ss)
+  "foundry" ->
+    if len == 1 then
+      map (\s -> createTest (AssertionTest True s r))
+        (filter (\(n, xs) -> T.isPrefixOf "test" n && not (null xs)) ss)
+    else
+      map (\s -> createTest (AssertionTest True s r))
+          (filter (\(n, xs) -> T.isPrefixOf "invariant_" n || not (null xs)) ss)
```
elopez, line 111: what is len here?

lib/Echidna/Campaign.hs
```diff
@@ -288,6 +288,8 @@ runSymWorker callback vm dict workerId _ name = do
     lift callback
 
     (symTxs, partials) <- liftIO $ takeMVar symTxsChan
+    liftIO $ print symTxs
```
gustavo-grieco: Remove this and the next line

lib/Echidna/Test.hs
```diff
@@ -215,7 +220,9 @@ checkStatefulAssertion vm sig addr = do
     -- Whether the last transaction executed opcode 0xfe, meaning an assertion failure.
     isAssertionFailure = case vm.result of
       Just (VMFailure (UnrecognizedOpcode 0xfe)) -> True
+      Just (VMFailure (Revert (ConcreteBuf msg))) -> "assertion failed" `BS.isPrefixOf` BS.drop txtOffset msg
```
elopez, line 225: where is this magic string defined?

lib/Echidna/Test.hs
gustavo-grieco: that's seqLen, foundry defines different fuzz behavior according to the invariant mode (here represented by seqLen == 1 or seqLen > 1)

lib/Echidna/SymExec/Common.hs
gustavo-grieco: A foundry test cannot accept ether, so sending ether to it will trigger a revert that will be classified as a failure. This needs to be fixed checking the actual test mode (e.g. isFoundry)

lib/Echidna/Test.hs
radubahmata: This is the `hevm` convention for Foundry `assertX` failures. Example: https://github.com/argotorg/hevm/blob/ed90053fa0ed69e658a75ab0ed64d467f5a5448d/src/EVM.hs#L2114

lib/Echidna/Campaign.hs
gustavo-grieco: Fixed

lib/Echidna/SymExec/Common.hs
gustavo-grieco: Done

lib/Echidna/SymExec/Verification.hs
gustavo-grieco: Done

lib/Echidna/Solidity.hs
gustavo-grieco: Done

lib/Echidna/Test.hs
gustavo-grieco: Done

lib/Echidna/Test.hs
gustavo-grieco: Done

lib/Echidna/Test.hs
gustavo-grieco: Added a comment

lib/Echidna/Test.hs
```diff
@@ -215,24 +222,28 @@ checkStatefulAssertion vm sig addr = do
...
+    txtOffset = 4+32+32 -- selector + offset + length
     -- Test always passes if it doesn't target the last executed contract and function.
     -- Otherwise it passes if it doesn't cause an assertion failure.
     events = extractEvents False dappInfo vm
     eventFailure = not (null events) && (checkAssertionEvent events || checkPanicEvent "1" events)
     isFailure = isCorrectTarget && (eventFailure || isAssertionFailure)
   pure (BoolValue (not isFailure), vm)
 
+-- | Magic return code used by hevm to signal a foundry vm.assume failure.
+-- See: https://github.com/ethereum/hevm/blob/main/src/EVM.hs (search for FOUNDRY::ASSUME)
```
elopez: I cannot see this being emitted anywhere by hevm. Looks like upstream uses an AssumeCheatFailed error to signal this condition, instead of this magic string. Can you check this again? https://github.com/argotorg/hevm/pull/986

gustavo-grieco: @BowTiedRadone can you fix this one?

gustavo-grieco: This should be solved already

lib/Echidna/Output/Foundry.hs
```diff
@@ -76,7 +76,10 @@ foundryTx senders tx =
         prelude =
           (if time > 0 || blocks > 0 then "    _delay(" ++ show time ++ ", " ++ show blocks ++ ");\n" else "") ++
           "    _setUpActor(" ++ senderName ++ ");"
-        call = "    Target." ++ unpack name ++ "(" ++ foundryArgs (map abiValueToString args) ++ ");"
+        -- Handle fallback function (empty name).
+        call = if unpack name == ""
+          then "    address(Target).call(\"\");"
```
elopez, line 81: Should the fallback also support argument(s)?

README.md
```diff
@@ -54,6 +54,17 @@ $ echidna tests/solidity/basic/flags.sol
...
 
+### Testing modes
+
+The example above uses the default **property** mode, but Echidna supports several testing modes, configured via `testMode` in the config file or `--test-mode` on the CLI:
+
+* **`property`** (default): Test `echidna_`-prefixed functions that return `bool`.
+* **`assertion`**: Detect assertion failures from `assert()` and Foundry's `assertX` helpers (`assertTrue`, `assertEq`, etc.).
+* **`foundry`**: Run Foundry-style `test`-prefixed unit tests and `invariant_`-prefixed stateful invariants.
+* **`overflow`**: Detect integer over/underflows (Solidity >= 0.8.0).
+* **`optimization`**: Maximize the return value of `echidna_opt_`-prefixed functions.
```
elopez: I think the prefix for optimization is the same as the one configured for property mode. Functions must return something special though, it might be worth documenting here.

lib/Echidna/Solidity.hs
```diff
@@ -223,16 +231,17 @@ loadSpecified env mainContract cs = do
                                            solConf.contractAddr
                                            unlimitedGasPerBlock
                                            (0, 0)
-    vm4 <- if isDapptestMode solConf.testMode && setUpFunction `elem` abi
+    vm4 <- if is_testFunction `elem` abi && setUpFunction `elem` abi
```
elopez, line 240: is this intentional? if so, we should document that setUp() gets called on all test modes if the contract has an IS_TEST.

lib/Echidna/Output/Foundry.hs
gustavo-grieco: @BowTiedRadone can you take a look to this and the other comments? There is a big chance we can merge it tomorrow 🤞

radubahmata: @elopez IMO it shouldn't. Echidna always models the fallback as `("", [])`, see: https://github.com/crytic/echidna/blob/1ef15b9c9b72b4dd333ee6428e3e64574d40c229/lib/Echidna/ABI.hs#L41
So args is always empty here. The Solidity fallback function receives raw bytes calldata (not typed ABI parameters), so the typed args list wouldn't be the right representation for it anyway. The current `address(Target).call("")` seems correct to me.

lib/Echidna/Solidity.hs
radubahmata: This seems intentional. The change from `isDapptestMode` to checking `IS_TEST()` means any contract following the Foundry convention gets `setUp()` called regardless of test mode. To me this sound like better behavior since a contract with `IS_TEST()` likely expects its `setUp()` to run even in assertion or property mode. @gustavo-grieco please correct me if my understanding is wrong.

README.md
gustavo-grieco: Fixed by @BowTiedRadone

lib/Echidna/Solidity.hs
gustavo-grieco: Yeap, let's add an IS_TEST guard, if we are using foundry, it makes sense to run that. Watch for our tests.

gustavo-grieco: Fixed, let's check the CI ..

PR #1506, opened 2025-12-27
Improve identification of CBOR bytecode metadata

elopez:
* pick the last prefix if several are present, to work in bytecode-in-bytecode scenarios (e.g., a contract that deploys other contracts)
* identify CBOR length and drop other trailing data (typically constructor arguments)


gustavo-grieco:
I think this will need some amount of testing, in particular, deploying factories.

PR #1512, opened 2026-01-06
Extend Foundry support testing

BowTiedRadone:
This PR fixes compilation errors in `checkAssertions` for Foundry support and adds an end-to-end test for Foundry stateless bug reproduction.

### DIFF
--- a/src/test/Tests/FoundryTestGen.hs
+++ b/src/test/Tests/FoundryTestGen.hs
@@ -47,9 +47,114 @@ foundryTestGenTests = testGroup "Foundry test generation"
               [ ("should be detected", solved "test_assert_not_eq")
               ]
           ]
+      , testGroup "assertEqDecimal"
+          [ testContract' "foundry/FoundryAsserts.sol"
+              (Just "AssertEqDecimalTest") Nothing (Just "foundry/FoundryAsserts.yaml")
+              True FuzzWorker
+              [ ("should be detected", solved "test_assert_eq_decimal")
+              ]
+          ]
+      , testGroup "assertNotEqDecimal"
+          [ testContract' "foundry/FoundryAsserts.sol"
+              (Just "AssertNotEqDecimalTest") Nothing (Just "foundry/FoundryAsserts.yaml")
+              True FuzzWorker
+              [ ("should be detected", solved "test_assert_not_eq_decimal")
+              ]
+          ]
+      , testGroup "assertLt"
+          [ testContract' "foundry/FoundryAsserts.sol"
+              (Just "AssertLtTest") Nothing (Just "foundry/FoundryAsserts.yaml")
+              True FuzzWorker
+              [ ("should be detected", solved "test_assert_lt")
+              ]
+          ]
+      , testGroup "assertGt"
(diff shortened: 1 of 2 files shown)


gustavo-grieco:
Thanks a lot for extending this PR. We will need to test:
* Catching assertX functions: https://getfoundry.sh/reference/forge-std/std-assertions#std-assertions. This needs to be tested in concrete and symbolic mode (check the verification examples).
* Making sure that `assume` is not failing: https://getfoundry.sh/reference/cheatcodes/assume#assume

Besides that, we will need to ensure that the impact in the codebase is minimized. I think there are a few cases that functions that are already defined elsewhere such as isFailedAssertion. Additionally, use .yaml files instead of passing the config in the code.

tests/solidity/foundry/FoundryAssertsSymbolic.yaml
```diff
@@ -0,0 +1,7 @@
+testMode: dapptest
+seed: 1234
+cryticArgs: ["--solc-remaps", "forge-std/=foundry/forge-std/src/"]
+symExec: true
+symExecSMTSolver: z3
```
gustavo-grieco: We will probably use bitwuzla instead

src/test/Tests/FoundryTestGen.hs
```diff
@@ -1,65 +1,98 @@
...
   ]
 
--- | Verify generated test compiles with forge.
--- We use temp directories because we need to test the full forge workflow:
--- forge init (for dependencies) + our generated test + forge build.
-testForgeCompilation :: IO ()
-testForgeCompilation = do
+-- ============================================================================
+-- Test helpers
+-- ============================================================================
```
gustavo-grieco: Please keep the document comments format consistent with the rest of our comments.

src/test/Tests/FoundryTestGen.hs
```diff
@@ -1,65 +1,98 @@
...
+
+-- General utilities
+
+-- | Run an IO action that requires forge, failing if forge is not found.
+requireForge :: IO () -> IO ()
```
gustavo-grieco: Abstract these two in a single function using arguments. Move into a more general unit test code module.

src/test/Tests/FoundryTestGen.hs
```diff
@@ -74,3 +107,219 @@ mkMinimalTest = EchidnaTest
...
+
+-- ============================================================================
+-- Tests
+-- ============================================================================
+
+-- ============================================================================
+-- Concrete Tests
+-- ============================================================================
+
+-- | Verify generated test compiles with forge.
```
gustavo-grieco: Actually, I don't think we need all this code. This tests are the same as normal echidna test, but the target is a foundry project, so they should work with the normal test infrastructure we already have. Check other examples.

radubahmata: Good call! 2ca1e04

gustavo-grieco:
I think two cases remains untested: 
* assume does not produce a failing test
* reverts are failures

PR #1516, opened 2026-01-14
Fix bugs in Foundry test generation

BowTiedRadone:
This PR fixes two Foundry test generation bugs reported in #1465:

1. **Fallback functions**: Generated invalid syntax `Target.()` -> now generates `address(Target).call("")`
2. **Null bytes in arguments**: Generated invalid escape sequences like `hex"<nul><nul><nul>"` -> now generates proper hex like `hex"000102"`

Both fixes include tests that verify the generated code compiles with `forge build`.

### DIFF
--- a/src/test/Tests/FoundryTestGen.hs
+++ b/src/test/Tests/FoundryTestGen.hs
@@ -14,8 +14,7 @@ import System.IO.Unsafe (unsafePerformIO)
 import System.Process (readProcessWithExitCode)
 import Text.Read (readMaybe)
 
-import Common (testContract, testContract', solved, runContract, testConfig)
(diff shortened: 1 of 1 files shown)


elopez:
I think the second bug might be the one we solved here: https://github.com/crytic/echidna/pull/1514
I merged the master branch on the other PR and tried to sort out the conflicts, but please recheck and remove the redundant test if not needed.

radubahmata:
@elopez Thanks! Checked and it looks fine. I think the end-to-end null byte test brings value, because it is checking if the generated test actually compiles with forge. However, I'll leave the final decision up to @gustavo-grieco and you.

PR #1522, opened 2026-01-29
Fix foundry test generation linter hints and tests

BowTiedRadone:


tests/solidity/foundry/FoundryAsserts.sol
```diff
@@ -136,6 +136,20 @@ contract AssertApproxEqRelDecimalTest is Test {
...
+    uint256 public counter;
+
+    // State-mutating function called during sequences
+    function increase(uint256 x) public {
+        counter += x;
+    }
+
+    // Invariant that can be violated when counter > 100
+    function invariant_counter_below_limit() public view {
+        assertTrue(counter <= 100);
```
gustavo-grieco: Use 5 instead of 100, so we can detect this faster. We don't need something very complex.

radubahmata: Done in 146cf6c.

PR #1550, opened 2026-04-02
Allow foundry unit test generation to handle assertion failure with events

gustavo-grieco:
Fix for the remaning issue in #1465


gustavo-grieco:
@elopez can you take a quick look? it is a not a perfect solution, but it fixes one of the last missing features in the foundry unit test case generation

elopez (approved):
Looks good!

PR #1552, opened 2026-04-05
Allow to disable testing of view/pure function, except prefix ones in property mode

### LINKED ISSUE #1351: Config toggle to disable fuzzing view functions

Medusa has this. It would be great to have it in Echidna as well.

It can help the fuzzer waste less runs and prevent having to add blacklistings.

gustavo-grieco:
Fixes #1351

### DIFF
--- a/src/test/Common.hs
+++ b/src/test/Common.hs
@@ -17,6 +17,7 @@ module Common
   , solvedWith
   , solvedWithout
   , solvedUsing
+  , notPresent
   , countCorpus
   , overrideQuiet
   , loadSolTests
@@ -231,6 +232,13 @@ passed n (env, _) = do
     Just t | isOpen t   -> True
     Nothing             -> error ("no test was found with name: " ++ show n)
     _                   -> False
+
+notPresent :: Text -> (Env, WorkerState) -> IO Bool
+notPresent n (env, _) = do
+  tests <- traverse readIORef env.testRefs
+  pure $ case getResult n tests of
+    Nothing -> True
+    _       -> False
 
 verified :: Text -> (Env, WorkerState) -> IO Bool
 verified n (env, _) = do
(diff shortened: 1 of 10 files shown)


src/test/Tests/Integration.hs
```diff
@@ -35,6 +35,11 @@ integrationTests = testGroup "Solidity Integration Testing"
       , ("echidna_revert_always failed",                   passed      "echidna_revert_always")
       , ("echidna_sometimesfalse passed",                  passed      "echidna_sometimesfalse")
       ]
+  , testContractV "basic/excludeviewpure.sol" (Just (>= solcV (0,8,0))) (Just "basic/excludeviewpure.yaml")
+      -- echidna_counter_small is a view property: must not be excluded, and must be solved
+      -- (which also proves increment() is still being called by the fuzzer)
+      [ ("view property must not be excluded and should be solved", solved "echidna_counter_small")
+      ]
```
elopez, line 42: It would be nice if this asserted somehow that the view functions are not called. Maybe add an assert(false); on a view function to make sure nobody calls it?

src/test/Common.hs
```diff
@@ -232,6 +233,13 @@ passed n (env, _) = do
     Nothing             -> error ("no test was found with name: " ++ show n)
     _                   -> False
 
+notPresent :: Text -> (Env, WorkerState) -> IO Bool
+notPresent n (env, _) = do
+  tests <- traverse readIORef env.testRefs
+  pure $ case getResult n tests of
+    Nothing -> True
+    _       -> False
+
```
elopez, line 242: This appears to be unused, please remove it if not needed.

gustavo-grieco: Fixed

src/test/Tests/Integration.hs
gustavo-grieco: Fixed

PR #1554, opened 2026-04-06
Fix shrinking desync between reproducer and vm traces

rappie:
**WARNING: This is 100% vibecoded**

Tested locally and works without any obvious issues.

When a shrink attempt fails, the reproducer was being overwritten with the output of `removeReverts` while the VM was left unchanged. This caused the displayed call sequence and traces to come from different executions.

The root cause is that `removeReverts` replays the sequence from the initial VM, and transactions that previously succeeded can now revert under different timing (due to `catNoCalls` merging delays). These transactions get replaced with `NoCalls`, silently removing critical calls from the reproducer while the VM still reflects the original successful execution.

The fix: don't update the reproducer when shrinking fails. This keeps the reproducer and VM consistent from the last successful shrink.


rappie:
Update: Looks like this causes OOM issues during shrinking

gustavo-grieco:
As a temporary workaround, reduce the shrinking steps, but this is a leak we need to investigate.

rappie:
Latest commit is code from @elopez. Tested and seems to work.

Issue #1476, opened 2025-10-27
[Bug-Candidate]: Test reason to fail (e.g. revert, assertion failure, etc) is not shown consistently


gustavo-grieco:
### Describe the issue:

The reason for a test to fail is very valuable for users, as it quickly shows why it fails. This is particularly valuable when there are no events or the list of transactions to reproduce the failure is empty.

### Code example to reproduce the issue:

A test that fails with one or more properties vs a test that fails with no properties will show the bug.

### Version:

`master`

### Relevant log output:

```shell

```

monperrus:
Great! The diagnosis message should be present in all formats (txt, json, ui).

gustavo-grieco:
Fixed!

monperrus:
Thanks a lot!

PR #1572, opened 2026-05-20
Make sure failure reason is consistent across outputs

gustavo-grieco:
Fix for #1476


elopez (approved):
Looks good, let's make sure CI passes before merging

PR #1580, opened 2026-06-11
Update hevm to support dynamic argument inputs in symbolic mode

elopez:
Update hevm to support dynamic argument inputs in symbolic mode. Also correct the filter guard, `not $ null m.inputs && ...` parsed as `not (null m.inputs && ...)`, negating the whole conjunction.
Bind `not` to the emptiness check alone so the filter means what it says: explore methods that have arguments, not opted out.


gustavo-grieco:
Blocked by https://github.com/argotorg/hevm/pull/1057/. Once that's merged, we will change/remove this check to allow dynamic-argument inputs

PR #1581, opened 2026-06-19
Bump tls to 2.2.2 in flake and stack

elopez:
tls < 2.2.2 is affected by https://github.com/haskell-tls/hs-tls/pull/514.

tls 2.2.2 needs the crypton-x509 1.8 line, which still builds against crypton 1.0.x (the 1.9 line would force crypton >= 1.1 and a much wider cascade). No crypton-connection release targets tls 2.2 (0.4.5 caps < 2.2, 0.4.6 needs >= 2.3), so its tls < 2.2 bound is relaxed, along with warp-tls and tls-session-manager (pulled in by Hoogle in the dev shell); all compile fine against 2.2.2.

In flake.nix the codegen cabal2nix is pinned to the un-overridden base set, otherwise regenerating tls via callHackageDirect loops (cabal2nix transitively depends on tls). stack.yaml additionally pins the crypton-asn1-*/crypton-pem/time-hourglass/hpke/ech-config packages that lts-23.24 lacks; nixpkgs-unstable already carries compatible versions.


gustavo-grieco:
This works correctly in stack, let merge.

Issue #1430, opened 2025-08-23
Improve validation of config values


elopez:
### Describe the desired feature

We should improve validation of some config values and alert the user. For example:

* deployContracts / deployBytecodes lets you deploy contracts on the precompile address space (0x1 ... 0xA) but that is not possible
* contractAddr can possibly also be set to those addresses
* sender / psender might also be set to those addresses (this is potentially fine though)
* function filters lets you have potentially invalid signatures (eg with spaces)

gustavo-grieco:
We also need some validation for the verification and exploration mode. We could have a function that enforces constraints in Env

PR #1585, opened 2026-07-04
Reject precompile deployment addresses

Ap4sh:
Part of #1430

## Summary
- reject `contractAddr` values in the EVM precompile range
- reject `deployContracts` and `deployBytecodes` entries targeting precompile addresses
- add config parser coverage for rejected and allowed deployment addresses

## Tests
- `stack test --test-arguments '-p "Configuration tests"'`
- `stack build --test --no-run-tests --ghc-options='-Werror'`
- `hlint lib/Echidna/Config.hs src/test/Tests/Config.hs`

### DIFF
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -72,6 +72,15 @@ instance FromJSON EConfigWithUsage where
           fail $ show k <> ": value does not fit in 256 bits"
         else
           pure $ fromIntegral value
+      rejectPrecompileAddress field addr
+        | addr `Set.member` precompileAddresses =
+          fail $ field <> ": address " <> show addr <> " is reserved for an EVM precompile"
+        | otherwise = pure addr
+      rejectPrecompileDeployments field =
+        traverse $ \(addr, target) -> do
+          addr' <- rejectPrecompileAddress field addr
+          pure (addr', target)
+      precompileAddresses = Set.fromList [1..10]
 
       txConfParser = TxConf
         <$> v ..:? "propMaxGas" ..!= maxGasPerBlock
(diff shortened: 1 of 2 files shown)


lib/Echidna/Config.hs
```diff
@@ -72,6 +72,15 @@ instance FromJSON EConfigWithUsage where
...
           pure $ fromIntegral value
+      rejectPrecompileAddress field addr
+        | addr `Set.member` precompileAddresses =
+          fail $ field <> ": address " <> show addr <> " is reserved for an EVM precompile"
+        | otherwise = pure addr
+      rejectPrecompileDeployments field =
+        traverse $ \(addr, target) -> do
+          addr' <- rejectPrecompileAddress field addr
+          pure (addr', target)
+      precompileAddresses = Set.fromList [1..10]
```
elopez: There's [a few more precompiles](https://www.evm.codes/precompiled) nowadays; maybe you can use the helper from hevm to keep things in sync? https://github.com/argotorg/hevm/blob/8963e3df0a95936fd34b348afe55e5ac03510236/src/EVM.hs#L3587-L3595

gustavo-grieco:
Beside @elopez comments, I think the rest of the PR looks good, please fix that so we can merge it. Thanks!

Ap4sh:
updated this to use hevm's `isPrecompileAddr'` helper instead of keeping a local list

also added coverage for `0x11` and `0x100`, plus `0x12` / `0x101` as non-precompile boundaries

checked locally with the focused config test and HLint

gustavo-grieco (approved):

PR #1587, opened 2026-07-14
Drain campaign event queue to fix root-cursor memory leak

elopez:
Every consumer reads campaign events from its own dupChan, so the original Chan's read end is never advanced and pins every event ever written, keeping the whole campaign's event history (including full VM snapshots) alive. Drain the original queue from a dedicated thread so the GC can reclaim delivered events, and publish VM-free, strictly forced copies of falsified/optimized tests so queued events can't retain a VM through a lazy thunk.

### DIFF
--- a/lib/Echidna.hs
+++ b/lib/Echidna.hs
@@ -1,6 +1,8 @@
 module Echidna where
 
-import Control.Concurrent (newChan)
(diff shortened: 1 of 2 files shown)


elopez:
Alternative to #1576

PR #1595, opened 2026-08-05
fix: handle verification and Foundry campaign modes

gustavo-grieco:
## Summary

- add an explicit `verification` test mode and enforce its stateless symbolic campaign settings
- avoid zero-worker UI division and dispatch symbolic verification by mode
- make Foundry ABI filtering sequence-length-aware so stateless fuzzing only calls parameterized `test*` functions
- add regression coverage for verification configuration and both Foundry filtering paths

These are kept together because both are small, prerequisite campaign-mode correctness fixes. They remain split into two commits for review.

## Testing

- `stack test echidna:test:echidna-testsuite --fast --test-arguments='--pattern "Configuration tests" --hide-successes --color never'` (16 passed)
- `stack test echidna:test:echidna-testsuite --fast --test-arguments='--pattern "Foundry Integration Testing" --hide-successes --color never'` (3 passed)
- `stack test echidna:test:echidna-testsuite --fast --test-arguments='--pattern "Symbolic tests" --hide-successes --color never'` (3 passed)

### DIFF
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -26,9 +26,22 @@ import Echidna.Types.Solidity
 import Echidna.Types.Test (TestConf(..))
 import Echidna.Types.Tx (TxConf(TxConf), maxGasPerBlock, defaultTimeDelay, defaultBlockDelay)
 
+-- | Verification is a stateless symbolic mode: it uses one symbolic worker
+-- and exactly one transaction per sequence.
+adjustForVerificationMode :: EConfig -> EConfig
+adjustForVerificationMode cfg
+  | isVerificationMode cfg.solConf.testMode =
+      cfg { campaignConf = cfg.campaignConf
+              { workers = Just 0
+              , seqLen = 1
+              , symExec = True
+              }
+          }
+  | otherwise = cfg
+
 instance FromJSON EConfig where
   -- retrieve the config from the key usage annotated parse
-  parseJSON x = (.econfig) <$> parseJSON @EConfigWithUsage x
+  parseJSON x = adjustForVerificationMode . (.econfig) <$> parseJSON @EConfigWithUsage x
 
 instance FromJSON EConfigWithUsage where
   -- this runs the parser in a StateT monad which keeps track of the keys
--- a/lib/Echidna/Config.hs
+++ b/lib/Echidna/Config.hs
@@ -27,7 +27,9 @@ import Echidna.Types.Test (TestConf(..))
 import Echidna.Types.Tx (TxConf(TxConf), maxGasPerBlock, defaultTimeDelay, defaultBlockDelay)
 
 -- | Verification is a stateless symbolic mode: it uses one symbolic worker
--- and exactly one transaction per sequence.
+-- and exactly one transaction per sequence. This is applied when parsing a
+-- config file; the CLI applies it again after its own overrides, since the
+-- test mode can also be selected with @--test-mode@.
 adjustForVerificationMode :: EConfig -> EConfig
 adjustForVerificationMode cfg
   | isVerificationMode cfg.solConf.testMode =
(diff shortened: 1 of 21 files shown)


gustavo-grieco:
@elopez this is ready for your review

tests/solidity/symbolic/verify.yaml
```diff
@@ -1,5 +1,4 @@
-testMode: assertion
-symExec: true
+testMode: verification
 symExecSMTSolver: z3
 workers: 0
 seqLen: 1
```
elopez:
```suggestion
```

If these are applied by default now we should drop them

tests/solidity/symbolic/verify.bitwuzla.yaml
```diff
@@ -1,5 +1,4 @@
-testMode: assertion
-symExec: true
+testMode: verification
 symExecSMTSolver: bitwuzla
 workers: 0
 seqLen: 1
```
elopez:
```suggestion
```

If these are applied by default now we should drop them

lib/Echidna/Solidity.hs
```diff
@@ -156,11 +156,25 @@ filterMethods contractName (Blacklist ig) ms =
...
+--
+-- The set of callable functions depends on @seqLen@, mirroring how tests are
+-- selected in 'Echidna.Test.createTests':
+-- - In fuzz mode (@seqLen == 1@) each "test" function is itself the fuzz
+--   target and is called directly by the fuzzer. Only "test" functions with
+--   at least one parameter are kept; calling other functions (e.g. handlers
+--   like @mint@) would only waste effort since no test checks them.
+-- - In invariant mode (@seqLen > 1@) test functions, "invariant_" functions,
+--   and parameterized handlers retain the existing callable set, which is
+--   exercised in randomized sequences.
```
elopez: so this is meant to match foundry stateless vs stateful fuzzing modes, right?

lib/Echidna/Solidity.hs
```diff
@@ -156,11 +156,25 @@ filterMethods contractName (Blacklist ig) ms =
...
+filterMethodsWithArgs :: Int -> NonEmpty SolSignature -> NonEmpty SolSignature
+filterMethodsWithArgs seqLen ms =
+  case NE.filter keep ms of
     [] -> error "No foundry tests found"
     fs -> NE.fromList fs
+  where
+    keep (n, xs)
+      | seqLen == 1 = T.isPrefixOf "test" n && not (null xs)
+      | otherwise   = T.isPrefixOf "test" n || T.isPrefixOf "invariant_" n || not (null xs)
```
elopez: The underscore of invariants appears to not be required. Also foundry appears to support "statefulFuzz" as a prefix too.

https://github.com/foundry-rs/foundry/blob/master/crates/common/src/traits.rs#L195-L219

elopez: We should also check if we correctly handle "testFail"

elopez: Support for "check" and "prove" in symbolic mode could be cool as well.

lib/Echidna/Config.hs
```diff
@@ -26,9 +26,22 @@ import Echidna.Types.Solidity
...
+              , seqLen = 1
+              , symExec = True
+              }
+          }
+  | otherwise = cfg
+
 instance FromJSON EConfig where
   -- retrieve the config from the key usage annotated parse
-  parseJSON x = (.econfig) <$> parseJSON @EConfigWithUsage x
+  parseJSON x = adjustForVerificationMode . (.econfig) <$> parseJSON @EConfigWithUsage x
```
elopez: Do we need this here if Main.hs already does `let cfg = adjustForVerificationMode cfg`?

lib/Echidna/Solidity.hs
gustavo-grieco: Done

lib/Echidna/Config.hs
gustavo-grieco: Fixed

lib/Echidna/Solidity.hs
gustavo-grieco: Yes

tests/solidity/symbolic/verify.bitwuzla.yaml
gustavo-grieco: Fixed

tests/solidity/symbolic/verify.yaml
gustavo-grieco: Fixed

gustavo-grieco:
@elopez please take another look

elopez (approved):

PR #1609, opened 2026-08-25
fix(ui): isolate shrinking worker state

### LINKED ISSUE #1601: Worker call and gas counters reset to zero when a fuzz worker hits --timeout
### Describe the issue:

When a fuzz worker stops because the campaign time limit was reached, and at least one test still needs shrinking, the final report claims `Total calls: 0` no matter how many calls the campaign actually made. The status line drops to `fuzzing: 0/<testLimit>` and the gas rate goes negative.

Only the reporting is wrong — findings, corpus and coverage are all correct.

**Root cause**

`spawnWorker` gives each worker a single `stateRef` (`lib/Echidna/UI.hs:249`), and that ref is the only channel through which a worker's `WorkerState` reaches the reporting layer — `workerStates` just reads it back (`lib/Echidna/UI.hs:287-288`). The worker publishes into it through the callback `get >>= writeIORef stateRef` (`lib/Echidna/UI.hs:259`).

When a fuzz worker stops on `TimeLimitReached` with tests still needing shrinking, a second, shrink-only worker run is started outside the timeout (`lib/Echidna/UI.hs:276-278`, added for #839) — and it is handed **the same `stateRef`**:

```haskell
when (any needsShrinking tests) $ void $
  runReaderT (runWorker FuzzWorker (get >>= writeIORef stateRef)
              vm dict workerId [] 0 cliSelectedContract) env
```

`runFuzzWorker` unconditionally builds a fresh `initialState` (`lib/Echidna/Campaign.hs:339-347`) with `ncalls = 0`, `ncallseqs = 0`, `totalGas = 0`, `newCoverage = False` and the pristine `dict`. The first thing that run does inside `runStateT` is `lift callback` (`lib/Echidna/Campaign.hs:351`), *before* any shrinking happens — so the zeroed state is written straight over the accumulated one, and every later reader sees zeros:

- `ppTotalCalls` (`lib/Echidna/UI/Report.hs:52-55`) → `Total calls: 0`
- `statusLine` (`lib/Echidna/UI.hs:431-438`) → `fuzzing: 0/N`; and since `deltaGas = totalGas - gasTracker.totalGasConsumed`, the reset makes the delta negative, hence the negative `gas/s`
- the TUI's `Total calls` and gas rate (`lib/Echidna/UI/Widgets.hs:199`, `:273-276`)

**Scope**

Triggers when both hold, so it is easy to hit with any `--timeout`/`timeout:` run that finds something:

1. a fuzz worker stops with `TimeLimitReached`, and
2. at least one test still needs shrinking at that moment.

Confirmed **not** affected:

- Findings and exit code — tests live in `env.testRefs`, not `WorkerState`.
- Corpus and coverage — also `Env` refs.
- `Seed:` — `ppSeed` reads `genDict.defSeed`, which the second run recomputes identically.
- Shrinking quality, despite the `genDict` reset: `shrinkTest` is `(MonadIO, MonadThrow, MonadRandom, MonadReader Env)` with no `MonadState` constraint, and `updateTests` never touches the state, so the worker's dictionary never reaches the shrinker.

**Expected behaviour**

`Total calls:` should report the calls the campaign made, the status line should keep its count, and the gas rate should never be negative.

**Suggested fix**

Stop the shrink-only pass from publishing into the ref the campaign reports from — e.g. hand it a throwaway `IORef`, since nothing needs its live state once the campaign is over. Alternatively, seed the second run with the state the first one finished with so the counters keep accumulating across both.

### Code example to reproduce the issue:

Any contract with a falsifiable property works. Using `tests/solidity/basic/flags.sol` from this repo:

```solidity
contract Test {
  event Flag(bool);

  bool private flag0 = true;
  bool private flag1 = true;

  function set0(int val) public returns (bool){
    if (val % 100 == 0)
      flag0 = false;
  }

  function set1(int val) public returns (bool){
    if (val % 10 == 0 && !flag0)
      flag1 = false;
  }

  function echidna_alwaystrue() public returns (bool){
    return(true);
  }

  function echidna_revert_always() public returns (bool){
    revert();
  }

  function echidna_sometimesfalse() public returns (bool){
    emit Flag(flag0);
    emit Flag(flag1);
    return(flag1);
  }
}
```

Run it with a time limit short enough that shrinking is still outstanding when it fires:

```
echidna basic/flags.sol --format text --test-limit 100000000 --timeout 5 --workers 2
```

### Version:

```
Echidna 2.3.3   (built from master at 2bfa64ef)
slither 0.11.3
```

macOS (aarch64), GHC 9.8.4. Reproduces with both `--format text` and the interactive TUI.

### Relevant log output:

```shell
[19:07:29.90] [status] tests: 1/3, fuzzing: 67593/100000000, values: [], cov: 404, corpus: 3, gas/s: 500363892
[19:07:31.90] [Worker 0] Time limit reached. Stopping.
[19:07:31.90] [Worker 1] Time limit reached. Stopping.
[19:07:31.90] [status] tests: 1/3, fuzzing: 0/100000000, values: [], cov: 404, corpus: 3, gas/s: -750545838
echidna_sometimesfalse: failed!💥
  Reason: ReturnFalse
  Call sequence:
    Test.set0(0)
    Test.set1(0)

echidna_alwaystrue: passing
echidna_revert_always: passing

Unique instructions: 404
Unique codehashes: 1
Corpus size: 3
Seed: 8331856644448189321
Total calls: 0
```

Note the last status line before the summary: the call count drops from 67593 to 0, and `gas/s` is negative.

DefiAudit0x:
## Summary

Fixes #1601 by isolating the shrinking worker's mutable `WorkerState` from the reporting state used by the main fuzzing worker.

When a timeout triggers shrinking, the shrink-only pass starts from `initialWorkerState`. Reusing the same state reference causes the reporting counters to be reset to zero, so the final status can incorrectly report zero calls and zero gas even though the worker performed work before timing out.

This change gives the shrink-only agent a throwaway `IORef initialWorkerState`. The reporting reference remains owned by the main worker and is not overwritten by the shrinking pass.

## Changes

- Added an explicit `IORef WorkerState` parameter to the fuzzer-agent construction path.
- - Added `mkFuzzerAgent` to make agent construction testable.
- - Updated the timeout shrinking path to use an isolated worker-state reference.
- - Added a regression test covering preservation of the reporting state across shrinking.
## Validation

- Focused regression test: `All 1 tests passed`
- - Full Echidna suite: `All 167 tests passed (630.81s)`
- - `git diff --check`: passed
The full suite was run after configuring the required libff, Foundry, Bitwuzla/Z3, solc 0.8.25, and Slither/crytic-compile dependencies. No unrelated tests were changed or weakened.

## Scope

This PR is intentionally limited to the worker-state aliasing/reset issue described in #1601.

### DIFF
--- a/lib/Echidna/Types/Agent.hs
+++ b/lib/Echidna/Types/Agent.hs
@@ -31,6 +31,27 @@ data Agent
       , stateRef :: IORef WorkerState
       }
 
+-- | Construct a fuzzing agent with an explicit state publication reference.
+--
+-- Keeping the reference as an argument lets timeout-triggered auxiliary passes
+-- use a throwaway reference instead of overwriting the reporting state.
+mkFuzzerAgent
+  :: VM Concrete
+  -> GenDict
+  -> Int
+  -> IORef WorkerState
+  -> [(FilePath, [Tx])]
+  -> Int
+  -> Agent
+mkFuzzerAgent vm dict workerId stateRef initialCorpus testLimit =
+  FuzzerAgent { fuzzerId = workerId
+              , initialVm = vm
+              , initialDict = dict
+              , initialCorpus
+              , testLimit
+              , stateRef
+              }
+
 -- | The worker id this agent runs as. There is at most one symbolic worker and
 -- it is always worker 0.
 workerIdOf :: Agent -> WorkerId
--- a/lib/Echidna/UI.hs
+++ b/lib/Echidna/UI.hs
@@ -38,7 +38,7 @@ import Echidna.Output.Corpus (saveCorpusEvent)
 import Echidna.Output.JSON qualified
 import Echidna.Server (runSSEServer)
 import Echidna.SourceAnalysis.Slither (isEmptySlitherInfo)
-import Echidna.Types.Agent (Agent(..), workerTypeOf)
+import Echidna.Types.Agent (Agent(..), mkFuzzerAgent, workerTypeOf)
 import Echidna.Types.Campaign
 import Echidna.Types.Config
 import Echidna.Types.Corpus qualified as Corpus
@@ -249,17 +249,10 @@ ui vm dict initialCorpus cliSelectedContract = do
   spawnWorker env testLimit corpusChunk workerId = do
     stateRef <- newIORef initialWorkerState
 
-    let fuzzerAgent corpus limit =
-          FuzzerAgent { fuzzerId = workerId
-                      , initialVm = vm
-                      , initialDict = dict
-                      , initialCorpus = corpus
-                      , testLimit = limit
-                      , stateRef
-                      }
+    let fuzzerAgent = mkFuzzerAgent vm dict workerId
 
         agent = case workerIDToType env.cfg.campaignConf workerId of
-          FuzzWorker -> fuzzerAgent corpusChunk testLimit
+          FuzzWorker -> fuzzerAgent stateRef corpusChunk testLimit
           SymbolicWorker ->
             SymbolicAgent { initialVm = vm
                           , initialDict = dict
@@ -286,8 +279,9 @@ ui vm dict initialCorpus cliSelectedContract = do
       case stopReason of
         TimeLimitReached | workerType == FuzzWorker -> do
           tests <- traverse readIORef env.testRefs
-          when (any needsShrinking tests) $ void $
-            runAgent (fuzzerAgent [] 0) env
+          when (any needsShrinking tests) $ do
+            shrinkStateRef <- newIORef initialWorkerState
+            void $ runAgent (fuzzerAgent shrinkStateRef [] 0) env
         _ -> pure ()
 
       time <- liftIO getTimestamp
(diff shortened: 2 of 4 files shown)


DefiAudit0x:
Follow-up on the Windows CI result:

The only failing check is `Test with solc 0.8.25 on windows-latest`, where `coverage/boolean.sol` reports `unexpected corpus count` (1 of 167 tests). The same job passes all other tests, including the new Agent regression test, and the Ubuntu/macOS builds and solc matrices pass.

This failure is outside the scope of this PR: #1609 changes only the timeout-triggered shrinking worker state, while this coverage test does not exercise that path. The exact base `master` commit also passed the same Windows/solc 0.8.25 job previously. I reproduced the focused coverage test 20 consecutive times locally without the failure.

I prepared a separate test-only PR, #1610, which pins `seed: 0` and `workers: 1` in `coverage/boolean.yaml`; the focused test passed 10 consecutive times in an isolated baseline worktree. I have kept #1609 unchanged and separated the test-stability proposal so the worker-state fix remains narrowly scoped.

PR #1614, opened 2026-09-01
perf: size RTS capabilities to the worker count and cache the coverage lookup per step

elopez:
Two independent performance fixes found while profiling a property campaign
(`+RTS -s -hT` and a late-cost-centre profiling build). Same seed, same
workload, identical campaign results before and after.

## `perf(rts)`: size the RTS capabilities to the campaign's worker count

The executable was linked with `-with-rtsopts=-A64m -N`, so the RTS created one
capability per core, each with its own 64 MiB nursery, regardless of how many
workers actually run. With nursery chunking (`-A >= 16m`) a single busy worker
churns through all of them, so a **one-worker campaign on a 16-core machine sat
at 1.2 GiB of RTS memory while its live heap was ~13 MB**.

Now the RTS starts with one capability and `main` grows it to
`min cores (workers + 1)` once the config is resolved: one per fuzz/symexec
worker plus one for the UI, and never more than the old `-N` default created.
An explicit `+RTS -N` asking for more is still honoured.

Because the default worker count used to be derived from the capability count
(`getNFuzzWorkers`), it is now resolved in `overrideConfig` from
`getNumProcessors` — same value (cores clamped to 1..4). `getNFuzzWorkers`
keeps its fallback for other entry points.

| `--workers` | capabilities | memory in use |
|---|---|---|
| 1 | `-N2` | 158 MiB |
| 3 | `-N4` | 306 MiB |
| 40 (16-core box) | `-N16` | 1216 MiB — same as the old default |
| 1, `+RTS -N8` | `-N8` | 602 MiB — override honoured |

## `perf(exec)`: cache the coverage vector lookup across EVM steps

The per-instruction coverage hook did two `Map.lookup`s on `env.contracts`, a
`readIORef` plus a codehash `Map` lookup, and a `modifyIORef'` allocating a new
context tuple on **every EVM step** (~380 steps per call on this workload).

The step loop now carries a small `CoverageCache` keyed on the executing code
and only redoes the lookup when the code changes. Coverage is keyed by
codehash, so equal code always means the same vector; the `ByteString` `Eq`
short-circuits to a pointer compare when it is the same object, which the loop
maintains by storing the current step's code on every hit (clones stay on the
fast path). The coverage context is threaded as a value and written back to its
`IORef` once per `execCov` call, so it still spans query interruptions.

Recorded coverage is unchanged: same opIx, same handling of out-of-bounds pcs,
depth ≥ 64 and already-covered locations.

## Numbers

Benchmark: 3-property contract, `seqLen: 100`, 1 worker, fixed seed,
`--test-limit 400000`, 16-core machine.

| | `master` | this PR | Δ |
|---|---|---|---|
| Elapsed | 51.5 s | 45.0 s | **−13 %** |
| Bytes allocated | 252 GB | 232.5 GB | −7.7 % |
| Total memory in use | 1212 MiB | **189 MiB** | **−84 %** |
| Max RSS | 1.32 GB | 247 MB | −81 % |
| Campaign result | corpus 21, 4263 instrs | identical | |

The two effects are independent: `master` forced to `+RTS -N2` gets all of the
memory drop (194 MiB) at the same speed (51.5 s); this PR forced to `-N16` gets
all of the speedup (44.5 s) at the old memory (1225 MiB). Alternating A/B runs
at equal `-N` reproduce the gap (51.8–52.2 s vs 45.3–45.6 s, noise ≈ 0.4 s).

### DIFF
--- a/lib/Echidna/Exec.hs
+++ b/lib/Echidna/Exec.hs
@@ -11,18 +11,19 @@ import Control.Monad.ST (ST, stToIO, RealWorld)
 import Control.Monad.State.Strict (MonadState(get, put), execState, runStateT, MonadIO(liftIO), gets, modify', execStateT)
 import Data.Bits
 import Data.ByteString qualified as BS
-import Data.IORef (readIORef, newIORef, writeIORef, modifyIORef')
+import Data.IORef (readIORef, newIORef, writeIORef)
 import Data.Map qualified as Map
 import Data.Maybe (fromMaybe, fromJust)
 import Data.Text qualified as T
 import Data.Vector qualified as V
+import Data.Vector.Storable qualified as VS
 import Data.Vector.Unboxed.Mutable qualified as VMut
 import Optics.Core
 import Optics.State.Operators
 import System.Environment (lookupEnv, getEnvironment)
 import System.Process qualified as P
 
-import EVM (bytecode, replaceCodeOfSelf, loadContract, exec1, vmOpIx, clearTStorages)
+import EVM (bytecode, replaceCodeOfSelf, loadContract, exec1, clearTStorages)
 import EVM.ABI
 import EVM.Dapp (DappInfo)
 import EVM.Effects (defaultConfig)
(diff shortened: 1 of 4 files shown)


gustavo-grieco:
@aviggiano do we have support in scfuzzbench to trigger some short echidna runs using drips to compare echidna master and this PR?

aviggiano:
@gustavo-grieco yes
the website can create a github issue, and when it is merged, the CI starts a new run
you can select a specific ref from the fuzzer to get different results
one thing I have not tried (needs to be tested) is e.g. having echidna-ref1 vs echidna-ref2 vs echidna-ref3 etc

gustavo-grieco:
I don't think I have permissions to create benchmarks (e.g. https://github.com/scfuzzbench/scfuzzbench/issues/282). Beside that, the website doesn't allow to benchmark the same fuzzer in two different revisions (but I can ask codex to produce a PR for that).

gustavo-grieco:
Finally: https://scfuzzbench.com/runs/gh-33919985770-1/2d96c7ef1de12830a5951155df4c60f4/ thanks to @elopez

lib/Echidna/Exec.hs
```diff
@@ -256,27 +273,70 @@ execTxWithCov tx = do
...
       -- | Execute one instruction on the EVM
       stepVM :: VM Concrete -> IO (VM Concrete)
       stepVM = stToIO . execStateT (exec1 defaultConfig)
 
       -- | Add current location to the CoverageMap
-      addCoverage :: VM Concrete -> IO ()
-      addCoverage !vm = do
-        let (pc, opIx, depth) = currentCovLoc vm
-            contract = currentContract vm
+      addCoverage
```
gustavo-grieco, line 300: What if we extract this as a new function?

aviggiano:
> Finally: https://scfuzzbench.com/runs/gh-33919985770-1/2d96c7ef1de12830a5951155df4c60f4/ thanks to @elopez 

Sorry for taking a little bit long, AI did the PR to include alternative fuzzer versions which broke something hehe but it has been fixed. Also I've added @elopez to @scfuzzbench so that he can start benchmark runs

PR #1615, opened 2026-09-01
perf: stop re-encoding signatures and rendering events in assertion checks

elopez:
In assertion mode, `updateOpenTest` runs `checkETest` for every open test after every transaction, and two things in that path were doing far more work than the check needs:

- **`perf(test): precompute assertion test selectors`** — `checkStatefulAssertion`/`checkFoundryAssertion` re-encoded the test's signature and keccak-hashed it (`abiCalldata`) on every call, only to compare the first 4 bytes of calldata. `AssertionTest` now carries its selector (filled in by the `assertionTest` smart constructor) and the check is a `ByteString.isPrefixOf`.
- **`perf(test): check assertion events by name instead of rendering them`** — assertion and overflow detection went through `extractEvents`, which renders every event of the tx to `Text` (ABI-decoding the arguments, looking up contract names) and then prefix-matched `"AssertionFailed("` / `"Panic(n)"`. `showValues` always parenthesises, so that prefix test is exactly "an event with this name was emitted": the new `hasEventNamed` looks the topic up in the dapp's event map and compares the name. `Panic(n)` can only come from the revert reason, so `hasPanic` decodes just that — unless the contract declares an event literally named `Panic`, in which case it falls back to the old rendering path so behaviour stays identical.
- **`refactor(test): hide the cached assertion selector behind record fields`** — follow-up with no behavior change: `AssertionTest` gets named fields so pattern matches bind only what they use instead of carrying the selector positionally, and `checkFoundryAssertion` takes just the function name it actually uses. A same-seed campaign is identical before/after.

**Numbers** (assertion-mode benchmark, 33 functions, 200k calls, 1 worker, fixed seed; base = tip of PR 1):

| | wall time | allocation |
|---|---|---|
| base | 22.5 s | 121.5 GB |
| + selectors | 14.1 s | 64.2 GB |
| + event-name check | **12.1 s** | **54.0 GB** |

Campaign results are bit-identical at each step (same unique instructions, corpus size and call count). In a `--profiling-detail=late` profile of 60k calls the total drops 29.3 s → 9.0 s and `extractEvents` disappears from the top cost centres; what remains on the Echidna side is the coverage step loop (~8%), transaction generation (~4.5%) and ABI encoding in `setupTx` (~3.7%), with hevm's interpreter at ~60%.

### DIFF
--- a/lib/Echidna/Events.hs
+++ b/lib/Echidna/Events.hs
@@ -79,14 +79,8 @@ extractEvents decodeErrors dappInfo vm =
       _ -> Nothing
 
 -- | Whether the last transaction emitted an event with the given name, under
--- any overload.
---
--- This is the answer @any (isPrefixOf (name <> "(")) . extractEvents@ gives:
--- 'showValues' always parenthesises the arguments, so a rendered event starts
--- with @name <> "("@ exactly when the event has that name, and nothing else
--- 'extractEvents' produces (revert reasons, raw topics, error traces) can start
--- that way. Rendering the arguments of every event is what made that check
--- expensive, and none of it is needed to know the name.
+-- any overload. Checks the name via the event map instead of rendering the
+-- events, which is expensive.
 hasEventNamed :: Text -> DappInfo -> VM Concrete -> Bool
 hasEventNamed name dappInfo vm = any (any isNamed) (traceForest vm)
   where
--- a/lib/Echidna/Events.hs
+++ b/lib/Echidna/Events.hs
@@ -78,6 +78,25 @@ extractEvents decodeErrors dappInfo vm =
         Just $ humanPanic $ decodePanic d
       _ -> Nothing
 
+-- | Whether the last transaction emitted an event with the given name, under
+-- any overload.
+--
+-- This is the answer @any (isPrefixOf (name <> "(")) . extractEvents@ gives:
+-- 'showValues' always parenthesises the arguments, so a rendered event starts
+-- with @name <> "("@ exactly when the event has that name, and nothing else
+-- 'extractEvents' produces (revert reasons, raw topics, error traces) can start
+-- that way. Rendering the arguments of every event is what made that check
+-- expensive, and none of it is needed to know the name.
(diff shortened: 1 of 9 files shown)


lib/Echidna/Events.hs
```diff
@@ -78,6 +78,25 @@ extractEvents decodeErrors dappInfo vm =
...
+-- | Whether the last transaction emitted an event with the given name, under
+-- any overload.
+--
+-- This is the answer @any (isPrefixOf (name <> "(")) . extractEvents@ gives:
+-- 'showValues' always parenthesises the arguments, so a rendered event starts
+-- with @name <> "("@ exactly when the event has that name, and nothing else
+-- 'extractEvents' produces (revert reasons, raw topics, error traces) can start
+-- that way. Rendering the arguments of every event is what made that check
+-- expensive, and none of it is needed to know the name.
+hasEventNamed :: Text -> DappInfo -> VM Concrete -> Bool
```
gustavo-grieco, line 84: Keep these comment lean

lib/Echidna/Test.hs
```diff
@@ -337,10 +330,20 @@ checkAnySelfDestructed _ vm =
 checkPanicEvent :: T.Text -> Events -> Bool
 checkPanicEvent n = any (T.isPrefixOf ("Panic(" <> n <> ")"))
 
+-- | Whether the last transaction is reported with @Panic(n)@ by 'extractEvents',
+-- i.e. it reverted with that panic code. Only the revert reason can render that
+-- way, so the events themselves are not rendered -- unless the contract
+-- declares an event named Panic, which 'checkPanicEvent' would match too.
+hasPanic :: T.Text -> DappInfo -> VM Concrete -> Bool
```
gustavo-grieco, line 336: Same here, keep this comment lean

gustavo-grieco:
Benchmark this, the difference in memory should be measurable

gustavo-grieco:
https://github.com/scfuzzbench/scfuzzbench/issues/297

PR #1618, opened 2026-09-02
perf: cache corpus selection as prefix sums and generate fillers on demand

elopez:
## What

Three costs in the corpus path, all growing with campaign size:

1. **Selection walked the whole corpus per draw.** `selectFromCorpus` ran
   `Set.toDescList`, re-summed every weight, and scanned — O(corpus) list
   materialization per draw, one or two draws per generated sequence, forever.
   The corpus is now snapshotted into a `CorpusSelector` — a sequence vector
   plus cumulative weights — and each draw is one bounded draw plus a binary
   search. The snapshot is cached per worker and stamped with the corpus size,
   which is a sound version tag because the shared corpus only ever grows
   (its single writer is the `Set.insert` on new coverage); it is rebuilt only
   when the corpus grew.

2. **`randseq` generated a full `seqLen` of fresh transactions up front**,
   though the dominant mutations replay corpus transactions and discard most
   of them. Mutations now receive the generator and run it only for the
   filler positions the mutated sequence actually needs.

3. **Every corpus insertion deep-forced the entire corpus** ("corpus is a bit
   too lazy"), re-walking every stored sequence on every new-coverage event —
   quadratic over a campaign. Only the new entry is forced now; everything
   older was forced on its own insertion.

## Compatibility

Same executed-sequence distribution: the mutation mix, the
weight-proportional selection (∝ insertion stamp), and the filler
distribution are unchanged — but the generator is consumed in a different
order, so seeds don't reproduce campaigns across this change and
`values/extreme.yaml`'s pinned seed is repicked once more (all five
`extreme.sol` asserts plus `utf8.sol` pass with margin at a 3000-call limit;
suite-verified).

## Numbers

Selection cost per draw, isolated (GHC 9.8.4, `-O2`, aarch64):

| corpus size | Set walk (before) | prefix sums + binary search |
|---|---|---|
| 100 | 1.36 µs | 80 ns |
| 1,000 | 8.5 µs | 82 ns |
| 5,000 | 29.6 µs | 97 ns |
| 20,000 | 109 µs | 150 ns |

plus one O(corpus) snapshot rebuild per corpus *growth* instead of per draw.

Profiled on a corpus-heavy synthetic bench (180 branchy functions,
seqLen 10, 300k calls, corpus ≈250): `selectFromCorpus` drops from 0.3% of
runtime to 0.0% (28k draws, 225 snapshot rebuilds), and the transaction
generation path from 3.8% to 2.5% (568k → 311k generator entries) from the
deferred fillers.

End to end the fuzzing loop is EVM-bound, so on a real target with a small
corpus this is ~1%; the selection cost it removes is the one that used to
scale linearly with corpus size for the life of the campaign, so the win
grows with long campaigns, large corpora, and short `seqLen`.

### DIFF
--- a/lib/Echidna/Mutator/Corpus.hs
+++ b/lib/Echidna/Mutator/Corpus.hs
@@ -1,13 +1,15 @@
 module Echidna.Mutator.Corpus where
 
+import Control.Monad (replicateM)
 import Control.Monad.Random.Strict (MonadRandom, getRandomR)
 import Data.Maybe (maybeToList)
-import Data.Set qualified as Set
+import Data.Vector qualified as V
+import Data.Vector.Unboxed qualified as VU
 
 import Echidna.Mutator.Array
 import Echidna.Transaction (forceMutateTx, mutateTx, shrinkTx)
 import Echidna.Types (MutationConsts)
-import Echidna.Types.Corpus
+import Echidna.Types.Corpus (CorpusSelector(..))
 import Echidna.Types.Random (weighted)
 import Echidna.Types.Tx (Tx)
 
@@ -51,10 +53,10 @@ cutRange _ len = (min 1 len, len)
 selectAndMutate
   :: MonadRandom m
   => TxsMutation
-  -> Corpus
+  -> CorpusSelector
   -> m [Tx]
-selectAndMutate m corpus = do
-  rtxs <- selectFromCorpus corpus
+selectAndMutate m sel = do
+  rtxs <- selectFromCorpus sel
   k <- getRandomR (cutRange m (length rtxs))
   mutator m $ take k rtxs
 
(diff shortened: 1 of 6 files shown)


tests/solidity/values/extreme.yaml
```diff
@@ -2,4 +2,4 @@ testLimit: 5000
 shrinkLimit: 100
 testMode: assertion
 seqLen: 1
-seed: 10
```
gustavo-grieco: is this change needed?

elopez: yeah, the test is probabilistic and fails otherwise 😢

gustavo-grieco:
Benchmark this, maybe even with 6 hours runs?

gustavo-grieco:
https://github.com/scfuzzbench/scfuzzbench/issues/298

PR #1620, opened 2026-09-03
Sequence generation fixes

elopez:
* The test limit now counts only executed transactions. Each sequence used to be charged one extra call, which doubled the cost of every iteration when `seqLen` is 1
* Corpus mutations now always produce sequences of exactly `seqLen` transactions. The prepend mutation could produce shorter sequences and, when `seqLen` is 1, produced an empty sequence about a fifth of the time
* When `seqLen` is 1 the corpus is now used: about half of the generated transactions are tweaked copies of stored transactions. Stored transactions were previously never fed back into generation, because only strict prefixes of stored sequences were reused
* Fix extreme.sol assertions to check all of them


gustavo-grieco:
Let's keep the added comments lean and minimal, but this looks very close to be merged otherwise.

gustavo-grieco (approved):

PR #1631, opened 2026-09-15
Cap generated msg.value at the sender’s balance

### LINKED ISSUE #1556: [Bug-Candidate]: Sender with zero balance can call the fuzzing harness with msg.value being non-zero
### Describe the issue:

Sender with zero balance can call the fuzzing harness with msg.value being non-zero

### Code example to reproduce the issue:

Repro: https://github.com/rappie/echidna-sender-bug-repro

### Version:

master

### Relevant log output:

```shell
checkSenderBalance(): failed!💥
  Call sequence:
    SenderBalance.checkSenderBalance() Value: 0xd766e4f69b27688c

Traces:
emit Log(senderBalanceAtDeploy=0, senderBalanceNow=0, sender=0x0000000000000000000000000000000000010000, msgValue=15521344913595132044)
```

DefiAudit0x:
## Summary

A fuzzing transaction carrying more `msg.value` than its sender holds is rejected by any real chain before execution, but Echidna kept generating such impossible transactions. This PR caps the generated value at the sender's current balance in the fuzzing world.

Fixes #1556

## Root cause

- `toTx` picked a random sender and generated a value in `[0, maxValue]` with no regard for the sender's funds. With the reporter's config (`balanceAddr: 0`) every non-zero value was unreachable in reality, yet it was executed: `setupTx` credits the destination contract with `tx.value` (`incrementBalance`) and nothing debits the sender, so the campaign explores states no chain can reach and properties inspecting sender balances report false counterexamples.
- Sender balances *do* exist in the fuzzing world: `populateAddresses` seeds every sender with `balanceAddr` in `vm.env.contracts`, and internal transfers (e.g. a constructor's `transfer`) update them there — that store is the source of truth this fix reads.

## The change

- `toTx` now receives the deployed-contracts map (already threaded through `genTx`/`genTxFromPrototype` from `vm.env.contracts`) and caps the generated value at `min maxValue (senderBalance sender)`; a missing entry means zero funds.
- With `balanceAddr: 0` (as in the repro) senders now only transact with value 0; the default funding (`0xffffffff`) and `maxValue` (100 eth) still allow large values — users who want higher reachable values should raise `balanceAddr` alongside `maxValue`.
- Symbolic execution's `modelToTx` is untouched: its values come from an SMT model solved under hevm's balance constraints.

## Test plan

- New integration regression test `basic/zero-balance-sender.sol` + yaml mirroring the issue's repro: `balanceAddr: 0`, a payable function asserting `msg.value <= msg.sender.balance`. Before this change the fuzzer drives it to failure; after, it passes.
- Existing value-related tests (`values/payable.sol` needs value 129, `basic/balance.sol` pins sender balances) are unaffected since their required values are within sender funds.


gustavo-grieco:
I don't think we need this, as accounts can increase their balance at any time (e.g. account X will have Y balance at one tx and Z balance the next tx). On top of that, only a very small amount of contracts use account.balance, but almost never from EOA, so this code won't be useful in practice and only add complexity.

DefiAudit0x:
Thanks for the review. I agree that an account's balance can change over time, and the fix does not assume that it is static.

The issue is that Echidna currently generates a transaction with `msg.value` greater than the sender's balance at the time the transaction is generated. In the original report, the sender has a zero balance but Echidna still produces a non-zero `msg.value`, which is rejected by a real chain before contract execution.

The cap is applied against the sender's current balance in the VM state. If the sender receives funds earlier in the sequence, the updated balance is used for subsequent transactions. So this is not intended to model a permanently fixed balance; it prevents impossible transactions from being generated from the current state.

The regression test reproduces the reported zero-balance case. Without the change, Echidna reaches the payable function with a non-zero value from a zero-balance sender; with the change, only reachable values are generated.

I agree that the practical impact may be limited to contracts and campaigns that inspect sender balances, but for those campaigns the current behavior produces impossible states and false counterexamples. Do you think this narrower scope would address the complexity concern?

gustavo-grieco:
There is no point in decrementing the balance during the transaction. The code under fuzzing cannot see that effect. The only value that the contract can see is sender.balance that can be any value, even zero

